PR #98, opened 2014-11-07
add ability to define ScalaTest default super class per project.

jozic:
ScalaTest offers many options for base classes, so if you always use one which is not FunSuite, then you have to change it in the dialog every time you create a test. 
New text field is added to Scala Project Settings panel.


jozic:
when do you guys plan to merge it? or is there  anything else i should do?

Alefas:
Hi, Roman had long sick leave plus long public holidays in Russia. Sorry for inconveniences. We will take a look for that soon.

PR #116, opened 2015-01-07
Fixed "Chop down if long" formatter behaviour when parameters list is too long.

fkorotkov:
Fixed "Chop down if long" formatter behaviour when parameters list is too long.

Before because of an aligment for children of ScParameters a ScalaBlock for ScParameterClause child was indented according to '('

for example:

```
def foo(param1: Bar, param2: Bar) = {}
```

was reformatted to:

```
def foo(
              param1: Bar,
              param2: Bar
) = {}
```

now it will become:

```
def foo(
  param1: Bar,
  param2: Bar
) = {}
```

See a test case that I've added. Also all tests in org.jetbrains.plugins.scala.lang.formatter package are passing.

### DIFF
--- a/src/org/jetbrains/plugins/scala/lang/formatting/getDummyBlocks.scala
+++ b/src/org/jetbrains/plugins/scala/lang/formatting/getDummyBlocks.scala
@@ -120,9 +120,7 @@ object getDummyBlocks {
       }
       case _ =>
     }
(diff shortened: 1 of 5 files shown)


jamesbrain:
I can't reproduce the behaviour that is claimed to be fixed.
Could you please provide formatting settings you used to obtain the example with 'foo' function?

PR #117, opened 2015-01-09
Fixed run configurations to run IDEA within an imported SBT project.

fkorotkov:
Fixed run configurations to run IDEA within an imported SBT project.

Note: not sure how to test 'sbt test' because it is still failing for me.


dedoz:
Could you please elaborate on th issue you're trying to fix?
My guess is ```IDEA``` / ```IDEA (No ProcessCancelledException)``` configurations fail to launch for you, am I right?
Regarding sbt test, did you run ```packagePlugin``` task before trying to run tests from sbt?

P.S. scala plugin artifact has a relatively complicated structure, thereby simply providing root project's target  directory as plugin classpath to Idea won't do the trick.

PR #123, opened 2015-01-13
Fix for SCL-7831 + tests

ilinum:


niktrop:
Merged manually. Just replaced type checks with pattern matching.
https://github.com/JetBrains/intellij-scala/commit/7ae4a7d8d75b1c772ac9738c85503055a97cd404

PR #127, opened 2015-02-03
[ SCL-5775 ] ScalaTest always uses parent project folder when executing ...

jdevelop:
...a ScalaTest in a module of a multi-module project

https://youtrack.jetbrains.com/issue/SCL-5775


jamesbrain:
It makes integration tests for ScalaTest and Specs2 fail with "working directory does not exist". I'll see why exactly and how it can be fixed asap.

jdevelop:
I guess that's because by the convention that the test should be invoked in the **Maven module** directory - not parent project directory.
I've got another issue with my changes - when I build and install the plugin for IDEA Ultimate 139.1117 - it complains that scala/Option is not found. I wonder why, because my second commit gets rid of the Option completely and doesn't use any scala-specific dependencies.

jamesbrain:
Sorry for the delay, but the good news is I've fixed the issue with integration tests and (finally) merged the changes into master. Once again, my deepest apologies for the delay.

PR #130, opened 2015-03-16
Warning for non-shared source directories that are outside of project base

dancingrobot84:
@pavelfatin please :eyes:

### DIFF
--- a/src/org/jetbrains/sbt/project/ExternalSourceRootResolution.scala
+++ b/src/org/jetbrains/sbt/project/ExternalSourceRootResolution.scala
@@ -19,7 +19,18 @@ trait ExternalSourceRootResolution { self: SbtProjectResolver =>
 
     val projects = projectToModuleNode.keys.toSeq
 
-    externalSourceRootGroupsIn(projects).map { group =>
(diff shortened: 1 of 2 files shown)


src/org/jetbrains/sbt/project/ExternalSourceRootResolution.scala
```diff
@@ -83,24 +94,33 @@ trait ExternalSourceRootResolution { self: SbtProjectResolver =>
...
 
-    val sharedRoots = projectRoots
-            .filter(it => it.isExternal && !internalSourceDirectories.contains(it.root.directory))
-            .groupBy(_.root)
-            .mapValues(_.map(_.project).toSet)
-            .map(p => SharedRoot(p._1, p._2.toSeq))
-            .toSeq
-
+    projectRoots
+      .filter(it => it.isExternal && !internalSourceDirectories.contains(it.root.directory))
```
pavelfatin: !internalSourceDirectories.contains will exclude all roots that are "shared" with other projects. That's good for the project model, but not for the user warning.

pavelfatin:
This warning isn't perfect (see the comment), yet better than nothing. I've also changed the title from "Fix shared sources detection algorithm" to "Warning for non-shared source directories that are outside of project base".

PR #145, opened 2015-05-04
Fixing an issue where running tasks on a custom SBT jar would not use the correct main class

cowboygneox:
We use a custom SBT jar with a different main class listed in the Manifest. Loading the project into IntelliJ works fine, but running tasks does not. This should fix that.

Any comments, questions, or concerns are very welcome!

### DIFF
--- a/src/org/jetbrains/sbt/runner/SbtRunConfiguration.scala
+++ b/src/org/jetbrains/sbt/runner/SbtRunConfiguration.scala
@@ -1,6 +1,8 @@
 package org.jetbrains.sbt.runner
 
+import java.io.File
 import java.util
+import java.util.jar.{Attributes, JarFile}
 
 import com.intellij.execution.Executor
 import com.intellij.execution.configuration.EnvironmentVariablesComponent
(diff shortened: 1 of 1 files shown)


src/org/jetbrains/sbt/runner/SbtRunConfiguration.scala
```diff
@@ -78,6 +75,11 @@ class SbtRunConfiguration(val project: Project, val configurationFactory: Config
     envirnomentVariables.putAll(params.getEnvironmentVariables)
   }
 
+  def determineMainClass(launcherPath: String): String = {
+    val jf = new JarFile(new File(launcherPath))
+    val attributes = jf.getManifest.getMainAttributes
+    attributes.getValue("Main-Class")
```
nkly, line 81: `getValue` returns null if no "Main-Class" attribute is present. Could you change the signature of `determineMainClass` so it will return `Option[String]` instead? And fix places below so that if `determineMainClass` returns `None` default main class (`xsbt.boot.Boot`) is used?

nkly:
Could you please rebase your changes against `idea14.1.eap` branch? We're changing the structure of repository and deleting "master" branch

PR #153, opened 2015-06-04
Fix for SCL-8803

ilinum:


ilinum:
Messed up branches. Will send another pull-request

PR #160, opened 2015-06-17
Fixing an issue where running tasks on a custom SBT jar would not use the correct main class

cowboygneox:
We use a custom SBT jar with a different main class listed in the Manifest. Loading the project into IntelliJ works fine, but running tasks does not. This should fix that.

Also includes changes per comments in original pull request: https://github.com/JetBrains/intellij-scala/pull/145

Any comments, questions, or concerns are very welcome!


cowboygneox:
@dancingrobot84 I believe I have captured your comments. I did not use `Option[String]` in `determineMainClass` because it would create duplication of the fallback main class `xsbt.boot.Boot`; instead, I just added the `getOrElse` into the `determineMainClass`.

PR #168, opened 2015-07-02
Fix import as @uncheckedVariance

ilinum:
Now this works correctly:

```
import scala.annotation.unchecked.{ uncheckedVariance => uV }

Test[+T] {
  arr: Array[T@uV] = null
}
```


src/org/jetbrains/plugins/scala/annotator/ScalaAnnotator.scala
```diff
@@ -1065,7 +1065,14 @@ class ScalaAnnotator extends Annotator with FunctionAnnotator with ScopeAnnotato
 
   def childHasAnnotation(teOption: Option[ScTypeElement], annotation: String): Boolean = teOption match {
     case Some(te) => te.breadthFirst.exists {
-      case annot: ScAnnotations => annot.getAnnotations.exists(_.getText == annotation)
+      case annot: ScAnnotationExpr =>
+        annot.constr.reference match {
+          case Some(ref) => Option(ref.resolve()) match {
```
Alefas, line 1070: ref.bind()

PR #170, opened 2015-07-07
Fix for SCL-7465: Infix expressions can have type args

ilinum:
Minor change in parser, so that infix expression can have type arguments. 
Fixed [SCL-7465](https://youtrack.jetbrains.com/issue/SCL-7465)


ilinum:
"Convert to infix expression" and "Convert from infix expression" works

PR #180, opened 2015-07-15
Fix Java representation of Value types #SCL-5367 Fixed

ilinum:
Fix for [SCL-5367](https://youtrack.jetbrains.com/issue/SCL-5367)
Now value types work correctly in Java!

### DIFF
--- a/src/org/jetbrains/plugins/scala/lang/psi/types/ScTypePsiTypeBridge.scala
+++ b/src/org/jetbrains/plugins/scala/lang/psi/types/ScTypePsiTypeBridge.scala
@@ -10,7 +10,7 @@ import com.intellij.psi._
 import com.intellij.psi.search.GlobalSearchScope
 import org.jetbrains.plugins.scala.extensions._
 import org.jetbrains.plugins.scala.lang.psi.api.statements._
-import org.jetbrains.plugins.scala.lang.psi.api.toplevel.typedef.{ScObject, ScTypeDefinition}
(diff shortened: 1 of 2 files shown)


src/org/jetbrains/plugins/scala/lang/psi/types/ScTypePsiTypeBridge.scala
```diff
@@ -169,6 +171,11 @@ trait ScTypePsiTypeBridge {
       case ScCompoundType(Seq(typez, _*), _, _) => toPsi(typez, project, scope)
       case ScDesignatorType(c: ScTypeDefinition) if ScType.baseTypesQualMap.contains(c.qualifiedName) =>
         toPsi(ScType.baseTypesQualMap.get(c.qualifiedName).get, project, scope, noPrimitives, skolemToWildcard)
+      case ScDesignatorType(valType: ScClass) if isValueType(valType) && !noPrimitives =>
```
Alefas: No primitives looks wrong.
This parameter required for parameterized type, so if ValueClass(OtherNormalClass) => Foo[ValueClass] goes to Foo[OtherNormalClass].

ilinum: So the reason I did it was for the following case:

```
class DoubleWrapper(val d: Double)
```

`Foo[DoubleWrapper]` needs to be converted to `Foo[DoubleWrapper]`
NOT `Foo[Double]`, which is wrong because it would be converted to Java as `Foo<double>`

But you are right about Wrappers on non-primitive types.

PR #185, opened 2015-07-17
Fix : ScalaNameSuggestionProvider incorrectly removes previous suggestions

gdarmont:
A very small fix, but it will re-enable a little IntelliJ renaming goodness :)

Hope this helps.
Guillaume


Alefas:
What case it fixes?

gdarmont:
The problem occurs when asking for a field rename (MAJ + F6).

As a picture is worth a thousand word, here's an example :
When Scala Plugin is disabled :
Scala Plugin disabled
When Scala Plugin is enabled :
Scala Plugin enabled

As you can see, the images above show the bug in for a field in a Java class. There maybe others places where this bug can occur, but it's hard to check every context.

The origin of the bug lies in the code in MyLookupExpression (https://github.com/JetBrains/intellij-community/blob/master/platform/lang-impl/src/com/intellij/refactoring/rename/inplace/MyLookupExpression.java#L66-68), where every NameSuggestionProvider available is asked for suggestions, gathered in the same Set (variable name). As the ScalaNameSuggestionProvider is run after the JavaNameSuggestionProvider, all suggestions contributed by JavaNameSuggestionProvider are cleared.

Hope this helps.
Guillaume

Alefas:
Thank you very much for this fix, looks important.

PR #188, opened 2015-07-23
Kind Projector code cleanup

ilinum:


src/org/jetbrains/plugins/scala/lang/psi/impl/base/types/ScParameterizedTypeElementImpl.scala
```diff
@@ -105,27 +92,23 @@ class ScParameterizedTypeElementImpl(node: ASTNode) extends ScalaPsiElementImpl(
...
         case (param: ScParameterizedTypeElement, i) if inlineSyntaxIds.contains(param.typeElement.getText) =>
           val name = generateName(i)
-          parameters += param.getText.replace("?", name)
-          name
-        case (a, _) => a.getText
-      }
+          (Some(param.getText.replace("?", name)), name)
+        case (a, _) => (None, a.getText)
+      }.unzip
+      val paramText = paramOpt.collect {
```
Alefas, line 104: Just flatten

PR #195, opened 2015-08-29
Nested stateful monads inspection was added

Tolsi:
I found a useful opportunity to see in IDE that you made a mistake on the nested monads which carry state.

But I have doubts about the implementation of, in particular, verify that the code block returns the type of storing the state of the monads, I have not found a way to do it, but to check the string type name. How can I do it a better way?

ps I also want to make inspection checked that monads containing state code processed explicitly (Try using a pattern-matching or isSuccess or etc; Future with onResult or onSuccess and etc; may be Option's), if they are not passed to other methods and do not return from the method, but it is difficult to me have a deal with the representation of the code in the form of a psi tree and existing opportunities of processing this tree. In any case it will be the individual pull request. I'll be glad to get a reference to any material on this topic.

### DIFF
--- a/src/org/jetbrains/plugins/scala/codeInspection/monads/StatefulMonads.scala
+++ b/src/org/jetbrains/plugins/scala/codeInspection/monads/StatefulMonads.scala
@@ -1,0 +1,21 @@
+package org.jetbrains.plugins.scala.codeInspection.monads
+
+import com.intellij.openapi.project.Project
+import org.jetbrains.plugins.scala.codeInspection.InspectionsUtil
+import org.jetbrains.plugins.scala.lang.psi.types.ScType
+
+import scala.concurrent.Future
+import scala.util.Try
+
+/**
+ * @author Sergey Tolmachev (tolsi.ru@gmail.com)
+ * @since 29.09.15
+ */
+object StatefulMonads {
+  private[monads] final lazy val StatefulMonadsTypes: Set[Class[_]] = Set(classOf[Future[_]], classOf[Try[_]])
+  private[monads] final lazy val StatefulMonadsTypesNames = StatefulMonadsTypes.map(_.getCanonicalName)
+
+  private[monads] def isStatefulMonadType(t: ScType, projection: Project): Boolean = {
+    StatefulMonadsTypesNames.exists(typeName => InspectionsUtil.conformsToTypeFromClass(t, typeName, projection))
+  }
+}
--- a/resources/META-INF/scala-plugin-common.xml
+++ b/resources/META-INF/scala-plugin-common.xml
@@ -825,6 +825,10 @@
                          displayName="Convert expression to Single Abstract Method (SAM)" groupPath="Scala" groupName="General"
                          shortName="ConvertExpressionToSAM" level="WARNING"
                          enabledByDefault="true" language="Scala"/>
+        <localInspection implementationClass="org.jetbrains.plugins.scala.codeInspection.monads.NestedStatefulMonadsInspection"
+                         displayName="Nested stateful monads" groupPath="Scala" groupName="General"
+                         shortName="NestedStatefulMonads" level="WARNING"
+                         enabledByDefault="true" language="Scala"/>
         <!--<localInspection implementationClass="org.jetbrains.plugins.scala.codeInspection.allErrorsInspection.AnnotatorBasedErrorInspection"
                         displayName="All Errors Tool" groupPath="Scala" groupName="General"
                         shortName="ScalaAllErrors" level="WARNING"
--- a/resources/inspectionDescriptions/NestedStatefulMonads.html
+++ b/resources/inspectionDescriptions/NestedStatefulMonads.html
@@ -1,0 +1,5 @@
+<html>
+<body>
+This inspection reports nested stateful monads which may lead to architectural errors.
+</body>
+</html>
--- a/src/org/jetbrains/plugins/scala/codeInspection/monads/NestedStatefulMonadsInspection.scala
+++ b/src/org/jetbrains/plugins/scala/codeInspection/monads/NestedStatefulMonadsInspection.scala
@@ -1,0 +1,29 @@
+package org.jetbrains.plugins.scala.codeInspection.monads
+
+import com.intellij.codeInspection.ProblemsHolder
+import org.jetbrains.plugins.scala.codeInspection.AbstractInspection
+import org.jetbrains.plugins.scala.codeInspection.monads.NestedStatefulMonadsInspection._
+import org.jetbrains.plugins.scala.codeInspection.monads.StatefulMonads._
+import org.jetbrains.plugins.scala.lang.psi.api.expr.ScMethodCall
+import org.jetbrains.plugins.scala.lang.psi.types.{Any, ScParameterizedType}
+
+/**
+ * @author Sergey Tolmachev (tolsi.ru@gmail.com)
+ * @since 29.09.15
+ */
+object NestedStatefulMonadsInspection {
+  private[monads] final val Annotation = "Nested stateful monads"
+}
+
+final class NestedStatefulMonadsInspection extends AbstractInspection(Annotation) {
+  override def actionFor(holder: ProblemsHolder) = {
+    case call: ScMethodCall =>
+      val project = call.getProject
+      call.getType().getOrAny match {
+        case outer @ ScParameterizedType(_, typeArgs)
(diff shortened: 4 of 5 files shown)


src/org/jetbrains/plugins/scala/codeInspection/monads/StatefulMonads.scala
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * @author Sergey Tolmachev (tolsi.ru@gmail.com)
+ * @since 29.09.15
+ */
+object StatefulMonads {
+  private[monads] final val RootTypeCanonicalNamePrefix = "_root_."
+  private[monads] final lazy val StatefulMonadsTypes: Set[Class[_]] = Set(classOf[Future[_]], classOf[Try[_]])
+  private[monads] final lazy val StatefulMonadsTypesNames = StatefulMonadsTypes.map(t => RootTypeCanonicalNamePrefix + t.getCanonicalName)
+
+  // todo I hate myself :<
```
Tolsi: How can I do it a better way?

src/org/jetbrains/plugins/scala/codeInspection/monads/StatefulMonads.scala
```diff
@@ -0,0 +1,27 @@
...
+  private[monads] def isStatefulMonadType(t: ScType): Boolean = {
+    val psiEl = t.unpackedType.asInstanceOf[ScDesignatorType].element.asInstanceOf[PsiClass]
+    val superTypes = psiEl.getImplementsListTypes
+    val selfTypeAndSuperTypesNames = superTypes.map(_.getCanonicalText).map(t => RootTypeCanonicalNamePrefix + cleanTypeParams(t)) :+ cleanTypeParams(t.canonicalText)
+    selfTypeAndSuperTypesNames.exists(StatefulMonadsTypesNames.contains)
```
Tolsi: How can I do it a better way?

niktrop:
Hi, Sergey, 
Thank you very much for your help in making scala plugin better. To make this inspection actually work you need to add it to the scala-plugin-common.xml. I did it for you in this patch:
https://yadi.sk/d/lbDEfYtkimTy8
Also, you need to add description for your inspection to the file NestedStatefulMonad.html.

src/org/jetbrains/plugins/scala/codeInspection/monads/StatefulMonads.scala
niktrop: You can use method InspectionsUtil.isExpressionOfType, I just moved it from the package for collection inspections.
https://github.com/JetBrains/intellij-scala/commit/1db0970fa27dd947ffdd5db6b5d8a70198efe311

niktrop:
Your idea for second inspection is good. Unfortunately we have no documentation or guide for working with psi tree. But if you will have any question about implementation, don't hesitate to ask.

src/org/jetbrains/plugins/scala/codeInspection/monads/StatefulMonads.scala
Tolsi: @niktrop how can I check that expression if of type Future of Try[Any] with this method, not simply Future? If I can't, then how can I use it with ScType? isExpressionOfType method looks like my dirty method. 0:) ps может на русском?

niktrop: Yep, my refactoring was not well suited for your task. Now you can pass ScType:
https://github.com/JetBrains/intellij-scala/commit/d2de9f714086f253b2754ffcd11a82f1f101947a

The problem with your method is that you are trying to check conformance on the level of psi elements instead of using methods of ScType. It's the key purpose of our model of scala type system, after all.
PS. This discussion may be of interest for other people, who cannot speak our beautiful language :)

Tolsi:
@niktrop lgtm, u?

And about the second idea: I'll pull request with the tests, and with your help, I hope, we will reach the goal.

ps may be add Option[T] to this inspection? It's not quite the stateful monad. But it makes no sense to put one Option into another Option :) Or create a new inspection
pps tests are failed... I will fix them soon

src/org/jetbrains/plugins/scala/codeInspection/monads/StatefulMonads.scala
```diff
@@ -0,0 +1,21 @@
...
+/**
+ * @author Sergey Tolmachev (tolsi.ru@gmail.com)
+ * @since 29.09.15
+ */
+object StatefulMonads {
+  private[monads] final lazy val StatefulMonadsTypes: Set[Class[_]] = Set(classOf[Future[_]], classOf[Try[_]])
+  private[monads] final lazy val StatefulMonadsTypesNames = StatefulMonadsTypes.map(_.getCanonicalName)
+
+  private[monads] def isStatefulMonadType(t: ScType, projection: Project): Boolean = {
+    StatefulMonadsTypesNames.exists(typeName => InspectionsUtil.conformsToTypeFromClass(t, typeName, projection))
```
Tolsi, line 19: @niktrop `InspectionsUtil.conformsToTypeFromClass(ScDesignatorType(ScTraitImpl(Future)), "scala.concurrent.Future", call.projection))` returns false :( 
`typeFromClassName("scala.concurrent.Future", call.projection)` returns `Some(ScParameterizedType(Future[NotInferedT]))`
`ScDesignatorType(ScTraitImpl(Future)).conforms(ScParameterizedType(Future[NotInferedT])) == false`
What I doing wrong?

niktrop, line 19: I did not get it. What is ScTraitImpl(Future)? We have no apply method in ScTraitImpl, and what is `Future`? And you probably mean call.getProject in the 3rd argument?

niktrop, line 19: Ah, it looks like toString method. What expression do you evaluate?

Tolsi, line 19: @niktrop see `NestedStatefulMonadsInspection.scala:23` and `NestedStatefulMonadsInspection.scala:25`, `NestedStatefulMonadsInspectionTest.scala:39`. `scala.concurrent.Future` - it's a trait which contains interface of stateful async monad

niktrop, line 19: `InspectionsUtil.conformsToTypeFromClass(call.getType().getOrAny, "scala.concurrent.Future", call.getProject)`
returns true for both `Future { Future {}}` and `Future{}`

niktrop, line 19: About adding `Option`, you can make this inspection configurable, like inspections on collections.  Then it would be possible to add or remove types of monads you want.

Tolsi, line 19: @niktrop check also `scala.util.Success(Array(1))` (test_12), I need checks also nested class argument type. `Try[Array[T]]` is not nested stateful monad, but `Future[Future[T]]` or `Try[Future[T]]` does

`returns true for both Future { Future {}} and Future{}` because it checks only first types arguments level

niktrop, line 19: You need to use actual type of expression in this method, don't extract designator type from parameterized type. Something like this:

```
case call: ScMethodCall =>
      val project = call.getProject
      call.getType().getOrAny match {
        case outer @ ScParameterizedType(_, typeArgs)
          if isStatefulMonadType(outer, project) && typeArgs.exists(isStatefulMonadType(_, project)) =>
          holder.registerProblem(call, Annotation)
        case _ =>
      }
```

Also, you need to remove `_root_` prefix from qualified names. They are stored in caches without it.

Tolsi:
@niktrop looks like I'm finished with this. May I improve this inspection with configurable types later or it's should be totally finished before merge? It's looks like not so easy to do this.

src/org/jetbrains/plugins/scala/codeInspection/monads/StatefulMonads.scala
Tolsi, line 19: @niktrop thanks you, it's working

niktrop:
I disabled your inspection by default. It turned out not that safe:
https://youtrack.jetbrains.com/issue/SCL-9128

Tolsi:
@niktrop okey

PR #199, opened 2015-09-03
Tests for fixed issues

ilinum:


.idea/inspectionProfiles/profiles_settings.xml
```diff
@@ -4,7 +4,7 @@
     <option name="USE_PROJECT_PROFILE" value="true" />
     <version value="1.0" />
     <list size="7">
-      <item index="0" class="java.lang.String" itemvalue="GITIGNORE.UNUSED_ENTRY" />
+      <item index="0" class="java.lang.String" itemvalue="UNUSED ENTRY" />
```
Alefas: Do not commit .idea files without necessity.

ilinum: That was an accident. I will fix now

PR #206, opened 2015-09-14
Make it obvious when using empty substitutor during debugging

ilinum:
After this commit, it is obvious when you're using Empty Substitutor

Before:
emptynonobvious
After:
emptyobvious

### DIFF
--- a/src/org/jetbrains/plugins/scala/lang/psi/types/ScSubstitutor.scala
+++ b/src/org/jetbrains/plugins/scala/lang/psi/types/ScSubstitutor.scala
@@ -8,20 +8,22 @@ import com.intellij.psi._
 import org.jetbrains.plugins.scala.lang.psi.api.base.ScFieldId
 import org.jetbrains.plugins.scala.lang.psi.api.base.patterns.ScBindingPattern
 import org.jetbrains.plugins.scala.lang.psi.api.statements.ScFunction
-import org.jetbrains.plugins.scala.lang.psi.api.statements.params.{ScTypeParam, ScParameter}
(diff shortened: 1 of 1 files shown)


src/org/jetbrains/plugins/scala/lang/psi/types/ScSubstitutor.scala
```diff
@@ -8,20 +8,22 @@ import com.intellij.psi._
...
 
-import scala.annotation.tailrec
 import scala.collection.immutable.{HashMap, HashSet, Map}
 
 /**
 * @author ven
 */
 object ScSubstitutor {
-  val empty: ScSubstitutor = new ScSubstitutor()
+  val empty: ScSubstitutor = EmptySubstitutor
```
Alefas: It can be object with name 'empty'. There is no any needs to have two of them.

ilinum: @Alefas Actually, what you are suggesting will not work, a lot of code just won't compile in the places where ScSubstitutor type is inferred from ScSubstitutor.empty. Here is an example:

```
def foo() = {
  Seq(1, 2, 3).foldLeft(ScSubstitutor.empty) {(s, p) =>
    s.bindT(null, null) 
    //Type mismatch: expression of type ScSubstitutor does not conform to expected type ScSubstitutor.empty.type
  }
}
```

Alefas: Then just anonymous class, but not unnecessary object.

PR #217, opened 2015-10-03
SAM Autotupling inspection Fix

ilinum:
This PR closes [SCL-9069](https://youtrack.jetbrains.com/issue/SCL-9069).

The problem: Method Calls with SAMs are highlighted

The cause: SAM is not calculated during shape resolve

The solutions: 
- Check for SAM during shape resolve as well (this is what I implemented now)
  - The downside is that it might be more expensive (but not really because (almost) everything is cached)
  - The other downside is that it makes already complicated code more complicated
- Do a quick hack in AutoTupling inspection not to check SAMs
  - This is ugly but it may work (I don't think we had any other problems because of SAM not being calculated during shape resolve)


ilinum:
@Alefas what do you think?

Also, should ScalaPsiUtil.toSAMType() be cached? This wouldn't be very hard

Alefas:
Test is missing. Please add.

ilinum:
Added test here: https://github.com/JetBrains/intellij-scala/pull/226

PR #186, opened 2015-07-21
SCL-2507 & minor refactor

SergeyAfanasev:
default parameters should drive type inference

PR #218, opened 2015-10-04
Default parameters should drive type inference: SCL-2507

ilinum:
This PR is reworked https://github.com/JetBrains/intellij-scala/pull/187 and https://github.com/JetBrains/intellij-scala/pull/186

It closes [SCL-2507](https://youtrack.jetbrains.com/issue/SCL-2507)

[SCL-2508](https://youtrack.jetbrains.com/issue/SCL-2508) is somewhat closed. However, the error highlighting is in the wrong place.

Consider the following example:

```
def blerg[A](a: A = ""): A  = a //the string here is highlighted
blerg[Int]() //this line is not highlighted
```

The error should be fixed, since often functions you call are in a different file and you won't be able to see the error highlighting. Perhaps, this does more harm than good and the highlighting should be turned off for now?


ilinum:
I turned off highlighting for now until we can do it properly: https://github.com/ilinum/intellij-scala/commit/f0c35a05c4bd9593f292c21744f899e4767f01ae

ilinum:
I ended up fixing the problem with [SCL-2508](https://youtrack.jetbrains.com/issue/SCL-2508): https://github.com/ilinum/intellij-scala/commit/933470863785b84592a059951da5de1457af7687

PR #219, opened 2015-10-05
show canonical text when presentable text is the same

ilinum:
There was a small bug in this commit: https://github.com/JetBrains/intellij-scala/commit/e3e9bf8ae674105497634d06278f31e2a969677e

The calculation of ScTypePresentation was made, but actualType and expectedType were never used. 

The error would appear in the following case:

```
  def foo(b: java.math.BigDecimal*) = ???
  foo(new java.math.BigDecimal(1), BigDecimal(1))
```


ilinum:
@niktrop could you please look at this?

niktrop:
Thanks, I missed it.

PR #241, opened 2015-10-23
Adding 'line comments at first column' option

LPTK:
See: https://youtrack.jetbrains.com/issue/SCL-7898


epishkin:
something is wrong with this pull request. There shouldn't be so many commits

jamesbrain:
This might be an artifact of our plugin recent repository modifications. Whetever the reason for these werid commits (they were not present in the original PR) is, the original PR has already been merged into our repository (with proper credits to LPTK too), so there is no need to keep it open anymore.

epishkin:
thanks @LPTK!

PR #249, opened 2015-10-29
Convert throws to @throws[Exception] instead of @throws(classOf[Exception])

soc:


Alefas:
This change will be added manually as we are in the middle of refactoring of converter code.

soc:
@Alefas Cool, thanks!

I had another four annoying issues identified ... what should I do with them?

Alefas:
@soc Please read https://github.com/JetBrains/intellij-scala/blob/idea15.x/CONTRIBUTING.mkdn
If it's about Java -> Scala convertion then creating issues will have perfect timing as they will be fixed soon.

soc:
@Alefas Here we go:
- SCL-9368 `@throws[Exception]`
- SCL-9369 better constructors
- SCL-9370 no `return`
- SCL-9370 no `simpleMethod = { ... }`
- SCL-9371 wrong equals
- SCL-9372 missing parenthesis around `eq`
- SCL-9373 `@param <T>` ---> `@tparam T`
- SCL-9374 Javadoc to class instead of object
- SCL-9375 array initializer
- SCL-9376 fallthrough not handled

Thanks!

PR #251, opened 2015-10-30
Big performance improvement: Invalidate caches less often

ilinum:
When you're inside a function that has a return type, you don't need to
invalidate all caches, just the caches inside this function
This algorithm is currently implemented in ReSharper but not in IDEA


He-Pin:
I have to say,this really make coding in scala a better enjoy,thank you,the speed is really great now.

PR #259, opened 2015-11-09
Add settings to enable/configure alternate continuation indentation for constructor args and parameter declarations

jpbetz:
This adds an alternate Scala indentation setting for constructor arg and parameter declarations.

The goal is to make it possible to customize code style for the Scala editor to match the [official Scala style guide](http://docs.scala-lang.org/style/).

The relevant rules from the guide are:
- Continuation indent ([Line Wrapping in the Scala style guide](http://docs.scala-lang.org/style/indentation.html#line-wrapping)) should be 2-space.
- Constructor Arguments continuations should 4-space (["...indented four spaces"](http://docs.scala-lang.org/style/declarations.html#classes)).
- (While no guidance is explicitly provided in the official Scala style guide on indentation for continuations of long parameter list declarations, it is often assumed to be 4-space, just like Constructor Arguments are. Conveniently, `ScParameterClause` is used for both.)

Example from official guide:

``` scala
class Person(
    name: String,
    age: Int,
    birthdate: Date,
    astrologicalSign: String,
    shoeSize: Int,
    favoriteColor: java.awt.Color) {
  def firstMethod: Foo = ...
}
```

I looked briefly into how to add the indentation settings to the `Tabs and Indents` Scala Code Style tab, but was unable to figure out a reasonable way to add it. So I have instead added it as a setting to the "Other" page.  If there is a better way to introduce this setting, let me know.

I was also unable to figure out how to respect "Use tab character". So this settings always uses spaces...

I've set this alternate indentation setting to disabled by default.

Here's what the settings UI look like with the new "Alternate indentation for constructor args and parameter declarations" option:

screen shot 2015-11-09 at 5 01 42 pm

### DIFF
--- a/src/org/jetbrains/plugins/scala/lang/formatting/ScalaBlock.scala
+++ b/src/org/jetbrains/plugins/scala/lang/formatting/ScalaBlock.scala
@@ -13,12 +13,13 @@ import org.jetbrains.plugins.scala.lang.formatting.processors._
 import org.jetbrains.plugins.scala.lang.formatting.settings.ScalaCodeStyleSettings
 import org.jetbrains.plugins.scala.lang.lexer.ScalaTokenTypes
 import org.jetbrains.plugins.scala.lang.parser.ScalaElementTypes
+import org.jetbrains.plugins.scala.lang.psi.ScalaPsiUtil
 import org.jetbrains.plugins.scala.lang.psi.api.ScalaFile
-import org.jetbrains.plugins.scala.lang.psi.api.base.ScLiteral
+import org.jetbrains.plugins.scala.lang.psi.api.base.{ScPrimaryConstructor, ScLiteral}
 import org.jetbrains.plugins.scala.lang.psi.api.base.patterns._
 import org.jetbrains.plugins.scala.lang.psi.api.expr._
 import org.jetbrains.plugins.scala.lang.psi.api.expr.xml._
-import org.jetbrains.plugins.scala.lang.psi.api.statements.ScValue
+import org.jetbrains.plugins.scala.lang.psi.api.statements.{ScFunction, ScValue}
 import org.jetbrains.plugins.scala.lang.psi.api.statements.params._
 import org.jetbrains.plugins.scala.lang.psi.api.toplevel.ScEarlyDefinitions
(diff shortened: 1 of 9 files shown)


jpbetz:
This addresses https://youtrack.jetbrains.com/issue/SCL-8313.

jpbetz:
Here's an example of the same code style from the Spark project:

https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/Aggregator.scala

jamesbrain:
@jpbetz Hi, thank you very much for the PR.

I've taken a look into this issue myself a few days ago, and what you provide is quite similar to what I intended to do.However, for the fix to work correctly in all cases, it is required that not Indent.getSpaceIndent(), but rather Indent.getContinuationWithoutFirstIndent() method be used with custom continuation indent length supplied to getContinuationWithoutFirstIndent(). It is not possible right now to do that (the required methods are present in Idea platform, but they have package-access, so there is no way to use them right now without messing up the plugin structure). Probably it is worth using the solution you provide as a temporary fix. I will do some testing and, if everything seems to work fairly good, merge the PR.

jpbetz:
@jamesbrain Great! Yeah, I noticed quite quickly when writing up this change that I was unable to integrate with the indentation system properly, so including it as temporary fix sound right to me.

Let me know if you need anything changed.  And please feel free to commandeer it and modify it to your liking if that's easier for you.

jpbetz:
@jamesbrain We've been using this for a few days with good success with our scala developers here at Coursera on a fairly large Scala codebase. It's looking stable.

Any concerns from your side?

jamesbrain:
@jpbetz I've done some testing myself, and it seems like the only case that gets broken is when you want opening parenthesis on the next line:
With continuation indent set to 4 you will get this:

```
class Foo
(
    foo: Int,
    bar: Int
)
```

However, with alternate indentation active and set to 4 you will get this:

```
class Foo
    (
    foo: Int,
    bar: Int
    )
```

I think there should be a setting for this kind of behavior (there is none right now I believe), but for now it should at leaft be consistent.
I've fixed this and mergred the PR manually. So the next nightly and EAP builds should contain the fix.

Thank you.

PR #269, opened 2015-11-13
[PatternAnnotator]Tuple crushing doesn't happen with synthetic functions

ilinum:
Specifically, with case classes #SCL-8970


He-Pin:
thank you

PR #271, opened 2015-11-16
Do not invalidate caches when inside while loop

ilinum:
If you are inside a while statement, you can only return if you use
return keyword. However, if you use `return`, you have to have return
type declared, so return type of the function cannot change when you
are inside a while loop

### DIFF
--- a/src/org/jetbrains/plugins/scala/lang/psi/api/expr/ScModificationTrackerOwner.scala
+++ b/src/org/jetbrains/plugins/scala/lang/psi/api/expr/ScModificationTrackerOwner.scala
@@ -56,6 +56,7 @@ trait ScModificationTrackerOwner extends ScalaPsiElement with PsiModifiableCodeB
       case v: ScValue => v.typeElement.isDefined
       case v: ScVariable => v.typeElement.isDefined
       case _: ScWhileStmt => true
+      case _: ScFinallyBlock => true
       case _ => false
     }
   }
--- a/src/org/jetbrains/plugins/scala/lang/psi/api/expr/ScModificationTrackerOwner.scala
+++ b/src/org/jetbrains/plugins/scala/lang/psi/api/expr/ScModificationTrackerOwner.scala
@@ -57,6 +57,7 @@ trait ScModificationTrackerOwner extends ScalaPsiElement with PsiModifiableCodeB
       case v: ScVariable => v.typeElement.isDefined
       case _: ScWhileStmt => true
       case _: ScFinallyBlock => true
+      case _: ScDoStmt => true
       case _ => false
     }
   }
--- a/src/org/jetbrains/plugins/scala/lang/psi/api/expr/ScModificationTrackerOwner.scala
+++ b/src/org/jetbrains/plugins/scala/lang/psi/api/expr/ScModificationTrackerOwner.scala
@@ -55,6 +55,7 @@ trait ScModificationTrackerOwner extends ScalaPsiElement with PsiModifiableCodeB
       }
       case v: ScValue => v.typeElement.isDefined
       case v: ScVariable => v.typeElement.isDefined
+      case _: ScWhileStmt => true
       case _ => false
     }
   }


retronym:
I'm not familiar with this infrastructure, but it you might also be able to exclude edits inside a `finally` clause from invalidating the cached type of the enclosing method.

ilinum:
@retronym that's a great idea! If you have any more, please let me know!

retronym:
How does IDEA handle edits in the statement position within a block? 

Perhaps its useful to contrast these examples:

``` scala
def foo = {
  val x = "".to<CURSOR>
  42
}
```

``` scala
def foo = {
  val x = "".to<CURSOR>
  x
}
```

retronym:
It's pretty seldom used, but what's good for `ScWhileStmt` is also good for `ScDoStmt`.

ilinum:
Currently if you're typing within a function with no explicit return type caches will be invalidated. 

So what you're suggesting is when return type of a function did not change, we should not invalidate caches? @Alefas and I have been thinking about it. I think I'm going to try to implement it in the near future and see if it is possible.
Here's how it might work:
1. Do not invalidate caches when you're typing inside a function, even if it has no explicit return type.
2. When we call getType on a function and it returns a different type than before, increment the modification count and restart the highlighting
Is this what you were referring to?

retronym:
Maybe it is easier than that. (Or maybe not, I'm no longer familiar enough with IDEA internals to know...)

I take it the trick here is to syntactically determine whether editing a particular range can change the signature of the enclosing definition.

Recapping my earlier suggestion:

```
def foo = try x finally y
def foo' = try x finally y' // editing y doesn't influence type of foo
```

Coming to an example with blocks:

```
def foo = { s1; e1 }
def foo' = { s1'; e1 }
// if neither s1 nor s1' bind identifiers, the type of foo is unchanged
```

ilinum:
So here's the main limitation: this needs to work extremely fast because a lot of these computations have to be done on the UI thread (for instance, updating the modification count) and can't take long. Typing in IDEA is already somewhat slow; we should not make it slower :)

So for your algorithm, we should: 
1. Find all places where the function returns (it is already implemented somewhere. You know when you put cursor on `def` keyword, it highlights all the places where a function returns?)
2. For every place where return type could've changed:
   - If it's a reference expression, resolve it and see if its declaration was changed. If it did, increment modification count
   - If it's a literal, see if it was changed. If it did, increment modification count. 

I am worried that doing on this on every typed character would be too expensive. We could do that on a non-UI thread, but then we would have to restart the highlighting.

retronym:
I was proposing something different:

For every subtree that is not in a return position, and does not introduce names (is not a `def`, `val`, `var`, `class`, `object`, `trait` or pattern), changes in the subtree need not update the modification counter of the enclosing method. This does not require reference resolving.

Not sure if this is something that you can exploit, or if it is worth it.

retronym:
> You know when you put cursor on def keyword, it highlights all the places where a function returns

[Yep](https://github.com/JetBrains/Intellij-scala/commit/bfd7fc2762c17fbe22d5d35ad7645fa1a89afe06) :wink:

Alefas:
@retronym I thought about your suggestion, but then I understood that in code without vars and side-effects everything will lead to return position or something with declared type.

I just realized even better algorithm (no need of IDEA internals knowledge):
1. Update modification count of function in any case, don't invalidate any other caches.
2. If function has return type => ok, it's already works.
3. If not, then put our function as some global object.
4. If global object is not empty (possible during some refactoring, or fast typing in few places) then old way to invalidate all caches.
Now we need to evaluate this global object by evaluating its type and comparing with previous value.
5. If we want to get caches somewhere, look if we have some global object first. If we have it, then wait, start evaluating global object. If we want to get caches during evaluating global object then don't wait.
6. If global object updates return type, then invalidate caches by changing modification counts. We can move to the next outer function of our possibly local function and do everything from step 1.

PR #272, opened 2015-11-16
[Cache]Remove optional provider + do not invalidate caches in libraries

ilinum:
Should use ModificationTracker.NEVER_CHANGED instead of no dependency


ilinum:
Not invalidating caches libraries does not improve performance (at least not as dramatically as not invalidating caches in definitions with explicit return type). Oh well, I guess it can't hurt

ilinum:
Closed in favor of https://github.com/JetBrains/intellij-scala/pull/281
Not invalidating caches in libraries did not yield any performance improvement

PR #274, opened 2015-11-16
fixed assigning new names to compilation profiles #SCL-9234 fixed

ghik:
Here's a fix for this very annoying bug: https://youtrack.jetbrains.com/issue/SCL-9234


niktrop:
Looks good. Thanks for contribution!

PR #288, opened 2015-12-06
Fixed Scala compiler encoding setting when imported from Gradle

jthurne:
This commit fixes #SCL-9597 (Importing/opening a Gradle Scala project
incorrectly configures the Scala compiler encoding setting).

The fix is to add the "-encoding" and the user-provided encoding value
as separate compiler options in the ScalaGradleDataService. Otherwise,
the encoding setting will ultimately be passed to the compiler as a
qouted combined string (e.g., literally, "-encoding utf-8"). With the
fix in place, the values are correctly passed and parsed by scalac
(e.g., literally "-encoding" "utf-8").


nkly:
Thanks! :+1:

PR #289, opened 2015-12-11
fix loosing comments when removing brackets

katejim:
removing brackets may lead to loosing comments in RemoveBracesIntention (removing curly brackets now available when there is one statement and comments on current line) or in ScalaUnnecessaryParenthesesInspectionBase (removing parenthesis available as previous, but without loosing comments)

### DIFF
--- a/src/org/jetbrains/plugins/scala/codeInsight/intention/IntentionUtil.scala
+++ b/src/org/jetbrains/plugins/scala/codeInsight/intention/IntentionUtil.scala
@@ -1,0 +1,50 @@
+package org.jetbrains.plugins.scala.codeInsight.intention
+
+import com.intellij.openapi.util.text.StringUtil
+import com.intellij.psi.impl.source.codeStyle.CodeEditUtil
+import com.intellij.psi.util.PsiTreeUtil
+import com.intellij.psi.{PsiComment, PsiElement, PsiWhiteSpace}
+import org.jetbrains.plugins.scala.extensions.PsiElementExt
+
+object IntentionUtil {
+  def collectComments(element: PsiElement, onElementLine: Boolean = false): CommentsAroundElement = {
+    def hasLineBreaks(whiteSpace: PsiElement): Boolean = {
+      if (!onElementLine) false
+      else StringUtil.containsLineBreak(whiteSpace.getText)
+    }
+
+    def getElements(it: Iterator[PsiElement]) = {
+      def acceptableElem(elem: PsiElement) = {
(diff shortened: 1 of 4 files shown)


src/org/jetbrains/plugins/scala/codeInsight/intention/IntentionUtil.scala
```diff
@@ -0,0 +1,52 @@
+package org.jetbrains.plugins.scala.codeInsight.intention
+
+import com.intellij.openapi.util.text.StringUtil
+import com.intellij.psi.impl.source.codeStyle.CodeEditUtil
+import com.intellij.psi.util.PsiTreeUtil
+import com.intellij.psi.{PsiWhiteSpace, PsiComment, PsiElement}
+
+import scala.collection.mutable.ArrayBuffer
+
+object IntentionUtil {
```
niktrop, line 9: I think you can create a case class instead of such return type.

src/org/jetbrains/plugins/scala/codeInsight/intention/IntentionUtil.scala
```diff
@@ -0,0 +1,52 @@
...
+  def collectComments(element: PsiElement, onElementLine: Boolean = false): (Seq[PsiElement], Seq[PsiElement]) = {
+    def hasLineBreaks(whiteSpace: PsiElement): Boolean = {
+      if (!onElementLine) false
+      else StringUtil.containsLineBreak(whiteSpace.getText)
+    }
+
```
niktrop, line 15: It can be made more idiomatic with extension methods PsiElementExtTrait.prevSiblings and PsiElementExtTrait.nextSiblings.

src/org/jetbrains/plugins/scala/codeInsight/intention/IntentionUtil.scala
```diff
@@ -0,0 +1,63 @@
...
+import scala.collection.mutable.ArrayBuffer
+
+object IntentionUtil {
+  def collectComments(element: PsiElement, onElementLine: Boolean = false): CommentsAroundElement = {
+    def hasLineBreaks(whiteSpace: PsiElement): Boolean = {
+      if (!onElementLine) false
+      else StringUtil.containsLineBreak(whiteSpace.getText)
+    }
+
+    def getElements(newElement: (PsiElement) => Option[PsiElement]): Seq[PsiElement] = {
```
Alefas: It's definitely tailrec function. Just immutable collection instead of ArrayBuffer can be used.

PR #290, opened 2015-12-11
Toggle Type Annotation: offer a list of available types #SCL-5043 Fixed

ilinum:
This is how it will like after this PR:
toggle-type-new


ilinum:
It is unnecessary to offer so many types. Need to offer types based on context. Closing this PR and starting to work on context-aware type suggestion

PR #295, opened 2015-12-14
[TypeInference]Default expression does not always affect type inference

ilinum:
# SCL-9555 Fixed


ilinum:
This solution does not work and at this point it is not clear how to make it work

PR #297, opened 2015-12-21
Fix cache invalidation bug: use Queue instead of ArrayBuffer

ilinum:


OndrejSpanel:
Just an inquiry: what happened to this and other pull request from ilunum? There are still marked as open, some of them seem interesting or perhaps even important.

ilinum:
@OndrejSpanel I am still alive if that's what you're asking :)
I'm just not JB employee anymore. My internship ended and I went back to get my degree at the university.

I think @Alefas just didn't get around to looking at them

Alefas:
I'll merge them in February.

PR #298, opened 2015-12-23
Toggle type improvements

ilinum:


ilinum:
@niktrop are you talking about [this](https://github.com/ilinum/intellij-scala/blob/toggle-type-second-attempt/src/org/jetbrains/plugins/scala/codeInsight/intention/types/UpdateStrategy.scala#L148) line?

I'm not filtering them out, they are a special case, I want to treat them separately

niktrop:
In this commit https://github.com/ilinum/intellij-scala/commit/79fe915045d855491596c50f8ab75b404cc9085f you added `.filter(_.isInstanceOf[ScClass])` call. Why do you need it? As I can see case of sealed traits is not handled anywhere.

niktrop:
Merged manually into idea16.x. Also I refactored it a bit, please take a look:
https://github.com/JetBrains/intellij-scala/commit/fd88ae745ee5c09a82311a681afdaef3b220df85

ilinum:
@niktrop Thanks!

PR #300, opened 2015-12-25
Fix broken test

ilinum:
It was broken because implementation of cached annotation was changed
@mutcianm


Alefas:
Merged manually to idea16.x

PR #301, opened 2015-12-28
[Scala->Java]Convert type parameters properly #SCL-5852 Fixed

ilinum:


Alefas:
Merged manually to idea16.x

PR #302, opened 2015-12-29
[Scala->Java]Overriding non-static method from companion object

ilinum:
Need to track names of function wrappers with default arguments


Alefas:
Merged manually to idea16.x

PR #303, opened 2015-12-29
[Scala->Java]Call-by name params cannot have primitives #SCL-8823 Fixed

ilinum:


Alefas:
Merged manually to idea16.x

PR #310, opened 2016-02-01
Sort completion list

katejim:
treate caseClassLabel as local variable.
separate byType weigher from kind weigher.
return lift upper current class members instead ancestor members,
prefer local variable  then named params
add completion for case class params
fix SCL-2022
move deprecated down to suggestion list


src/org/jetbrains/plugins/scala/lang/completion/ScalaCaseClassParametersNameContributer.scala
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    def getCorrespondedParameterForPosition(position: PsiElement, classParams: Seq[ScParameter]): (ScParameter, Int) = {
+      val me = PsiTreeUtil.getContextOfType(position, classOf[ScPattern])
+      if (me == null) return (null, -1)
+
+      val patterns = Option(PsiTreeUtil.getContextOfType(position, classOf[ScPatternArgumentList])).map(_.patterns)
+
+      if (patterns.isEmpty || (patterns.isDefined && patterns.get.length > classParams.length))
+        return (null, -1) //try to type more param than can be
```
Alefas, line 93: Why null and not Option?

src/org/jetbrains/plugins/scala/lang/completion/ScalaCaseClassParametersNameContributer.scala
```diff
@@ -0,0 +1,102 @@
...
+
+      if (caseClassParams.isEmpty) return
+
+      val (corespondedParameter, myPosition) = getCorrespondedParameterForPosition(position, caseClassParams)
+      val result = addByOrderSorter(parameters, _result, myPosition, caseClassParams)
+
+      byClassParamCompletionsItems(caseClassParams, result)
+      byTypeCompletionsItems(position, corespondedParameter, result)
+
+      result.stopHere()
```
Alefas, line 45: Why we need to stop here?

katejim, line 45: In this completion  we don't want data from other contributors(like kewordContributer...)

PR #312, opened 2016-02-08
[Inspections] Hash code uses a var

itegulov:


niktrop:
Merged manually into idea16.x branch. 
Also I have changed the visitor to check inner classes: 
https://github.com/JetBrains/intellij-scala/commit/9bb34bcfb48e8eb0de5cc6d9da223252811b4b8d
You wrote a test for this situation but it was green only due to a bug in the visitClass() method.

PR #313, opened 2016-02-09
Scala completion statistic

katejim:


src/org/jetbrains/plugins/scala/lang/completion/statistician/ScalaCompletionStatistician.scala
```diff
@@ -11,6 +13,26 @@ import com.intellij.psi.statistics.StatisticsInfo
...
+  }
+
+  def helper(element: PsiNamedElement, location: CompletionLocation): StatisticsInfo = {
+    element match {
+      case member: PsiMember =>
+        val key = ScalaStatisticManager.memberKey(member).getOrElse(return StatisticsInfo.EMPTY)
+        val containingClass = member.getContainingClass
+        if (containingClass != null) {
+          val context = ScalaStatisticManager.memberKey(containingClass)
+            .getOrElse(return new StatisticsInfo("scalaMember#", key))
```
Alefas, line 31: It's better to do it without return, just getOrElse("scalaMember#"). One of the reasons is performance as return here in by-name-parameter position.

src/org/jetbrains/plugins/scala/lang/completion/statistician/ScalaCompletionStatistician.scala
```diff
@@ -11,6 +13,26 @@ import com.intellij.psi.statistics.StatisticsInfo
...
+          return new StatisticsInfo(context, key)
+        }
+        new StatisticsInfo("scalaMember#", key)
```
Alefas, line 34: The second reason is that you can avoid return at all (what is good) just put "else" here.

PR #320, opened 2016-03-01
[Sort completion list] fix SCL-9898. fix bold lookup. 

katejim:


src/org/jetbrains/plugins/scala/lang/completion/weighter/ScalaByTypeWeigher.scala
```diff
@@ -3,54 +3,133 @@ package org.jetbrains.plugins.scala.lang.completion.weighter
...
         case s: ScalaLookupItem =>
+          lazy val byTextResult = handleByText(s.element)
           s.element match {
-            case ta: ScTypeAlias if ta.isLocal => localType
-            case ta: ScTypeAlias => typeDefinition
-            case te: ScTypeDefinition if !te.isObject && (te.isLocal || inFunction(te)) => localType
-            case te: ScTypeDefinition => typeDefinition
-            case te: PsiClass => typeDefinition
-            case _ => normal
+            case (_: ScTypeAlias) | (_: ScTypeDefinition) | (_: PsiClass) if byTextResult.isDefined => byTextResult
```
Alefas: Parentheses are not necessary here.

PR #323, opened 2016-03-02
Minor fix in org.jetbrains.plugins.scala.Version

pawelprazak:
Fixed a small bug in org.jetbrains.plugins.scala.Version <= method


Alefas:
Thank you!

pawelprazak:
No problem :-)

PR #325, opened 2016-03-15
SCL-9861 Inspection: unnecessary partial function definition

whorbowicz:
It's up for discussion whether the scope of this task is complete - @pavelfatin was not very responsive recently. Some more conversions could be done in quick fix but I am not sure if they should.

Since I wanted to reuse part of ConvertToParenthesesIntention code I moved it to ScalaPsiUtils and did some cleanup, added few tests. I also reported recurring issue with hardcoded messages in YouTrack.

### DIFF
--- a/src/org/jetbrains/plugins/scala/codeInspection/functionExpressions/UnnecessaryPartialFunctionInspection.scala
+++ b/src/org/jetbrains/plugins/scala/codeInspection/functionExpressions/UnnecessaryPartialFunctionInspection.scala
@@ -1,7 +1,7 @@
 package org.jetbrains.plugins.scala.codeInspection.functionExpressions
 
 import com.intellij.codeInspection.{ProblemHighlightType, ProblemsHolder}
-import com.intellij.psi.{PsiClass, PsiElement}
+import com.intellij.psi.{PsiClass, PsiElement, PsiFile}
 import org.jetbrains.plugins.scala.codeInspection.functionExpressions.UnnecessaryPartialFunctionInspection._
 import org.jetbrains.plugins.scala.codeInspection.{AbstractInspection, InspectionBundle}
 import org.jetbrains.plugins.scala.lang.psi.api.base.patterns._
@@ -22,9 +22,9 @@ class UnnecessaryPartialFunctionInspection
   override def actionFor(holder: ProblemsHolder): PartialFunction[PsiElement, Any] = {
     case expression: ScBlockExpr =>
       def isNotPartialFunction(expectedType: ScType) =
-        findPartialFunctionType(holder).exists(!expectedType.conforms(_))
+        findPartialFunctionType(holder.getFile).exists(!expectedType.conforms(_))
       def conformsTo(expectedType: ScType) = (inputType: ScType, resultType: ScType) =>
-        findType(holder, Function1ClassName, _ => Seq(inputType, resultType)).exists(_.conforms(expectedType))
+        findType(holder.getFile, Function1ClassName, _ => Seq(inputType, resultType)).exists(_.conforms(expectedType))
 
       for{
         expectedExpressionType <- expression.expectedType()
@@ -36,19 +36,19 @@ class UnnecessaryPartialFunctionInspection
           caseKeyword,
           inspectionName,
           ProblemHighlightType.GENERIC_ERROR_OR_WARNING,
-          new UnnecessaryPartialFunctionQuickFix(singleCaseClause, expression))
+          new UnnecessaryPartialFunctionQuickFix(expression))
   }
 
-  private def findType(holder: ProblemsHolder, className: String, parameterTypes: PsiClass => Seq[ScType]): Option[ValueType] =
+  private def findType(file: PsiFile, className: String, parameterTypes: PsiClass => Seq[ScType]): Option[ValueType] =
     ScalaPsiManager
-      .instance(holder.getProject)
-      .getCachedClass(holder.getFile.getResolveScope, className)
+      .instance(file.getProject)
+      .getCachedClass(file.getResolveScope, className)
         .map(clazz =>
           ScParameterizedType(ScDesignatorType(clazz), parameterTypes(clazz)))
 
 
-  private def findPartialFunctionType(holder: ProblemsHolder): Option[ValueType] =
-    findType(holder, PartialFunctionClassName, undefinedTypeParameters)
+  private def findPartialFunctionType(file: PsiFile): Option[ValueType] =
+    findType(file, PartialFunctionClassName, undefinedTypeParameters)
 
   private def undefinedTypeParameters(clazz: PsiClass): Seq[ScUndefinedType] =
     clazz
@@ -62,10 +62,13 @@ class UnnecessaryPartialFunctionInspection
       caseClause.pattern.exists {
         case reference: ScReferencePattern => true
         case wildcard: ScWildcardPattern => true
-        case typed: ScTypedPattern =>
-          typed.typePattern.map(_.typeElement.calcType).exists(inputType =>
-            caseClause.expr.flatMap(_.expectedType()).exists(returnType =>
-                conformsToExpectedType(inputType, returnType)))
+        case typedPattern: ScTypedPattern =>
+          val patternType = typedPattern.typePattern.map(_.typeElement.calcType)
+          val expressionType = caseClause.expr.flatMap(_.getType().toOption)
+          (patternType, expressionType) match {
+            case (Some(inputType), Some(returnType)) => conformsToExpectedType(inputType, returnType)
+            case _ => false
+          }
         case _ => false
       }
 }
--- a/src/org/jetbrains/plugins/scala/codeInspection/functionExpressions/UnnecessaryPartialFunctionInspection.scala
+++ b/src/org/jetbrains/plugins/scala/codeInspection/functionExpressions/UnnecessaryPartialFunctionInspection.scala
@@ -1,0 +1,71 @@
+package org.jetbrains.plugins.scala.codeInspection.functionExpressions
+
+import com.intellij.codeInspection.{ProblemHighlightType, ProblemsHolder}
+import com.intellij.psi.{PsiClass, PsiElement}
+import org.jetbrains.plugins.scala.codeInspection.functionExpressions.UnnecessaryPartialFunctionInspection._
+import org.jetbrains.plugins.scala.codeInspection.{AbstractInspection, InspectionBundle}
+import org.jetbrains.plugins.scala.lang.psi.api.base.patterns._
+import org.jetbrains.plugins.scala.lang.psi.api.expr.ScBlockExpr
+import org.jetbrains.plugins.scala.lang.psi.impl.ScalaPsiManager
+import org.jetbrains.plugins.scala.lang.psi.types.{ScTypeParameterType, _}
+
+object UnnecessaryPartialFunctionInspection {
+  private val inspectionId = "UnnecessaryPartialFunction"
+  private val PartialFunctionClassName = classOf[PartialFunction[_, _]].getCanonicalName
+  private val Function1ClassName = classOf[(_) => _].getCanonicalName
+  val inspectionName = InspectionBundle.message("unnecessary.partial.function")
+}
+
+class UnnecessaryPartialFunctionInspection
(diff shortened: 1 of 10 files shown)


src/org/jetbrains/plugins/scala/codeInspection/functionExpressions/UnnecessaryPartialFunctionInspection.scala
```diff
@@ -0,0 +1,71 @@
...
+      .map(typeParameter =>
+        new ScUndefinedType(new ScTypeParameterType(typeParameter, ScSubstitutor.empty)))
+      .toSeq
+
+  private def canBeConvertedToFunction(caseClause: ScCaseClause, conformsToExpectedType: (ScType, ScType) => Boolean) =
+    caseClause.guard.isEmpty &&
+      caseClause.pattern.exists {
+        case reference: ScReferencePattern => true
+        case wildcard: ScWildcardPattern => true
+        case typed: ScTypedPattern =>
```
whorbowicz: I am not 100% sure if types taken for check in this case are correct. Please review carefully.

src/org/jetbrains/plugins/scala/codeInspection/functionExpressions/UnnecessaryPartialFunctionInspection.scala
```diff
@@ -0,0 +1,71 @@
...
+        Seq(singleCaseClause) <- expression.caseClauses.map(_.caseClauses)
+        if canBeConvertedToFunction(singleCaseClause, conformsTo(expectedExpressionType))
+        caseKeyword <- singleCaseClause.firstChild
+      } holder.registerProblem(
+          caseKeyword,
+          inspectionName,
+          ProblemHighlightType.GENERIC_ERROR_OR_WARNING,
+          new UnnecessaryPartialFunctionQuickFix(singleCaseClause, expression))
+  }
+
```
niktrop, line 41: `ProblemsHolder` is intended for passing problems around. It looks strange as an argument of methods like `findType`. You can pass a `file` for example, it has `getProject` method too.

src/org/jetbrains/plugins/scala/codeInspection/functionExpressions/UnnecessaryPartialFunctionInspection.scala
niktrop: It is correct, but it's hard to read. I would extract a couple variables and replace nested exists with a pattern match on a tuple, but it's up to you.

src/org/jetbrains/plugins/scala/codeInspection/functionExpressions/UnnecessaryPartialFunctionQuickFix.scala
```diff
@@ -0,0 +1,54 @@
...
+import org.jetbrains.plugins.scala.lang.psi.impl.ScalaPsiElementFactory
+
+object UnnecessaryPartialFunctionQuickFix {
+  val hint = InspectionBundle.message("convert.to.anonymous.function")
+}
+
+class UnnecessaryPartialFunctionQuickFix(caseClause: ScCaseClause, expression: ScBlockExpr)
+  extends AbstractFixOnPsiElement(hint, caseClause) {
+
+  override def doApplyFix(project: Project): Unit = {
```
niktrop, line 20: When you extend `AbstractFixOnPsiElement` you shouldn't access constructor parameters directly, they may be invalidated by the time the quickfix is invoked. It probably should be better documented, I'll fix it. You can get rid of `caseClause` parameter and use `getElement` to get `expression`. Also you need to check again that expression.caseClauses has a single case.

src/org/jetbrains/plugins/scala/codeInspection/functionExpressions/UnnecessaryPartialFunctionQuickFix.scala
```diff
@@ -0,0 +1,54 @@
...
+  private def canConvertBraces(clause: ScCaseClause): Boolean =
+    patternIsReferenceOrWildcard(clause) && bodyIsExpressionOrOneStatementBlock(clause)
+
+  private def patternIsReferenceOrWildcard(clause: ScCaseClause): Boolean =
+    clause.pattern.exists {
+      case reference: ScReferencePattern => true
+      case wildcard: ScWildcardPattern => true
+      case _ => false
+    }
+
```
niktrop, line 65: I would suggest to replace braces with parentheses only if case clause takes one line.

src/org/jetbrains/plugins/scala/codeInspection/functionExpressions/UnnecessaryPartialFunctionQuickFix.scala
```diff
@@ -0,0 +1,54 @@
...
+class UnnecessaryPartialFunctionQuickFix(caseClause: ScCaseClause, expression: ScBlockExpr)
+  extends AbstractFixOnPsiElement(hint, caseClause) {
+
+  override def doApplyFix(project: Project): Unit = {
+    val expressionCopy = expression.copy().asInstanceOf[ScBlockExpr]
+    expressionCopy.caseClauses.map(_.caseClauses).foreach {
+      case Seq(clause) =>
+        removeCaseKeyword(clause)
+        if(canConvertBraces(clause)) ScalaPsiUtil.replaceBracesWithParentheses(expressionCopy)
```
niktrop:
```
list.map {
  case i => i + 1
}
```

should be transformed into
`list.map(i => i + 1)`.
You need to remove new lines yourself, our postformatter fixes indents and single line whitespaces only.

src/org/jetbrains/plugins/scala/codeInspection/functionExpressions/UnnecessaryPartialFunctionInspection.scala
whorbowicz: I agree that it can be hard to read. Can you elaborate a bit more about your suggestion? Where exactly you want to pattern match on a tuple?

src/org/jetbrains/plugins/scala/codeInspection/functionExpressions/UnnecessaryPartialFunctionInspection.scala
```diff
@@ -0,0 +1,71 @@
...
+      .toSeq
+
+  private def canBeConvertedToFunction(caseClause: ScCaseClause, conformsToExpectedType: (ScType, ScType) => Boolean) =
+    caseClause.guard.isEmpty &&
+      caseClause.pattern.exists {
+        case reference: ScReferencePattern => true
+        case wildcard: ScWildcardPattern => true
+        case typed: ScTypedPattern =>
+          typed.typePattern.map(_.typeElement.calcType).exists(inputType =>
+            caseClause.expr.flatMap(_.expectedType()).exists(returnType =>
```
whorbowicz: @niktrop I used `expectedType` as a return type since I am more interested in input type in this check. How to obtain actual return type of an expression? `expectedType` seem to be related to the code that accepts expression's value.

src/org/jetbrains/plugins/scala/codeInspection/functionExpressions/UnnecessaryPartialFunctionQuickFix.scala
whorbowicz, line 20: So class declaration should look like this?

```
class UnnecessaryPartialFunctionQuickFix(expression: ScBlockExpr)
  extends AbstractFixOnPsiElement(hint, expression)
```

From what I was able to understand, element passed as constructor argument to AbstractFixOnPsiElement was the element on which fix was available. Is that correct?

whorbowicz, line 20: The check for single case clause is done in

```
case Seq(clause) =>
```

Which matches only when `caseClauses` returns sequence.with one element. I will rename value to singleClause to outline that a bit more

src/org/jetbrains/plugins/scala/codeInspection/functionExpressions/UnnecessaryPartialFunctionQuickFix.scala
whorbowicz, line 65: How I can check that?

```
val x : Int => String = {case x => val foo = x + 2; foo.toString} 
```

takes one line but it's not legal to convert => multi-statement block
Actually it looks like body of case clause always evaluates to ScBlock even for cases where i thought it would go along expression branch like:

```
val x : Int => String = {case x => x.toString}
```

src/org/jetbrains/plugins/scala/codeInspection/functionExpressions/UnnecessaryPartialFunctionQuickFix.scala
niktrop, line 20: I understand that. It is about lifecycle of a quickfix. If another case clause is added and quickfix is invoked before the file gets rehighlighted, it will throw a MatchError.

niktrop, line 20: And it can take several seconds to rehighlight big files.

src/org/jetbrains/plugins/scala/codeInspection/functionExpressions/UnnecessaryPartialFunctionQuickFix.scala
niktrop, line 65: You are right, our parser works that way. `ScBlockExpr` is `ScBlock` with case clauses inside, probably not the best name. So, the first case may be removed. You need to check `block.statements.size == 1` as before and add something like `!block.getText.contains('\n')`

src/org/jetbrains/plugins/scala/codeInspection/functionExpressions/UnnecessaryPartialFunctionInspection.scala
niktrop:
```
val patternType = typed.typePattern.map(_.typeElement.calcType)
val exprType = caseClause.expr.flatMap(_.getType().toOption)
(patternType, exprType) match {
  case (Some(inputType), Some(returnType)) =>
      conformsToExpectedType(inputType, returnType)
  case _ => false
}
```

src/org/jetbrains/plugins/scala/codeInspection/functionExpressions/UnnecessaryPartialFunctionInspection.scala
niktrop: `expr.getType().toOption`

src/org/jetbrains/plugins/scala/codeInspection/functionExpressions/UnnecessaryPartialFunctionQuickFix.scala
whorbowicz, line 65: It doesn't seem intuitive to me. What about multi-line expressions that were broken into lines for clarity? eg.

```
val f: Int => String = {
  case x => List.fill(5)(x)
    .zipWithIndex
    .toString()
}
```

niktrop, line 65: As it spans for multiple lines it remains in braces:

```
val f: Int => String = {
  x => List.fill(5)(x)
    .zipWithIndex
    .toString()
}
```

Quickfix shouldn't change formatting too much.

whorbowicz:
@niktrop Any update on your side? Fixes were added few days ago.

niktrop:
Hi Waldemar,
Sorry for a long delay. We have a busy time cleaning up our issue tracker, it seems that I missed the notification about your last commit. All is good, I merged it manually with couple of minor fixes: https://github.com/JetBrains/intellij-scala/commit/d59b743095d495e615e857aa1c18b6475ea326ee.
Also I have added a comment about `AbstractFixOnPsiElement`: https://github.com/JetBrains/intellij-scala/commit/3386e3fefab9ca548ac0addb7918da8184d1035e, hope it makes things a little bit clearer.

If you are interested in improving this inspection futher you may consider the case when partial function with a tuple pattern may be replaced with FunctionN: [SCL-8824](https://youtrack.jetbrains.com/issue/SCL-8824)

Thank you for contribution! Your set of tests is very well, too.

PR #329, opened 2016-03-31
SCL-5456 SCL-9514 Move a class with companion object fails

jeantil:
I created a test case for the issue and named it after the earliest bug report and made sure I didn't break any other refactoring related tests. I do get many test failures locally but I don't see how they can be related to my change. 

I was pretty conservative in my changes and would like a reviewer's opinion on whether I should remove ScalaMoveClassesOrPackagesProcessor.scala entirely or leave it as it is. I couldn't tell if the remaining code was still useful ...


src/org/jetbrains/plugins/scala/lang/refactoring/move/MoveScalaClassHandler.java
```diff
@@ -16,6 +16,7 @@
 import org.jetbrains.plugins.scala.lang.psi.api.ScalaFile;
 import org.jetbrains.plugins.scala.lang.psi.api.toplevel.ScNamedElement;
 import org.jetbrains.plugins.scala.settings.ScalaApplicationSettings;
+import scalariform.formatter.ScalaFormatter;
```
jeantil: I'll remove this asap, it shouldn't have been committed

src/org/jetbrains/plugins/scala/lang/refactoring/move/ScalaMoveUtil.scala
```diff
@@ -44,6 +44,8 @@ object ScalaMoveUtil {
 
   def doMoveClass(@NotNull aClass: PsiClass, @NotNull moveDestination: PsiDirectory, withCompanion: Boolean): PsiClass = {
     var fileWasDeleted: Boolean = false
+    val maybeCompanion = if(withCompanion) ScalaPsiUtil.getBaseCompanionModule(aClass) else None
```
jeantil, line 47: need to extract this here before anything has actually been moved so we can use it once the class is moved

src/org/jetbrains/plugins/scala/lang/refactoring/move/ScalaMoveUtil.scala
```diff
@@ -97,8 +101,15 @@ object ScalaMoveUtil {
       newClass
     }
 
-    if (withCompanion) ScalaPsiUtil.getBaseCompanionModule(aClass).foreach(c => moveClassInner(c, moveDestination))
```
jeantil, line 100: The original error came from this line which would create the target file. When moveClassInner was called at the next line, it considered that moving the class to the newly created file would result in creating a fakeCompanion as per `ScalaMoveUtil#canBeCompanions`. moving these lines around wasn't enough though as the expandedElements in `ScalaMoveClassesOrPackagesProcessor` ended up trying to execute the move handler twice in an uncoordinated manner triggering an erroneous fakeCompanion detection again.

src/org/jetbrains/plugins/scala/lang/refactoring/move/MoveScalaClassHandler.java
```diff
@@ -16,6 +16,7 @@
 import org.jetbrains.plugins.scala.lang.psi.api.ScalaFile;
 import org.jetbrains.plugins.scala.lang.psi.api.toplevel.ScNamedElement;
 import org.jetbrains.plugins.scala.settings.ScalaApplicationSettings;
```
niktrop, line 18: Very strange unused import.

jeantil, line 18: Sorry about that, I updated the commit to remove the unused ScalaFormatter.  The `Move` action leaves empty line in the source file. 
I considered removing them by reformatting the source file after the move. However a quick check showed that leaving empty lines for removed elements was the current behaviour. I stopped trying to add reformat but forgot to remove the import.

src/org/jetbrains/plugins/scala/lang/refactoring/move/ScalaMoveClassesOrPackagesProcessor.scala
```diff
@@ -21,13 +21,7 @@ class ScalaMoveClassesOrPackagesProcessor(project: Project,
                                           searchInNonJavaFiles: Boolean,
                                           moveCallback: MoveCallback) extends
 {
```
niktrop: You are right, this class can be removed now.

niktrop:
You can write `#SCL-XXXX fixed` on a new line of the commit message to automatically close tickets on Youtrack.

jeantil:
I updated the commit to remove the ScalaMoveClassesOrPackagesProcessor
I also added the markers to fix both issue

niktrop:
Thanks for contributing once more!

Alefas:
I'm not sure if YouTrack close ticket with commits of user, which do not have such access. But in any case it's good advice as such commit message will be collected by our release notes.

PR #332, opened 2016-04-03
#SCL-8312 fixed check for .flatten emulation

lpiepiora:
This is a PR for https://youtrack.jetbrains.com/issue/SCL-8312, as discussed here: https://youtrack.jetbrains.com/issue/SCL-9562.

### DIFF
--- a/src/org/jetbrains/plugins/scala/codeInspection/collections/EmulateFlattenInspection.scala
+++ b/src/org/jetbrains/plugins/scala/codeInspection/collections/EmulateFlattenInspection.scala
@@ -1,0 +1,43 @@
+package org.jetbrains.plugins.scala.codeInspection.collections
+
+import org.jetbrains.plugins.scala.codeInspection.InspectionBundle
+import org.jetbrains.plugins.scala.extensions.ResolvesTo
+import org.jetbrains.plugins.scala.lang.psi.api.expr._
+
+/**
+  * @author Lukasz Piepiora
+  */
+class EmulateFlattenInspection extends OperationOnCollectionInspection {
+  override def possibleSimplificationTypes: Array[SimplificationType] =
+    Array(FlattenSimplification)
+}
+
+object FlattenSimplification extends SimplificationType {
+
+  override def hint: String = InspectionBundle.message("replace.with.flatten")
+
+  override def getSimplification(expr: ScExpression): Option[Simplification] =
+    expr match {
+      case seqOfSeqs `.flatMap` (identityOperation()) =>
+        Some(replace(expr).withText(invocationText(seqOfSeqs, "flatten")))
+      case _ => None
+    }
+
+  private object identityOperation {
+    def unapply(expr: ScExpression): Boolean = stripped(expr) match {
+      case _ `.identity` (underscore()) => true
+      case _ `.identity` () => true
+      case undSect: ScUnderscoreSection =>
+        undSect.bindingExpr match {
+          case Some(_ `.identity` ()) => true
+          case _ => false
+        }
+      case ScFunctionExpr(Seq(x), Some(ResolvesTo(param))) if x == param => true
+      case ScFunctionExpr(Seq(x), Some(`.identity`(_, ResolvesTo(param)))) if x == param => true
+      case _ => false
+    }
+  }
+
+  private val `.identity` = invocation("identity").from(Array("scala.Predef"))
+
+}
--- a/resources/META-INF/scala-plugin-common.xml
+++ b/resources/META-INF/scala-plugin-common.xml
@@ -1091,6 +1091,11 @@
                          displayName="Replace to with until" groupPath="Scala, Collections" groupName="Simplifications: other"
                          shortName="ReplaceToWithUntil" level="WARNING"
                          enabledByDefault="true" language="Scala"/>
+        <localInspection implementationClass="org.jetbrains.plugins.scala.codeInspection.collections.EmulateFlattenInspection"
+                         displayName="Replace with flatten" groupPath="Scala, Collections" groupName="Simplifications: other"
+                         shortName="ReplaceWithFlatten" level="WARNING"
+                         enabledByDefault="true" language="Scala"/>
+
         <!--maps-->
         <localInspection implementationClass="org.jetbrains.plugins.scala.codeInspection.collections.GetGetOrElseInspection"
                          displayName="Get and getOrElse to getOrElse" groupPath="Scala, Collections" groupName="Maps"
--- a/resources/inspectionDescriptions/ReplaceWithFlatten.html
+++ b/resources/inspectionDescriptions/ReplaceWithFlatten.html
@@ -1,0 +1,17 @@
+<html>
+<body>
+Replaces code, which emulates <code>.flatten</code> method on collections of collections.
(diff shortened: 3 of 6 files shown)


src/org/jetbrains/plugins/scala/codeInspection/collections/EmulateFlattenInspection.scala
```diff
@@ -0,0 +1,55 @@
...
+      case identity(underscore()) => true
+      case identity() => true
+      case ScFunctionExpr(Seq(x), Some(ResolvesTo(param))) if x == param => true
+      case ScFunctionExpr(Seq(x), Some(identity(ResolvesTo(param)))) if x == param => true
+      case _ => false
+    }
+  }
+
+  private object identity {
+
```
niktrop: Method `identity` may also be used with type argument or in the form of postfix expression. You may create an invocation template for `identity` method from `scala.Predef`, like examples in the package object show. It will handle all such cases. It's `unapplySeq` method match qualifier as a first subpattern, you'll just need to ignore it.

lpiepiora: @niktrop Thanks for the tip! I've changed this, and added couple of tests for the new cases, which you pointed out wouldn't work in the original setup. However I have noticed something pretty strange.

Consider this test case:

```scala
  def testSuggests10(): Unit = {
    val selected = s"Set(Set(1), Set(2), Set(3)).${START}flatMap(identity _)$END"
    checkTextHasError(selected, hint, inspectionClass)
    val text = "Set(Set(1), Set(2), Set(3)).flatMap(identity _)"
    val result = "Set(Set(1), Set(2), Set(3)).flatten"
    testFix(text, result, hint)
  }
```

Which is pattern matched by this match (not exactly how I coded, but this is easier to see):

```scala
case seqOfSeqs `.flatMap` (underscore()) =>
```

Where I would expect it to be matched by something like this:

```scala
case seqOfSeqs `.flatMap` (`.identity`(_, underscore())) =>
```

Is my expectation incorrect, or there is some problem with the code. If it's the latter I would like to fix it, but I don't really know where, either PSI is incorrectly generated or the InvocationTemplate is parsing it wrong I guess. I think the PSI is incorrect but I'm not sure.

If I have 

```scala
flatMap(identity(_))
```

The PSI would be like this:

screenshot 2016-04-04 21 22 00

Which is more or less logical especially in regards to `UnderscoreSection` which wraps `_`, and is placed in the argument list.

However when I have code like this:

```scala
flatMap(identity _)
```

The PSI will look like this:

screenshot 2016-04-04 21 25 32

Here the `UnderscoreSection spans from one parenthesis to another, and the identity call is wrapped in the `UnderscoreSection`. Not sure if this is correct, but if so, then this check:

```scala
object underscore {
    def unapply(expr: ScExpression): Boolean = {
      stripped(expr) match {
        case ScParenthesisedExpr(underscore()) => true
        case typed: ScTypedStmt if typed.expr.isInstanceOf[ScUnderscoreSection] => true
        case und: ScUnderscoreSection => true
        case _ => false
      }
    }
  }
```

Seems to be too relaxed in the case `case und: ScUnderscoreSection => true`

Is my thinking correct?

lpiepiora:
@niktrop Thanks for the tip! I've changed this, and added couple of tests for the new cases, which you pointed out wouldn't work in the original setup. However I have noticed something pretty strange.

Consider this test case:

``` scala
  def testSuggests10(): Unit = {
    val selected = s"Set(Set(1), Set(2), Set(3)).${START}flatMap(identity _)$END"
    checkTextHasError(selected, hint, inspectionClass)
    val text = "Set(Set(1), Set(2), Set(3)).flatMap(identity _)"
    val result = "Set(Set(1), Set(2), Set(3)).flatten"
    testFix(text, result, hint)
  }
```

Which is pattern matched by this match (not exactly how I coded, but this is easier to see):

``` scala
case seqOfSeqs `.flatMap` (underscore()) =>
```

Where I would expect it to be matched by something like this:

``` scala
case seqOfSeqs `.flatMap` (`.identity`(_, underscore())) =>
```

Is my expectation incorrect, or there is some problem with the code. If it's the latter I would like to fix it, but I don't really know where, either PSI is incorrectly generated or the InvocationTemplate is parsing it wrong I guess. I think the PSI is incorrect but I'm not sure.

If I have 

``` scala
flatMap(identity(_))
```

The PSI would be like this:

screenshot 2016-04-04 21 22 00

Which is more or less logical especially in regards to `UnderscoreSection` which wraps `_`, and is placed in the argument list.

However when I have code like this:

``` scala
flatMap(identity _)
```

The PSI will look like this:

screenshot 2016-04-04 21 25 32

Here the `UnderscoreSection` spans from one parenthesis to another, and the identity call is wrapped in the `UnderscoreSection`. Not sure if this is correct, but if so, then this check:

``` scala
object underscore {
    def unapply(expr: ScExpression): Boolean = {
      stripped(expr) match {
        case ScParenthesisedExpr(underscore()) => true
        case typed: ScTypedStmt if typed.expr.isInstanceOf[ScUnderscoreSection] => true
        case und: ScUnderscoreSection => true
        case _ => false
      }
    }
  }
```

Seems to be too relaxed in the case `case und: ScUnderscoreSection => true`

Is my thinking correct?

niktrop:
Good catch! It's too relaxed indeed. 

`ScUnderscoreSection` is generated by our parser in two cases. First is placeholder for a parameter in the short form of function expression (like in `seq.map(_ + 1)`) and second is a method value. Probably it would be better to separate them to different classes. But now you can distinguish them with `bindingExpr` method. It returns reference to the method in the case of method value.

You can check extractor for a function expression `contains` in `FilterOtherContainsInspection.scala`, it does similar thing.

src/org/jetbrains/plugins/scala/codeInspection/collections/EmulateFlattenInspection.scala
```diff
@@ -0,0 +1,39 @@
...
+  override def getSimplification(expr: ScExpression): Option[Simplification] =
+    expr match {
+      case seqOfSeqs `.flatMap` (identityOperation()) =>
+        Some(replace(expr).withText(invocationText(seqOfSeqs, "flatten")))
+      case _ => None
+    }
+
+  private object identityOperation {
+    def unapply(expr: ScExpression): Boolean = stripped(expr) match {
```
niktrop, line 27: I would use syntax for infix patterns here: `_'.identity' (underscore())` , but it's up to you.

lpiepiora, line 27: Good point, I will change it, because it actually reads more natural I think, since the first param is really a qualifier.

lpiepiora:
I've changed the code as you've suggested to check for `bindingExpr`. 

I think I understand now the intention. It wasn't clear to me (maybe still is a bit unclear), why `ScUnderscoreSection` once wraps the method value, and once it just wraps underscore. However I guess, that when it wraps the method value, then the `bindingExpr` is set to that method. Given that assumption I have changed the check in `underscore()`, so it does not pattern match against the example from my previous comment, by checking if `bindingExpr` is empty.

niktrop:
Thanks for contributing! I merged your changes manually.

PR #333, opened 2016-04-07
move left right for scala

niktrop:


src/org/jetbrains/plugins/scala/codeInsight/ScalaMoveLeftRightHandler.scala
```diff
@@ -0,0 +1,22 @@
+package org.jetbrains.plugins.scala.codeInsight
+
+import com.intellij.codeInsight.editorActions.moveLeftRight.MoveElementLeftRightHandler
+import com.intellij.psi.PsiElement
+import org.jetbrains.plugins.scala.lang.psi.api.expr.ScArgumentExprList
+import org.jetbrains.plugins.scala.lang.psi.api.statements.params.ScParameterClause
+
+/**
+  * @author Nikolay.Tropin
+  */
```
niktrop, line 10: Tests, huh?

mutcianm, line 10: Srsly, we always write tests for new features, how could you?

niktrop:
NOTE: 
This PR was used in my talk about contributing to our plugin, but it was merged too early. It will be completed with a separate commit, which will be referenced here.

niktrop:
This feature is finally finished: https://github.com/JetBrains/intellij-scala/commit/f5d7d219c24865a419493e0e43d39580e3e1720b

PR #337, opened 2016-04-29
Collection Inspection #SCL-9841 convert map.find(_ == k).map(_._2) to map(k)

mattfowler:
Adds the functionality described in #SCL-9841 to convert map.find(_ == k).map(_._2) to map(k). Inspection is named "find and map to apply" but there may be a better name for this inspection.


niktrop:
@mattfowler Thank you for contribution!
You need to update name of the description file, it should match inspection short name in the `scala-plugin-common.xml`. And it would be better to have all your changes in a single commit.

mattfowler:
@niktrop No problem! I updated the name of the description file and squashed the commits into commit b9c22d2. Not sure if any other branch manipulation is needed.  Let me know if there is.

PR #338, opened 2016-05-05
Dotty parsers

adkozlov:
Small fixes.

### DIFF
--- a/resources/META-INF/scala-plugin-common.xml
+++ b/resources/META-INF/scala-plugin-common.xml
@@ -1156,6 +1156,10 @@
                          displayName="Multiple arg lists in annotation constructor" groupPath="Scala" groupName="General"
                          shortName="MultipleArgListsInAnnotation" level="WARNING"
                          enabledByDefault="true" language="Scala"/>
+        <localInspection implementationClass="org.jetbrains.plugins.dotty.codeInspection.deprecated.WithTypeInspection"
+                         displayName="Deprecated with type" groupPath="Scala" groupName="Other"
+                         shortName="DottyDeprecatedWith" level="WARNING"
+                         enabledByDefault="true" language="Scala"/>
 
 
         <!-- scaladoc inspections -->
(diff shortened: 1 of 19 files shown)


src/org/jetbrains/plugins/dotty/lang/psi/types/ScTypePsiTypeBridge.scala
```diff
@@ -33,4 +38,34 @@ object ScTypePsiTypeBridge extends api.ScTypePsiTypeBridge {
       case _ => super.toScType(`type`, project, scope, visitedRawTypes, paramTopLevel, treatJavaObjectAsAny)
     }
   }
+
+  override def toPsiType(`type`: ScType, project: Project, scope: GlobalSearchScope, noPrimitives: Boolean, skolemToWildcard: Boolean) = {
+    def createComponent: ScType => PsiType =
+      toPsiType(_, project, scope, noPrimitives, skolemToWildcard)
+
+    val javaObject = createJavaObject(project, scope)
```
niktrop: javaObject may be created only when it is actually used

resources/META-INF/scala-plugin-common.xml
```diff
@@ -1156,6 +1156,10 @@
                          displayName="Multiple arg lists in annotation constructor" groupPath="Scala" groupName="General"
                          shortName="MultipleArgListsInAnnotation" level="WARNING"
                          enabledByDefault="true" language="Scala"/>
+      <localInspection implementationClass="org.jetbrains.plugins.dotty.codeInspection.deprecated.WithTypeInspection"
+                       displayName="Deprecated with type" groupPath="Scala" groupName="Other"
+                       shortName="DottyDeprecatedWith" level="WARNING"
```
niktrop: You need to add `DottyDeprecatedWith.html` file to `resources/inspectionDescriptions`

niktrop:
Merged manually

PR #339, opened 2016-05-09
Inspection #SCL-9844 - Remove Redundant `new` on Case Class Instantiation

mattfowler:
Adds an inspection to remove a redundant new on case class instantiation. 

There's a few corner cases in this (see unit tests), I believe I've got them all.  If there's any more you can think of I'm happy to take a look given a failing test case.

### DIFF
--- a/src/org/jetbrains/plugins/scala/codeInspection/caseClassRedundantNew/RedundantNewCaseClassInspection.scala
+++ b/src/org/jetbrains/plugins/scala/codeInspection/caseClassRedundantNew/RedundantNewCaseClassInspection.scala
@@ -1,0 +1,92 @@
+package org.jetbrains.plugins.scala.codeInspection.caseClassRedundantNew
+
+import com.intellij.codeInspection.{ProblemHighlightType, ProblemsHolder}
+import com.intellij.psi.PsiElement
+import org.jetbrains.plugins.scala.ScalaBundle
+import org.jetbrains.plugins.scala.codeInspection.AbstractInspection
+import org.jetbrains.plugins.scala.lang.psi.api.base.{ScConstructor, ScPrimaryConstructor}
+import org.jetbrains.plugins.scala.lang.psi.api.expr.ScNewTemplateDefinition
+import org.jetbrains.plugins.scala.lang.psi.api.toplevel.templates.ScClassParents
+import org.jetbrains.plugins.scala.lang.psi.api.toplevel.typedef.ScClass
+import org.jetbrains.plugins.scala.lang.psi.types.api.designator.DesignatorOwner
+import org.jetbrains.plugins.scala.lang.psi.types.{ScParameterizedType, ScType}
+
+/**
+  * mattfowler
+  * 5/7/2016
+  */
+class RedundantNewCaseClassInspection extends AbstractInspection("RedundantNewCaseClass", "Redundant New on Case Class") {
+
+  override def actionFor(holder: ProblemsHolder): PartialFunction[PsiElement, Any] = {
+    case newTemplate: ScNewTemplateDefinition =>
+      if (hasRedundantNew(newTemplate)) {
+        holder.registerProblem(newTemplate.getFirstChild, ScalaBundle.message("new.on.case.class.instantiation.redundant"),
+          ProblemHighlightType.GENERIC_ERROR_OR_WARNING, new RemoveNewQuickFix(newTemplate))
+      }
+  }
+
+  private def hasRedundantNew(newTemplate: ScNewTemplateDefinition): Boolean = {
+    val constructor = getConstructorFromTemplate(newTemplate)
+    val maybeScType: Option[ScType] = getConstructorType(constructor)
+
+    isTypeCaseClass(maybeScType) && isCreatingSameType(newTemplate.getType().toOption, maybeScType) &&
+      constructorCallHasArgumentList(constructor) && isProblemlessPrimaryConstructor(constructor)
+  }
+
+  /**
+    * Determines if the type of the extends block is the same as the type of the new template type.
+    * This prevents us from incorrectly displaying a warning when creating anonymous classes or instances with
+    * mixin traits.
+    */
+  private def isCreatingSameType(newTemplateType: Option[ScType], extendsBlockType: Option[ScType]): Boolean = {
+    (for {
+      templateType <- newTemplateType
+      extendsType <- extendsBlockType
+    } yield {
+      templateType.equals(extendsType)
+    }).getOrElse(false)
+  }
+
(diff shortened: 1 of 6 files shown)


retronym:
Overloaded `apply` methods can cause a problem:

```
scala> object O { case class C(a: Int, x: String = "default"); object C { def apply(a: Int) = new C(a, "xxx") } }
defined object O

scala> new O.C(0)
res5: O.C = C(0,default)

scala> O.C(0)
res6: O.C = C(0,xxx)
```

retronym:
Might need to check that the constructor call is not via a type alias:

```
scala> object O { case class C(a: Any); type CC = C }
defined object O

scala> new O.CC("")
res10: O.C = C()

scala> O.CC("")
<console>:15: error: value CC is not a member of object O
       O.CC("")
         ^
```

src/org/jetbrains/plugins/scala/codeInspection/caseClassRedundantNew/RedundantNewCaseClassInspection.scala
```diff
@@ -0,0 +1,92 @@
...
+
+    isTypeCaseClass(maybeScType) && isCreatingSameType(newTemplate.getType().toOption, maybeScType) &&
+      constructorCallHasArgumentList(constructor) && isProblemlessPrimaryConstructor(constructor)
+  }
+
+  /**
+    * Determines if the type of the extends block is the same as the type of the new template type.
+    * This prevents us from incorrectly displaying a warning when creating anonymous classes or instances with
+    * mixin traits.
+    */
```
niktrop, line 39: It would be better to check this by PSI tree only, without computing types. Something like this: `ScTemplateParents.typeElementsWithoutConstructor.isEmpty`

src/org/jetbrains/plugins/scala/codeInspection/caseClassRedundantNew/RedundantNewCaseClassInspection.scala
```diff
@@ -0,0 +1,92 @@
...
+      ref <- constructor.reference
+    } yield {
+      ref.advancedResolve.filter(_.problems.isEmpty)
+        .map(_.element).exists {
+        case primary: ScPrimaryConstructor => true
+        case _ => false
+      }
+    }).getOrElse(true)
+  }
+
```
niktrop, line 71: Same thing, working with types is not necessary. You already do resolve in `isProblemlessPrimaryConstructor` method. So you may check that it's `containingClass` is case class. `ScPrimaryConstructor.ofClass` extractor can save you couple of lines here.

src/org/jetbrains/plugins/scala/codeInspection/caseClassRedundantNew/RemoveNewQuickFix.scala
```diff
@@ -0,0 +1,21 @@
...
+import org.jetbrains.plugins.scala.lang.psi.impl.ScalaPsiElementFactory
+
+/**
+  * mattfowler
+  * 5/7/16.
+  */
+class RemoveNewQuickFix(param: ScNewTemplateDefinition) extends AbstractFixOnPsiElement(ScalaBundle.message("new.on.case.class.instantiation.redundant"), param) {
+  override def doApplyFix(project: Project): Unit = {
+    val p = getElement
+    if (!p.isValid) return
```
niktrop, line 17: You should use `p` instead of `param` all over the place. References to constructor arguments should be avoided in inheritors of `AbstractFixOnPsiElement`.

niktrop:
Quick and dirty way to address problem with overloaded `apply` methods is to create expression without new and try to resolve reference in it. It should resolve to a companion object's `apply` method and `isSyntheticApply` should be true. For correct resolve of a created reference you need to use `ScalaPsiElementFactory.createExpressionWithContextFromText`.

mattfowler:
@niktrop The last few commits add tests for the cases @retronym pointed out (thanks!) and cleaned up the code per your suggestions (not using types and simplifying the case class check).  It looks like we still need to use types to check if the constructor call is being made through a type alias.

It looks like the test case for the overloaded apply methods is passing.  If there's another case I'm missing let me know.

src/org/jetbrains/plugins/scala/codeInspection/caseClassRedundantNew/RemoveNewQuickFix.scala
```diff
@@ -15,7 +15,7 @@ class RemoveNewQuickFix(param: ScNewTemplateDefinition) extends AbstractFixOnPsi
   override def doApplyFix(project: Project): Unit = {
     val p = getElement
     if (!p.isValid) return
-    param.findFirstChildByType(ScalaTokenTypes.kNEW).delete()
-    param.replaceExpression(ScalaPsiElementFactory.createExpressionFromText(param.getText, p.getManager), false)
+    p.findFirstChildByType(ScalaTokenTypes.kNEW).delete()
```
niktrop, line 18: You missed one `param` here.

niktrop:
Result of `ref.advancedResolve` contains a lot of useful stuff. In particular `getActualElement` will return a type alias if the constructor was called that way.

Your set of tests looks good to me. I'd ask you to squash all commits for this PR too.

mattfowler:
Switched to using getActualElement and squashed the commits.

niktrop:
Great, thanks! 
Merged manually.

niktrop:
@mattfowler Hi again,
Our QA engineer found a couple of false positives and a minor UI problem for this inspection:
[SCL-10285](https://youtrack.jetbrains.com/issue/SCL-10285), [SCL-10287](https://youtrack.jetbrains.com/issue/SCL-10287) and [SCL-10289](https://youtrack.jetbrains.com/issue/SCL-10289). I can fix them myself but decided to ask you first.

mattfowler:
@niktrop I'm taking a look now, for SCL-10289 it looks like I can call `ScalaPsiElementFactory.createExpressionWithContextFromText(...)` to create a ScMethodCall with the extends block text, but the reference is always null.  My plan is to get the reference and check if `reference.multiResolve(false) != 1` as in the 'cannot resolve apply' annotation. Is this the correct approach?

niktrop:
@mattfowler You need to use `ScMethodCall.deepestInvokedExpression`, it must be an instance of `ScReferenceExpresssion` in this case. I think the way described in [my comment](https://github.com/JetBrains/intellij-scala/pull/339#issuecomment-218129523) is better. Reference may resolve to a single user-defined apply method, we don't want to remove `new` in this case.

Actually, I just noticed that `testOverriddenApplyMethod` is wrong, inspection shouldn't be available there for the same reason.

mattfowler:
@niktrop Makes sense, looks like I'm getting both bugs fixed with `isSyntheticApply`. I'll clean it up and create a PR for this later tonight. Thanks!

PR #340, opened 2016-05-14
New intention to regenerate type annotation for var/val/def

markus1189:
Adds a new intention to regenerate type annotations.  This essentially performs a "remove type annotation" followed by "add type annotation".  

The reason for adding this is that during refactoring, when I change the type of a method/var/val I always did the two steps above.  I thought it would be really helpful to do this in one step, so I added this as a separate intention.

I am not sure if I need a `youtrack` ticket for this?

Best,
Markus

### DIFF
--- a/resources/META-INF/scala-plugin-common.xml
+++ b/resources/META-INF/scala-plugin-common.xml
@@ -414,6 +414,11 @@
 
         <intentionAction>
             <category>Scala/Type</category>
+            <className>org.jetbrains.plugins.scala.codeInsight.intention.types.RegenerateTypeAnnotation</className>
+        </intentionAction>
+
+        <intentionAction>
+            <category>Scala/Type</category>
             <className>org.jetbrains.plugins.scala.codeInsight.intention.types.MakeTypeMoreSpecificIntention</className>
         </intentionAction>
 
--- a/resources/intentionDescriptions/RegenerateTypeAnnotation/after.scala.template
+++ b/resources/intentionDescriptions/RegenerateTypeAnnotation/after.scala.template
@@ -1,0 +1,3 @@
+var actuallyChar: <spot>Char</spot> = 'a'
+val actuallInt: <spot>Int</spot> = 42
+def actuallyString: <spot>String</spot> = "hello world"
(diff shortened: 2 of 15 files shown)


niktrop:
Hi, Markus,
It's a great idea for intention indeed. It would be better to have a ticket on youtrack so we could reference it in release notes.

niktrop:
I think it would be better to reuse `ToggleTypeAnnotation.complete` method and provide two new implementations of `Strategy` to it. You may extend `StrategyAdapter` and override only `removeFromFunction`, `removeFromValue` and `removeFromVariable` methods.

markus1189:
> It would be better to have a ticket on youtrack so we could reference it in release notes.

Added as SCL-10293 in youtrack

@niktrop: Thank you for your feedback. Can you explain a little more what you mean?  My (totally uninformed) reasoning was that:
- `RegenerateTypeAnnotation` can be documented separately
- I need to provide different messages for the intention than what add/remove does
- I honestly didn't try out whether I can generate more than one intention in the popup from the old implementation that shows `Add/Remove type annotation`

As I am not super deep into the codebase I would be glad about your opinion on this.

niktrop:
Creating a new intention is perfectly fine. But it can use (static) `ToggleTypeAnnotation.complete` method instead of implementing very similar `RegerenerateTypeAnnotation.complete`. Also you don't need to change `Strategy` trait at all. All you need to do in the new intention is to create appropriate `Strategy` for showing a message in `isAvailable` method and another one for doing actual work in `invoke` method. These two strategies should actually implement only three methods of `Strategy` trait. Second one may create `AddOrRemoveStrategy` inside and call corresponding methods of it.

markus1189:
Okay I see what you mean ;)

I applied your suggestions to remove the additional methods inside Strategy and instead overwrite the functions in a `RegenerateStrategy`.  I was unhappy with the names, because in my case `removeValueType` would then not remove it but regenerate it, so I _renamed the functions inside `Strategy`_ to `...WithType` and `...WithoutType` because the concrete Strategy is the one that decides what will be done (add/remove/regenerate).  

I am still a little unhappy about the `ToggleTypeAnnotation.complete` name, should we move it into a separate object and call it something else?  Maybe `ApplyTypeStrategy.apply`?

niktrop:
Yes, I think your suggested names is better then existing ones.

niktrop:
Thank you for contribution, merged manually: https://github.com/JetBrains/intellij-scala/commit/fe5dfd1ad5839700231d1d146fd60557c50bc31d

PR #341, opened 2016-05-17
Fixing SCL-9840 and Documentation

neowulf:
Fixes the following bug
[SCL-9840: Collection inspection: convert map.get(k).get to map(k)](https://youtrack.jetbrains.com/issue/SCL-9840)

and updating the documentation as well.


niktrop:
Thank you for contributing!

I tried to import project from scratch and run configurations weren't added without checkout or pull from repo. So I left that item in README. Also, I edited info about heap size for compiler to cover more cases.

Merged manually.

PR #342, opened 2016-05-20
Fixes for SCL-10285 (language fixes) and SCL-10285, SCL-10287

mattfowler:
Fixes the language issues and issues with overloaded apply methods.

@niktrop I was able to get this to work by checking `isSyntheticApply` but the case with `implicitly` in the ticket was still failing.  Checking that `syntheticNavigationElement` mapped to `ScClass` seemed to get all the test cases passing.


niktrop:
Thanks! Merged manually: e9172a3a5c52f48a963003512851e7b04365b62f

PR #343, opened 2016-05-21
SCL-9843 Intention: Create Companion Object

mattfowler:
Adds the ability to create a companion object for a class.

The following test (not in this PR) works in the editor but the test itself fails with an `java.lang.IllegalStateException: Attempt to modify PSI for non-committed Document!` exception. It's not clear to me if this is an issue with the intention code as it does not throw an exception in the editor.

```
def testShouldCreateCompanionInNestedObject(): Unit = {
    val text =
      s"""
         |class Foo {}
         |
         |object Foo {}
         |
         |object Bar {
         |  class Fo<caret>o {}
         |}
       """.stripMargin

    val expected =
      s"""
         |class Foo {}
         |
         |object Foo {}
         |
         |object Bar {
         |  class Fo<caret>o {}
         |
         |  object Foo {
         |  }
         |}
       """.stripMargin

    doTest(text, expected)
  } 
```

### DIFF
--- a/src/org/jetbrains/plugins/scala/codeInsight/intention/CreateCompanionObjectIntention.scala
+++ b/src/org/jetbrains/plugins/scala/codeInsight/intention/CreateCompanionObjectIntention.scala
@@ -1,0 +1,49 @@
+package org.jetbrains.plugins.scala.codeInsight.intention
+
+import com.intellij.codeInsight.intention.PsiElementBaseIntentionAction
+import com.intellij.openapi.editor.Editor
+import com.intellij.openapi.project.Project
+import com.intellij.psi.PsiElement
+import org.jetbrains.plugins.scala.extensions.Parent
+import org.jetbrains.plugins.scala.lang.psi.api.toplevel.typedef.{ScClass, ScObject}
+import org.jetbrains.plugins.scala.lang.psi.impl.ScalaPsiElementFactory
+
+/**
+  * mattfowler
+  * 5/21/2016
+  */
+class CreateCompanionObjectIntention extends PsiElementBaseIntentionAction {
+  override def getText = "Create companion object for class"
+
+  override def invoke(project: Project, editor: Editor, psiElement: PsiElement): Unit = {
+    getClassIfAvailable(psiElement).foreach { clazz =>
+      val companion = ScalaPsiElementFactory.createObjectWithContext(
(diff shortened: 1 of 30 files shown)


niktrop:
We have `ScalaGenerateCompanionObjectHandler` which does the same task. We'd like to add an inspection because it is a more familiar way to do things in IDEA for many users. But code that inserts companion object may be reused here.

Also, it is better to create new git branch for each pull request. In this case only relevant commits will be shown.

src/org/jetbrains/plugins/scala/codeInsight/intention/CreateCompanionObjectIntention.scala
```diff
@@ -0,0 +1,49 @@
...
+/**
+  * mattfowler
+  * 5/21/2016
+  */
+class CreateCompanionObjectIntention extends PsiElementBaseIntentionAction {
+  override def getText = "Create companion object for class"
+
+  override def invoke(project: Project, editor: Editor, psiElement: PsiElement): Unit = {
+    getClassIfAvailable(psiElement).foreach { clazz =>
+      val companion = ScalaPsiElementFactory.createObjectWithContext(
```
niktrop, line 20: Multiline strings have different line separators on different platforms. IDEA expects "\n", but on Windows it gets "\r\n".

src/org/jetbrains/plugins/scala/codeInsight/intention/CreateCompanionObjectIntention.scala
```diff
@@ -0,0 +1,49 @@
...
+
+  override def isAvailable(project: Project, editor: Editor, psiElement: PsiElement): Boolean = {
+    getClassIfAvailable(psiElement).exists(clazz => clazz.getParent.getChildren.forall {
+      case obj: ScObject => obj.name != clazz.name
+      case _ => true
+    })
+  }
+
+  private def getClassIfAvailable(psiElement: PsiElement): Option[ScClass] = {
+    psiElement match {
```
niktrop, line 37: This intention makes sense for traits too.

niktrop:
I've merged your PR with some minor changes:
https://github.com/JetBrains/intellij-scala/commit/7217e9ee94fb34a00026d14630c2c133718070d7

PR #344, opened 2016-05-22
SCL-10281: Consistent formatting in the inspection description 

neowulf:
Fixing https://youtrack.jetbrains.com/issue/SCL-10281


niktrop:
Thank you for such thorough work. I checked descriptions in Inspections Settings view and most of them looks good. But I found a few inspections where I think readability get worse:
**Scala style inspection**, **Advanced language features** and **Type annotation required**. 

Also I found several inspection without description. You may fix them too, if you want:
**Match statement convertible to pattern matching anonymous function
Multiple arg lists in annotation constructor
Reference must be prefixed
Relative import
Unmatched tag
Unnecessary parentheses
Unnecessary partial function
Header tags unbalanced
Unknown parameter
Unknown tag**

neowulf:
Thank you for the feedback.
- Does the readability get worse in the code or in the IDE?
- I'll try to add additional descriptions for the missing ones.

niktrop:
It's about rendered version, for example in the inspections settings.
Before:
description_before
After:
description_after

"Before" version is not perfect, but I think that description of the inspection and instructions should be separated.

niktrop:
If you'll have any problems with an inspection without description don't hesitate to ask or just skip it.

neowulf:
Closing this ticket so that the PR can be done against idea162.x

PR #345, opened 2016-05-27
SCL-10351 Inspection: convert null initializer to _

Im-dex:
Implementation of [SCL-10351](https://youtrack.jetbrains.com/issue/SCL-10351)

### DIFF
--- a/resources/META-INF/scala-plugin-common.xml
+++ b/resources/META-INF/scala-plugin-common.xml
@@ -879,6 +879,10 @@
         <localInspection implementationClass="org.jetbrains.plugins.scala.codeInspection.functionExpressions.UnnecessaryPartialFunctionInspection"
                          displayName="Unnecessary Partial Function" groupPath="Scala" groupName="General" shortName="UnnecessaryPartialFunction"
                          id="UnnecessaryPartialFunction" level="WARNING" enabledByDefault="true" language="Scala"/>
+        <localInspection implementationClass="org.jetbrains.plugins.scala.codeInspection.convertNullInitializerToUnderscore.ConvertNullInitializerToUnderscore"
+                         displayName="Null initializer can be replaced by _" groupPath="Scala" groupName="General"
+                         shortName="ConvertNullInitializerToUnderscore" level="WEAK WARNING"
+                         enabledByDefault="true" language="Scala"/>
         <!--<localInspection implementationClass="org.jetbrains.plugins.scala.codeInspection.allErrorsInspection.AnnotatorBasedErrorInspection"
                         displayName="All Errors Tool" groupPath="Scala" groupName="General"
                         shortName="ScalaAllErrors" level="WARNING"
@@ -1182,7 +1186,6 @@
                          displayName="Deprecated with type" groupPath="Scala" groupName="Other"
                          shortName="DottyDeprecatedWith" level="WARNING"
                          enabledByDefault="true" language="Scala"/>
-
 
         <!-- scaladoc inspections -->
         <localInspection implementationClass="org.jetbrains.plugins.scala.codeInspection.scaladoc.ScalaDocUnclosedTagWithoutParserInspection"
--- a/resources/META-INF/scala-plugin-common.xml
+++ b/resources/META-INF/scala-plugin-common.xml
@@ -879,10 +879,6 @@
         <localInspection implementationClass="org.jetbrains.plugins.scala.codeInspection.functionExpressions.UnnecessaryPartialFunctionInspection"
                          displayName="Unnecessary Partial Function" groupPath="Scala" groupName="General" shortName="UnnecessaryPartialFunction"
                          id="UnnecessaryPartialFunction" level="WARNING" enabledByDefault="true" language="Scala"/>
-        <localInspection implementationClass="org.jetbrains.plugins.scala.codeInspection.convertNullInitializerToUnderscore.ConvertNullInitializerToUnderscore"
-                         displayName="Null initializer can be replaced by _" groupPath="Scala" groupName="General"
-                         shortName="ConvertNullInitializerToUnderscore" level="WEAK WARNING"
-                         enabledByDefault="true" language="Scala"/>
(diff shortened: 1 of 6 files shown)


resources/META-INF/scala-plugin-common.xml
```diff
@@ -879,6 +879,10 @@
         <localInspection implementationClass="org.jetbrains.plugins.scala.codeInspection.functionExpressions.UnnecessaryPartialFunctionInspection"
                          displayName="Unnecessary Partial Function" groupPath="Scala" groupName="General" shortName="UnnecessaryPartialFunction"
                          id="UnnecessaryPartialFunction" level="WARNING" enabledByDefault="true" language="Scala"/>
+        <localInspection implementationClass="org.jetbrains.plugins.scala.codeInspection.convertNullInitializerToUnderscore.ConvertNullInitializerToUnderscore"
+                         displayName="Null initializer can be replaced by _" groupPath="Scala" groupName="General"
```
niktrop: It should go to "Syntactic Simplification" group, like [here](https://github.com/JetBrains/intellij-scala/commit/c96adc5be3cd23395b8a13418741e4bcacc0f12e).

src/org/jetbrains/plugins/scala/codeInspection/convertNullInitializerToUnderscore/ConvertNullInitializerToUnderscore.scala
```diff
@@ -0,0 +1,59 @@
...
+    expr.getFirstChild.getNode.getElementType == ScalaTokenTypes.kNULL
+  }
+
+  override def actionFor(holder: ProblemsHolder) = {
+    case variable: ScVariableDefinition if variable.expr.nonEmpty && variable.hasExplicitType =>
+      variable.declaredType.get match {
+        case valType: api.ValType if valType ne api.Unit =>
+        case declaredType  =>
+          val expr = variable.expr.get
+          if (expr.isValid) {
```
niktrop: I think just checking `isNull(expr)` is enough, we don't need to compute type here.

src/org/jetbrains/plugins/scala/codeInspection/convertNullInitializerToUnderscore/ConvertNullInitializerToUnderscore.scala
```diff
@@ -0,0 +1,59 @@
...
+import org.jetbrains.plugins.scala.lang.psi.api.statements.ScVariableDefinition
+import org.jetbrains.plugins.scala.lang.psi.impl.ScalaPsiElementFactory
+import org.jetbrains.plugins.scala.lang.psi.types.api
+
+/**
+  * Created by a.tsukanov on 26.05.2016.
+  */
+class ConvertNullInitializerToUnderscore extends AbstractInspection(inspectionId, inspectionName) {
+  private def registerProblem(holder: ProblemsHolder, variable: ScVariableDefinition): Unit = {
+    val fix = new ConvertNullInitializerToUnderscoreQuickFix(variable)
```
niktrop: You may highlighight just `null`, and use `ProblemHighlightType.WARNING`

src/org/jetbrains/plugins/scala/codeInspection/convertNullInitializerToUnderscore/ConvertNullInitializerToUnderscore.scala
```diff
@@ -0,0 +1,59 @@
...
+    holder.registerProblem(variable, inspectionName, ProblemHighlightType.WEAK_WARNING, fix)
+  }
+
+  private def isNull(expr: ScExpression): Boolean = {
+    expr.getFirstChild.getNode.getElementType == ScalaTokenTypes.kNULL
+  }
+
+  override def actionFor(holder: ProblemsHolder) = {
```
niktrop: You also need to check that a variable is not local, because such syntax is valid only for fields.

src/org/jetbrains/plugins/scala/codeInspection/convertNullInitializerToUnderscore/ConvertNullInitializerToUnderscore.scala
Im-dex: Constant `ProblemHighlightType.WARNING` does not exists, so, I removed `highlightType` parameter as in AutoTuplingInspection

resources/META-INF/scala-plugin-common.xml
```diff
@@ -920,6 +920,10 @@
                          displayName="Concealed &quot;apply&quot; call" groupPath="Scala" groupName="Syntactic Clarification"
                          shortName="ConcealedApplyCall" level="WARNING"
                          enabledByDefault="false" language="Scala"/>
+        <localInspection implementationClass="org.jetbrains.plugins.scala.codeInspection.syntacticClarification.ConvertNullInitializerToUnderscore"
+                         displayName="Null initializer can be replaced by _" groupPath="Scala" groupName="Syntactic Clarification"
+                         shortName="ConvertNullInitializerToUnderscore" level="WARNING"
+                         enabledByDefault="true" language="Scala"/>
```
ilinum, line 926: I'm not sure it's a good idea to enable it by default. For people who are coming from C-like languages, `null` is more obvious than `_`. It is clearer that if you call methods on it, it will throw a `NullPointerException`. In addition, a lot of code will be highlighted yellow after that.

@niktrop @pavelfatin what do you think?

Im-dex, line 926: The main issue with the new inspections disabled by default that most users will miss them due to lack of normal changlelog. But the enabled inspection can always be disabled manually.

niktrop:
@ilinum Using default initializer is [more idiomatic](http://stackoverflow.com/questions/2440134/is-this-the-proper-way-to-initialize-null-references-in-scala). And I think it won't add this much yellow code anyway.

niktrop:
@Im-dex Thanks for contribution! I merged it manually with a couple of small changes: https://github.com/JetBrains/intellij-scala/commit/e7eb5bb5ba55eb2242fce94fbfe3e17c16a1d2e6

I would suggest to commit to a separate branch and create pull requests from it next time. This way you won't need to merge changes from master.

PR #346, opened 2016-05-27
Support backtickeds on index level

katejim:


niktrop:
I would suggest to change some method names in ScalaNamesUtil:
`convertMemberName` -> `clean`
`convertMemberFqn` -> `cleanFqn`
`fqnNamesEquals` -> `equivalentFqn`
`memberNamesEquals` -> `equivalent`
`addBacktickedIfScalaKeywordFqn` -> `escapeKeywordsFqn`
`changeKeyword` -> `escapeKeyword`

But with such names these methods should always be used with `ScalaNamesUtil` qualifier.

src/org/jetbrains/plugins/scala/lang/refactoring/util/ScalaNamesUtil.scala
```diff
@@ -96,8 +103,40 @@ object ScalaNamesUtil {
...
+  def fqnNamesEquals(l: String, r: String): Boolean = {
+    if (l == r) return true
+    convertMemberFqn(l) == convertMemberFqn(r)
   }
+
+  def memberNamesEquals(l: String, r: String): Boolean = {
+    if (l == r) return true
+    convertMemberName(l) == convertMemberName(r)
+  }
+
```
niktrop: It seems that these two methods are not needed anymore.

src/org/jetbrains/plugins/scala/lang/refactoring/util/ScalaNamesUtil.scala
```diff
@@ -96,8 +103,40 @@ object ScalaNamesUtil {
...
+    val toDecode = name match {
+      case ScalaNamesUtil.isBacktickedName(s) => s
+      case _ => name
+    }
+    NameTransformer.decode(toDecode)
+  }
+
+  def convertMemberFqn(fqn: String): String =
+    splitName(fqn).map(convertMemberName).mkString(".")
+
```
niktrop, line 116: `l == r || convertMemberFqn(l) == convertMemberFqn(r)`

katejim:
merged manually

PR #347, opened 2016-06-02
Problem with two full Scala.js cross projects

Starzu:
Lets assume that we've got a config like:

``` scala
val `A-shared` = crossProject.crossType(CrossType.Full).in(file("A/shared"))
val `B-shared` = crossProject.crossType(CrossType.Full).in(file("B/shared"))
```

The folders structure looks like:

```
/
|- A
|  |- shared
|  |  |- js
|  |  |  |- src
|  |  |- jvm
|  |  |  |- src
|  |  |- shared
|  |  |  |- src
|  |- ...
|- B
|  |- shared
|  |  |- js
|  |  |  |- src
|  |  |- jvm
|  |  |  |- src
|  |  |- shared
|  |  |  |- src
|  |- ...
```

Finally I have got two modules named `shared-sources` by IntelliJ.

### DIFF
--- a/src/org/jetbrains/sbt/project/ExternalSourceRootResolution.scala
+++ b/src/org/jetbrains/sbt/project/ExternalSourceRootResolution.scala
@@ -185,8 +185,12 @@ trait ExternalSourceRootResolution { self: SbtProjectResolver =>
     var counter = 1
 
     def nameFor(base: Option[File]) = {
-      val namedDirectory = if (base.exists(_.getName == "shared")) base.flatMap(_.parent) else base
-      val prefix = namedDirectory.map(_.getName + "-sources").getOrElse("shared-sources")
+      val namedDirectory = {
+        var d = base
+        while (d.exists(_.getName == "shared")) d = d.flatMap(_.parent)
+        d
+      }
+      val prefix = namedDirectory.map(_.getName + "-shared-sources").getOrElse("shared-sources")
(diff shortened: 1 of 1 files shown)


pavelfatin:
Thanks for the pull request! We'll need some time to make sure that such a change is compatible with other variants of project structure.

src/org/jetbrains/sbt/project/ExternalSourceRootResolution.scala
```diff
@@ -185,8 +185,12 @@ trait ExternalSourceRootResolution { self: SbtProjectResolver =>
     var counter = 1
 
     def nameFor(base: Option[File]) = {
-      val namedDirectory = if (base.exists(_.getName == "shared")) base.flatMap(_.parent) else base
-      val prefix = namedDirectory.map(_.getName + "-sources").getOrElse("shared-sources")
+      val namedDirectory = {
+        var d = base
+        while (d.exists(_.getName == "shared")) d = d.flatMap(_.parent)
```
pavelfatin, line 190: Ideally, it's better to avoid the "imperative" syntax and the mutable state (though local), for example, by defining `parents: Stream[File]` helper method in `RichFile`.

jastice:
PR was closed during archiving of old branches to tags. If it's still relevant, can you please reopen the PR to the current development branch (idea163.x)? Thanks!
(and sorry about the delay handling it)

PR #348, opened 2016-06-04
Auto convert to infix when converting from .equals() to ==

neowulf:
Providing an intermediate commit for feedback only.
1. There are two implementations - both work but I am not sure which one is desirable. Added a comment in the code.
2. One of the test cases does not insert a space. I have added a comment in the code as well.

### DIFF
--- a/src/org/jetbrains/plugins/scala/codeInsight/intention/booleans/ReplaceEqualsOrEqualityInMethodCallExprIntention.scala
+++ b/src/org/jetbrains/plugins/scala/codeInsight/intention/booleans/ReplaceEqualsOrEqualityInMethodCallExprIntention.scala
@@ -8,13 +8,15 @@ import com.intellij.openapi.util.TextRange
 import com.intellij.psi.util.PsiTreeUtil
 import com.intellij.psi.{PsiDocumentManager, PsiElement}
 import org.jetbrains.plugins.scala.extensions._
-import org.jetbrains.plugins.scala.lang.psi.api.expr.{ScMethodCall, ScReferenceExpression}
+import org.jetbrains.plugins.scala.lang.psi.api.expr.{ScInfixExpr, ScMethodCall, ScReferenceExpression}
 import org.jetbrains.plugins.scala.lang.psi.impl.ScalaPsiElementFactory
+import org.jetbrains.plugins.scala.lang.psi.impl.expr.ScInfixExprImpl
+import org.jetbrains.plugins.scala.util.IntentionUtils
 
 /**
- * @author Ksenia.Sautina
- * @since 4/23/12
- */
+  * @author Ksenia.Sautina
+  * @since 4/23/12
+  */
 
 object ReplaceEqualsOrEqualityInMethodCallExprIntention {
   def familyName = "Replace equals or equality in method call expression"
@@ -44,7 +46,12 @@ class ReplaceEqualsOrEqualityInMethodCallExprIntention extends PsiElementBaseInt
     false
   }
 
-  override def invoke(project: Project, editor: Editor, element: PsiElement) {
+  override def invoke(project: Project, editor: Editor, element: PsiElement): Unit = {
+    invoke_method_one(project, editor, element)
+    //invoke_method_two(project, editor, element)
+  }
+
+  def orig_invoke(project: Project, editor: Editor, element: PsiElement) {
     val methodCallExpr: ScMethodCall = PsiTreeUtil.getParentOfType(element, classOf[ScMethodCall], false)
     if (methodCallExpr == null || !methodCallExpr.isValid) return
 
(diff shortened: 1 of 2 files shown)


src/org/jetbrains/plugins/scala/codeInsight/intention/booleans/ReplaceEqualsOrEqualityInMethodCallExprIntention.scala
```diff
@@ -44,7 +46,12 @@ class ReplaceEqualsOrEqualityInMethodCallExprIntention extends PsiElementBaseInt
     false
   }
 
-  override def invoke(project: Project, editor: Editor, element: PsiElement) {
+  override def invoke(project: Project, editor: Editor, element: PsiElement): Unit = {
+    invoke_method_one(project, editor, element)
```
neowulf, line 50: Is this the desired method? If so, I will inline this method and remove the other methods in the final PR.

src/org/jetbrains/plugins/scala/codeInsight/intention/booleans/ReplaceEqualsOrEqualityInMethodCallExprIntention.scala
```diff
@@ -55,12 +62,100 @@ class ReplaceEqualsOrEqualityInMethodCallExprIntention extends PsiElementBaseInt
...
+          newMethodCallExpr.asInstanceOf[ScMethodCall].getInvokedExpr.asInstanceOf[ScReferenceExpression].nameId.
             getTextRange.getStartOffset - newMethodCallExpr.getTextRange.getStartOffset
+        case scInfixExpr: ScInfixExpr =>
+          newMethodCallExpr.asInstanceOf[ScInfixExpr].operation.nameId.getTextRange.getStartOffset - newMethodCallExpr.getTextRange.getStartOffset
+      }
+    }
+
+    inWriteAction {
+      methodCallExpr.replace(newMethodCallExpr)
+      methodCallExpr.replaceExpression(newMethodCallExpr, removeParenthesis = true)
```
neowulf, line 119: What's the difference between line 118 and line 119?

test/org/jetbrains/plugins/scala/codeInsight/intentions/booleans/ReplaceEqualsOrEqualityInMethodCallExprIntentionTest.scala
```diff
@@ -21,8 +21,34 @@ class ReplaceEqualsOrEqualityInMethodCallExprIntentionTest extends ScalaIntentio
...
+    val resultText = "(true, false) <caret>==(false, false)"
+
+    doTest(text, resultText)
+  }
+
+  def testReplaceTuple_2(): Unit = {
+    (true, false).equals(false, false)
+    val text = "if((true, false).equ<caret>als  (false, false)) return "
+    //TODO desired (true, false) <caret>== (false, false)
+    val resultText = "if ((true, false) <caret>==(false, false)) return"
```
neowulf, line 49: I can't seem to get the space inserted after `==` and before `(`.

When I getText of the psi element resulting from the expression listed in line 118 and [line 119](https://github.com/neowulf33/intellij-scala/blob/425912dbc9099ede89291ae62ad70b6953c62398/src/org/jetbrains/plugins/scala/codeInsight/intention/booleans/ReplaceEqualsOrEqualityInMethodCallExprIntention.scala#L119), the desired space exists.

Does the space get removed when the following line gets executed?

```
PsiDocumentManager.getInstance(project).commitDocument(editor.getDocument)
```

src/org/jetbrains/plugins/scala/codeInsight/intention/booleans/ReplaceEqualsOrEqualityInMethodCallExprIntention.scala
niktrop, line 50: I would rewrite this inspection with string interpolation instead of `StringBuilders`, I think it would be much cleaner.

src/org/jetbrains/plugins/scala/codeInsight/intention/booleans/ReplaceEqualsOrEqualityInMethodCallExprIntention.scala
```diff
@@ -55,12 +62,100 @@ class ReplaceEqualsOrEqualityInMethodCallExprIntention extends PsiElementBaseInt
...
+    val args = new StringBuilder().append(methodCallExpr.args.getText)
+
+    val oper = methodCallExpr.getInvokedExpr.asInstanceOf[ScReferenceExpression].nameId.getText
+    val desiredOper: String = replaceOper(oper)
+
+    val expr = new StringBuilder()
+      .append(methodCallExpr.getInvokedExpr.asInstanceOf[ScReferenceExpression].qualifier.get.getText)
+
+    if (desiredOper == "==") {
+      expr.append(" ")
```
niktrop, line 96: I think we should remove or refactor usages of `IntentionUtils.analyzeMethodCallArgs`. It's too java-style and has a misleading name. So in that regard I found `invoke_method_two` better.

src/org/jetbrains/plugins/scala/codeInsight/intention/booleans/ReplaceEqualsOrEqualityInMethodCallExprIntention.scala
niktrop, line 119: `replaceExpression` is more specialized. It works only for `ScExpression` and may remove/add enclosing parentheses. By the way, you can add a test for `x.equals(y).toString` example from the discussion of SCL-9579 where parentheses are needed.

test/org/jetbrains/plugins/scala/codeInsight/intentions/booleans/ReplaceEqualsOrEqualityInMethodCallExprIntentionTest.scala
niktrop, line 49: IDEA reformats changed lines of code after write action. There is a separate setting for that space:
https://youtrack.jetbrains.com/issue/SCL-3849
But probably for operators space shouldn't be removed, I'll fix it.

`commitDocument` method should be called after direct modification of document text to update psi. I think it's not needed here.

neowulf:
Thanks @niktrop for the feedback.
1. How would you refactor `IntentionUtils.analyzeMethodCallArgs`? How is it java-style?
   1. Indeed, it has a misleading name but that can be changed.
   2. This method seems to be comprehensive in that it handles different types of expressions. How else could it be improved?

I can see another type of improvement as follows:

Before:

```
  def analyzeMethodCallArgs(methodCallArgs: ScArgumentExprList, argsBuilder: scala.StringBuilder) {
    if (methodCallArgs.exprs.length == 1) {
      methodCallArgs.exprs.head match {
        case _: ScLiteral => argsBuilder.replace(argsBuilder.length - 1, argsBuilder.length, "").replace(0, 1, "")
        case _: ScTuple => argsBuilder.replace(argsBuilder.length - 1, argsBuilder.length, "").replace(0, 1, "")
        case _: ScReferenceExpression => argsBuilder.replace(argsBuilder.length - 1, argsBuilder.length, "").replace(0, 1, "")
        case _: ScGenericCall => argsBuilder.replace(argsBuilder.length - 1, argsBuilder.length, "").replace(0, 1, "")
        case _: ScXmlExpr => argsBuilder.replace(argsBuilder.length - 1, argsBuilder.length, "").replace(0, 1, "")
        case _: ScMethodCall => argsBuilder.replace(argsBuilder.length - 1, argsBuilder.length, "").replace(0, 1, "")
        case infix: ScInfixExpr if infix.getBaseExpr.isInstanceOf[ScUnderscoreSection] =>
          argsBuilder.insert(0, "(").append(")")
        case _ =>
      }
    }
  }
```

After:

```
def analyzeMethodCallArgs(methodCallArgs: ScArgumentExprList, argsBuilder: scala.StringBuilder) {
    if (methodCallArgs.exprs.length == 1) {
      methodCallArgs.exprs.head match {
        case _: ScLiteral | _: ScTuple | _: ScReferenceExpression | _: ScGenericCall | _: ScXmlExpr | _: ScMethodCall =>
          argsBuilder.replace(argsBuilder.length - 1, argsBuilder.length, "").replace(0, 1, "")
        case infix: ScInfixExpr if infix.getBaseExpr.isInstanceOf[ScUnderscoreSection] =>
          argsBuilder.insert(0, "(").append(")")
        case _ =>
      }
    }
  }
```

niktrop:
I think we may get rid of `StringBuilder` here, there is no need in a mutable state here. This method is called only when quickfix is invoked, so we should not worry much about garbage created.

It seems that it is used with text of argument list only, so signature may be `(ScArgumentExprList) => String`.

But actually, we may instead try to use `ScExpression.replaceExpression` here to handle parentheses. It uses more general `ScalaPsiUtil.needParentheses` method. So replacement of method call with infix expression would work like following:
1. Create infix expression from method call, without trying to do anything with parentheses.
2. If you get `ScParenthesizedExpr` as the right operand, invoke `replaceExpression` on it with it's content as an argument.
3. Finally, do `replaceExpression` on the whole infix expression.

We already have `ScalaPsiElementFactory.createEquivMethodCall(ScInfixExpr)`. It would be nice to have a method which makes the opposite.

niktrop:
Using `ScExpression.replaceExpression` for the right operand will probably fix [SCL-10381](https://youtrack.jetbrains.com/issue/SCL-10381) too, because autotupling is handled specially in the `needParentheses` method.

neowulf:
Closing this PR in favor of the final PR: https://github.com/JetBrains/intellij-scala/pull/361

PR #359, opened 2016-06-23
Fix error in intention examples

rce:
There's a simple syntax error in before/after examples for `ConvertToCurlyBracesIntention` and `ConvertToParenthesesIntention`.


adkozlov:
@rce Thanks!

PR #362, opened 2016-08-20
SCL-10573  Convert anonymous function parameter of foreach call to function value

likern:
Hello, guys! I've been waiting for any feedback.
Am I correct that inspections only analyze "feel and look" of expression, but not real environment.
I mean in `Seq("1", "2", "3").foreach(s => foo(s))` inspection doesn't analyze (or depends on another analyzer) whether `foo` really  exists and is really function in this _context_. It only analyzes only that `foo` looks like function call. Am I correct?

### DIFF
--- a/resources/META-INF/scala-plugin-common.xml
+++ b/resources/META-INF/scala-plugin-common.xml
@@ -1092,6 +1092,10 @@
                          shortName="FilterOtherContains" level="WARNING"
                          enabledByDefault="true" language="Scala"/>
         <!--simplifications: other-->
+        <localInspection implementationClass="org.jetbrains.plugins.scala.codeInspection.collections.SimplifiableForeachInspection"
+                         displayName="Simplifiable foreach method parameter" groupPath="Scala, Collections" groupName="Simplifications: other"
+                         shortName="SimplifiableForeach" level="WARNING"
+                         enabledByDefault="true" language="Scala"/>
         <localInspection implementationClass="org.jetbrains.plugins.scala.codeInspection.collections.SimplifiableFoldOrReduceInspection"
                          displayName="Simplifiable fold or reduce method" groupPath="Scala, Collections" groupName="Simplifications: other"
                          shortName="SimplifiableFoldOrReduce" level="WARNING"
--- a/resources/inspectionDescriptions/SimplifiableForeach.html
+++ b/resources/inspectionDescriptions/SimplifiableForeach.html
@@ -1,0 +1,19 @@
+<html>
+<body>
+Replaces anonymous function to a method value as parameter of <code>foreach</code> function.
+
+<p>
+Before:
+<pre>
+Seq("1", "2").foreach(s => foo(s))
+</pre>
+
+<br>
+After:
+<pre>
+Seq("1", "2").foreach(foo)
+</pre>
+</p>
+
+</body>
+</html>
--- a/src/org/jetbrains/plugins/scala/codeInspection/collections/SimplifiableForeachInspection.scala
+++ b/src/org/jetbrains/plugins/scala/codeInspection/collections/SimplifiableForeachInspection.scala
@@ -1,0 +1,35 @@
+package org.jetbrains.plugins.scala
+package codeInspection.collections
+
+import org.jetbrains.plugins.scala.codeInspection.InspectionBundle
+import org.jetbrains.plugins.scala.extensions.ResolvesTo
+import org.jetbrains.plugins.scala.lang.psi.api.expr.{ScExpression, ScFunctionExpr, ScMethodCall}
+
+/**
+  * @author Victor Malov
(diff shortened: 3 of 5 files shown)


niktrop:
Hi @likern,
In inspections you may access all semantic information about expression, including types and how references are resolved. The extractor `ResolvesTo` used in your PR does exactly that.

It seems that [SCL-10573](https://youtrack.jetbrains.com/issue/SCL-10573) is formulated not clear enough and is actually a duplicate of [SCL-10362](https://youtrack.jetbrains.com/issue/SCL-10362). Also see related [SCL-10333](https://youtrack.jetbrains.com/issue/SCL-10333) and [SCL-10330](https://youtrack.jetbrains.com/issue/SCL-10330). SCL-10362 is a combination of last two ones, so you may try one of them first.

likern:
OK, I'll try to implement these issues.

niktrop:
To sum it all up. We already have `ConvertibleToMethodValueInspection` but it is kind of messy and does not handle anonymous functions with explicit parameters. The idea behind these tickets is to make three different inspection instead. We may have:
A) Verbose anonymous function: `(x, y) => f(x, y)`
B) Anonymous function with placeholders: `f(_, _)`
C) Eta expansion: `f _`
D) Method value: `f`

SCL-10330: A -> B (see `IntroduceImplicitParameterIntention`)
SCL-10333: C -> D
SCL-10362: A or B -> C (or -> D if conditions of SCL-10333 are met)

likern:
Where I can ask questions about implementation help? Here or on YouTrack?
Could you explain a little bit futher about what is MethodRepr? I can't figure it out. Looks like it somehow related to methods, but also returns `Some()` for References like for **x** in `List(0) filter(x => f(x))`. 
And also don't understand what does mean `ResolvesTo`, what is it does actually doing?

Currently I'm trying to implement **SCL-10362**, the simplest form `x => f(x)  to f _`. As I understand:
1.  Pattern match that it is FunctionExpression (is anonymus function) (for this I think I can adopt this patch)
2. Figure out that body `f(x)` is method (here probably I have to use MethodRepr)
   Any help is very appreciated.

niktrop:
Hi,
You may ask here, I think it's best to have it all in one place. 

`MethodRepr` is an extractor to various syntactic forms of function invocations in Scala. It handles usual method calls, infix/prefix/postfix notations, calls with type arguments and calls without no parentheses. The latter syntactically is the same as any other reference. So we cannot distinguish it without resolve.

Resolving of a reference is finding a named element it points to. `ResolvesTo` is just a handy way to use `ScReference.resolve()` method in pattern matches.

The general idea is correct, but even in this seemingly simple case problems are possible, for example [SCL-9820](https://youtrack.jetbrains.com/issue/SCL-9820) and [SCL-9682](https://youtrack.jetbrains.com/issue/SCL-9682). Please check `ConvertibleToMethodValueInspection` implementation (which doesn't handle verbose anonymous function syntax).

likern:
Thanks a lot. Currently I'm working on [SCL-10362](https://youtrack.jetbrains.com/issue/SCL-10362), exactly on this part: x => f(x) to f _  and multiple parameters (later going to extend to f, where applicable).
I see the following use cases which I'm going to cover based on  [SCL-10362](https://youtrack.jetbrains.com/issue/SCL-10362):
1. val anon = **x => f(x)** to val anon = **f _**
2. val anon = **(x) => f(x)** to val anon = **f _**
3. val anon = **(x) => {f(x)}** to val anon = **f _** (detect inspection with block code)
4. val anon = **x => {f(x)}** to val anon = **f _** (detect inspection with block code)
5. val anon = **x => {   f(x)       }** to val anon = **f _** (detect inspection with block code and any arbitrary whitespaces inside block)
6. val anon = **x => {  
                                  f(x)
                       }** to val anon = **f _** (detect inspection with block code and any arbitrary whitespaces inside block, including newlines)
7. All as 1-6 but with additional semicolon at the end of function call like: val anon = **x => f(x);** to val anon = **f _** (detecting when removal this semicolon is safe to delete (doesn't split two expressions)
8. The same, but with multiple parameters
   If there is some non desired detection - please, give me a note.

I'm note sure about semicolons. Probably in this case val anon = **(x) => {f(x);}** it is part of anonymous function and should be converted. But in case  val anon = **(x) => {f(x)};** or val anon = **(x) => f(x);** it splits expressions (but right expression is absent) and should be another inspection?

niktrop:
Yes, all these cases should be highlighted. But most of them will be covered with `stripped` function which you have used already. Also, I think you don't need to treat semicolon in this inspection, we have a separate inspection for that.

jastice:
PR was closed during archiving of old branches to tags. If it's still relevant, can you please reopen the PR to the current development branch (idea163.x)? Thanks!

PR #364, opened 2016-09-10
Move LINE_COMMENT_AT_FIRST_COLUMN into ScalaCodeStyleSettings

chbrown:
This commit moves the `LINE_COMMENT_AT_FIRST_COLUMN` setting into the `ScalaCodeStyleSettings` class.
- It doesn't work as a CommonCodeStyleSettings value: https://youtrack.jetbrains.com/issue/SCL-9982
- I'm not the only person to run into this: https://youtrack.jetbrains.com/issue/SCL-9787

I'm not sure if https://youtrack.jetbrains.com/issue/SCL-7898 ever worked as intended, but it looks like the original contributor's code hasn't changed since s/he wrote it: https://github.com/LPTK/intellij-scala/commit/7870080


jamesbrain:
A fix for this issue is coming for nightly scala plugin builds for IDEA 2017.1 within the next few days. (https://github.com/JetBrains/intellij-scala/commit/c5b58d2028b161549e2c0369ebc56195e75801ee)

PR #365, opened 2016-09-24
SCL-10678 replace match { case true => ??? ; case false => ???} with if

wbars:
Hi to everyone! Waiting for any feedback.

Also, was not very clear from issue, how to handle match statements with expression of non-boolean type. I decided not highlight that case at all. Am i correct?


wannabearockstar:
Hi to everyone! Waiting for any feedback.

niktrop:
Hi, thanks for contributing!
You are right, this inspection should be available only if expression is of boolean type. Also, you need to check that there is no guards in case clauses:

```
b match {
  case true if b1 => ??? // inspection is not applicable
  case false => ???
}
```

I think that it is natural to also handle the case when second branch is a wildcard one, but you may leave it to someone else:

```
b match {
  case true => ???
  case _ => ???
}
```

wbars:
@niktrop i've just update PR, guard and wildcard processing included.

niktrop:
Hi, I merged your commits with some cosmetic changes: https://github.com/JetBrains/intellij-scala/commit/7f7f3976ae725ffbbbbab994c3acc6a0cbff4b91

Changes are mainly about using patterns instead of things like `option.get` and `tuple._1`.

Also, I decided to highlight only `match` keyword, because several yellow lines don't look very nice.
Thank you for your work!

PR #370, opened 2017-01-22
README improvements

holograph:


jastice:
Thank you! 
Also note that our current main development branch is now 171.x

holograph:
It is? Hmm. Do I need to bisect our dev branch? If you have any advice that'll make sure we're tracking your Git conventions, I'd be delighted.

jastice:
We create a development branch `idea<version>.x` per IDEA major release when it's in the EAP stage. So if you are developing against the upcoming release use the `idea171.x` one. Otherwise you can work on the old one and merge into the new one later. Depending on what you want to work on that may be straightforward or messy :)

holograph:
Our developers that'll end up working on the plugin are liable to use nightlies (and likely EAP version of IDEA too), and anyway we'd prefer to code against the most recent version. Anyway I'll refocus on that branch when I'm done with my current context. Thanks!

PR #373, opened 2017-03-04
Refactoring: lazy val in hashCode() instead of var

evis:
Just noticed "noinspection HashCodeUsesVar" comment and decided to fix it :)

### DIFF
--- a/src/org/jetbrains/plugins/scala/lang/psi/types/api/TypeParameterType.scala
+++ b/src/org/jetbrains/plugins/scala/lang/psi/types/api/TypeParameterType.scala
@@ -101,13 +101,8 @@ case class TypeParameterType(arguments: Seq[TypeParameterType],
 
   override val name = nameAndId._1
 
-  private var hash: Int = -1
-
-  override def hashCode: Int = {
-    if (hash == -1) {
-      hash = Seq(name, arguments, lowerType, upperType, psiTypeParameter).map(_.hashCode()).foldLeft(0)((a, b) => 31 * a + b)
-    }
-    hash
+  override lazy val hashCode: Int = {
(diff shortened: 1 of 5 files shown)


niktrop:
Hi! 
Thanks for PR, but  I cannot merge it. These `hashCode`s were converted from usual `def` like this for performance reasons. We use these types in caches a lot, so `hashCode` methods are very hot. Lazy vals have some overhead comparing to current implementation so it is not clear how it will affect overall performance.

src/org/jetbrains/plugins/scala/lang/psi/types/api/TypeParameterType.scala
```diff
@@ -101,13 +101,8 @@ case class TypeParameterType(arguments: Seq[TypeParameterType],
 
   override val name = nameAndId._1
 
-  private var hash: Int = -1
-
-  override def hashCode: Int = {
-    if (hash == -1) {
-      hash = Seq(name, arguments, lowerType, upperType, psiTypeParameter).map(_.hashCode()).foldLeft(0)((a, b) => 31 * a + b)
```
jastice, line 108: @niktrop should we expect a performance impact from seq/map/fold object allocations here or is it negligible because it's a one-time cost?

niktrop: I think I tried it both ways and it didn't have a difference.

jastice:
Just for reference, by the way the current lazy val implementation is described [in SIP-20](http://docs.scala-lang.org/sips/pending/improved-lazy-val-initialization.html), I wouldn't expect a major performance impact by using lazy vals, unless the initial calculation is a significant part of the cost because of the synchronized block, or the additional variable matters. Perhaps worth a look sometime?

niktrop:
It is not obvious and requires at least some verification. Probably I will return to it later.

PR #374, opened 2017-03-11
Don't mark final modifier on constant expressions as redundant

markus1189:
YouTrack Issue: https://youtrack.jetbrains.com/issue/SCL-11500
Consider the following code:

```
final class Foo {
  final val x = "" // final is NOT redundant
}
```

Currently Intellij marks the val `x` as `'final' is redundant for objects and final class members`.
This is wrong however because according to the Scala Language Specification 4.1 (http://www.scala-lang.org/files/archive/spec/2.12/04-basic-declarations-and-definitions.html):

> A constant value definition is of the form
> ```
> final val x = e
> ```
> where e is a constant expression. The final modifier must be present and no type annotation may be given. References to the constant value x are themselves treated as constant expressions; in the generated code they are replaced by the definition's right-hand side e.

So in the above code example the `final` modifier is not redundant on `x`, as long as there is no type annotation.

This removes the warning iff it is a `final val` without type annotations and adds a test suite for `ModifierChecker`.

### DIFF
--- a/src/org/jetbrains/plugins/scala/annotator/modifiers/ModifierChecker.scala
+++ b/src/org/jetbrains/plugins/scala/annotator/modifiers/ModifierChecker.scala
@@ -93,8 +93,8 @@ private[annotator] object ModifierChecker {
                   }
                 case e: ScMember if e.getParent.isInstanceOf[ScTemplateBody] || e.getParent.isInstanceOf[ScEarlyDefinitions] =>
                   val redundant = (e.containingClass, e) match {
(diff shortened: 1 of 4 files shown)


hseeberger:
+1

niktrop:
I merged your PR manually a weed ago, but forgot to update here: https://github.com/JetBrains/intellij-scala/commit/26ef1f7913533c9c97b1b68f71a9228474cb9ca6

Thanks for contribution!

PR #375, opened 2017-03-28
Displaying type names in ComparingUnrelatedTypesInspection.

greenjoe:
The inspection description now says which types cannot be compared so that it's easier to find an error, especially when comparing two types with same name but different packages.
ScalaInspectionTestBase methods that check if there are errors are slightly changed so that it is possible to test for inspections with non-constant description.

It looks like in some cases from tests the type description looks a bit strange (`B.type` when B is an object or `a.type` when comparing with `isInstanceOf`) but I believe it is still much better than not having types in descriptions at all.


niktrop:
I merged your fix manually with a couple of changes: https://github.com/JetBrains/intellij-scala/commit/7f9428b170a368eac5accdd51eb46cd8efbe9181
I have added extracting singleton types to the `generateComparingUnrelatedTypesMsg` method and a test for the case when types have the same simple name.
Very nice fix, thank you.

PR #376, opened 2017-04-04
Fix for SCL-11344 (apply inherited from generic trait)

greenjoe:
* I modified lines that handled curly brace imports in ScImportStmtImpl but appeared to have completely no effect (boolean result was ignored) - I believe I made the behavior similar to singular imports
* Added test for SCL-11344 in ApplyTest.scala

I've spent a few days trying to find the source of this bug and I am still not sure whether it is a good way to fix it and if there might be negative performance implications. I don't think I can do any better though so I thought it might be a good idea to just create a pull request and find out if the change makes any sense :-) The lines I changed in ScImportStmtImpl seemed really suspicious, so even if the fix is not appropriate I believe it's worth to take a look at the code there :-)


niktrop:
Thank you for pointing to the issue. I imagine that it was hard to find it. You may have a look to the fix: https://github.com/JetBrains/intellij-scala/commit/d8d1a54be190ff5095be6d1c06742e196e85d4df

PR #377, opened 2017-04-10
SCL-10175 - Filtered Option

mattfowler:
Adds an inspection to convert:

`if(f(x)) Option(x) else None`

to:

`Option(x).filter(f)`

### DIFF
--- a/src/org/jetbrains/plugins/scala/codeInspection/collections/IfElseToFilteredOptionInspection.scala
+++ b/src/org/jetbrains/plugins/scala/codeInspection/collections/IfElseToFilteredOptionInspection.scala
@@ -14,20 +14,28 @@ object FilterOption extends SimplificationType {
   override def hint: String = InspectionBundle.message("ifstmt.to.filteredOption")
 
   override def getSimplification(expr: ScExpression): Option[Simplification] = expr match {
-    case ex@IfStmt(ScMethodCall(method, Seq(methodArg)), scalaSome(someArg), scalaNone()) =>
-      replaceIfEqual(ex, method, methodArg, someArg)
-    case ex@IfStmt(ScMethodCall(method, Seq(methodArg)), scalaOption(optionArg), scalaNone()) =>
-      replaceIfEqual(ex, method, methodArg, optionArg)
+    case ex@IfStmt(ScMethodCall(method, Seq(methodArg)), some@scalaSome(_), scalaNone()) =>
+      replaceIfEqual(ex, method, methodArg, some)
+    case ex@IfStmt(ScMethodCall(method, Seq(methodArg)), option@scalaOption(_), scalaNone()) =>
+      replaceIfEqual(ex, method, methodArg, option)
     case _ => None
   }
 
   private def replaceIfEqual(expression: ScExpression,
                              methodCall: ScExpression,
                              methodArgument: ScExpression,
-                             optionArgument: ScExpression): Option[Simplification] = {
-    if (methodArgument.getText == optionArgument.getText)
-      Some(replace(expression).withText(s"Option(${methodArgument.getText}).filter(${methodCall.getText})"))
-    else
-      None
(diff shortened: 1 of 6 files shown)


pavelfatin:
Thanks!

It seems that we need to [refine](https://youtrack.jetbrains.com/issue/SCL-10175#comment=27-2135744) the transformation rules.

mattfowler:
Good catch!  I've updated the tests and implementation to use `Some` when appropriate.

src/org/jetbrains/plugins/scala/codeInspection/collections/IfElseToFilteredOptionInspection.scala
```diff
@@ -0,0 +1,41 @@
...
+                             methodArgument: ScExpression,
+                             option: ScExpression): Option[Simplification] = {
+    val replaceWith = getReplacement(_: String, expression, methodArgument, methodCall)
+    option match {
+      case scalaSome(arg) if argsEqual(methodArgument, arg) => Some(replaceWith("Some"))
+      case scalaOption(arg) if argsEqual(methodArgument, arg) => Some(replaceWith("Option"))
+      case _ => None
+    }
+  }
+
```
niktrop: It is better to compare elements with com.intellij.codeInsight.PsiEquivalenceUtil.areElementsEquivalent. It will work if there are some differences in whitespaces or comments.

test/org/jetbrains/plugins/scala/codeInspection/collections/IfElseToFilteredOptionInspectionTest.scala
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  def testShouldReplaceWhenReturningOption() = {
+    doTest(
+      s"$evenFunction ${START}if (isEven(2)) Option(2) else None$END",
+      s"$evenFunction if (isEven(2)) Option(2) else None",
+      s"$evenFunction Option(2).filter(isEven)"
+    )
+  }
+
```
niktrop: We need to be sure that result expression is equivalent to the old one. It is not the case if argument has side effects, because it is computed once instead of 2 times. We have org.jetbrains.plugins.scala.util.SideEffectsUtil.hasNoSideEffects for such cases.

src/org/jetbrains/plugins/scala/codeInspection/collections/IfElseToFilteredOptionInspection.scala
```diff
@@ -0,0 +1,41 @@
...
+  private def argsEqual(firstArg: ScExpression, secondArg: ScExpression) = firstArg.getText == secondArg.getText
+
```
niktrop: It is better to compare elements with com.intellij.codeInsight.PsiEquivalenceUtil.areElementsEquivalent. It will work if there are some differences in whitespaces or comments.

test/org/jetbrains/plugins/scala/codeInspection/collections/IfElseToFilteredOptionInspectionTest.scala
```diff
@@ -0,0 +1,46 @@
...
+  def testShouldReplaceWithMethodCallAsParam() = {
+    val getInt = "def getInt() = 2;"
+    doTest(
+      s"$evenFunction $getInt ${START}if (isEven(getInt())) Option(getInt()) else None$END",
```
niktrop: We need to be sure that result expression is equivalent to the old one. It is not the case if argument has side effects, because it is computed once instead of 2 times. We have org.jetbrains.plugins.scala.util.SideEffectsUtil.hasNoSideEffects for such cases.

mattfowler:
@niktrop thanks!  PR updated with your suggestions and a new test for the whitespace case.

niktrop:
Thanks for contribution, your changes were merged manually:
afa156572ddb81ce7c9c06b3d032b274ea3e7302

PR #378, opened 2017-04-29
Implement feature https://youtrack.jetbrains.com/issue/SCL-10329

Susak:


niktrop:
I think inspection should work the other way around, i.e. replace `update` call with an assignment. You can add an intention to do reverse transformation. Also, you need to add an inspection description and some tests (see, for example a9fdd59e88fb06fc8a29871fb778a233044d88f4).

PR #379, opened 2017-05-31
Fix method invocation with reflection on static method.

chaoren:


chaoren:
From [here](https://docs.oracle.com/javase/8/docs/api/java/lang/reflect/Method.html#invoke-java.lang.Object-java.lang.Object...-):

```
public Object invoke(Object obj,
                     Object... args)
              throws IllegalAccessException,
                     IllegalArgumentException,
                     InvocationTargetException
```
> If the underlying method is static, then the specified obj argument is ignored. It may be null.

niktrop:
Thanks! Your fix was merged manually:
57bb5ef92717500148ca1a866d7183d8ffdc2561

PR #380, opened 2017-06-08
Scala meta annotations fix for scala.jvm and scala.js cross projects

raulraja:
Fixes https://youtrack.jetbrains.com/issue/SCL-11817

Includes the project classpath instead of the module dependencies when loading scalameta annotations.
Before this fix scalameta annotations don't work in cross projects targeting scala.jvm and scala.js because the `shared` folder is seen as a source directory and it's out directory does not include the compiled scala meta annotations.

More context on how this fix was derived can be found here https://gitter.im/JetBrains/intellij-scala?at=59353456cf9c13503c554b22


mutcianm:
lgtm

PR #385, opened 2017-08-22
Add UnresolvedImplicitParametersInspection

loskutov:


niktrop:
We need to check that this inspection do not introduce new errors.

niktrop:
This was implemented as part of the annotator.

PR #387, opened 2017-08-24
Fixes "ideaRunner/run" settings in build.sbt

daniel-kullmann:
Hardcoded values for idea.system.path and idea.config.path were wrong, as well as the value for plugin.path.


jastice:
Thanks for this!

PR #388, opened 2017-08-24
Macro support for scalaz-deriving

fommil:
Will close https://github.com/fommil/stalactite/issues/1

This adds IntelliJ support for the `@deriving` macro https://github.com/fommil/stalactite/blob/30156ed07a9b0441b884edfbbd696092008830c2/src/main/scala/stalagtite/deriving.scala

I have created the boilerplate but I have a few questions where I'd really appreciate some help:

- [x] how do I get the list of parameters of the annotation?
- [x] can I get the resolved FQNs of those parameters?
- [x] how can I write a test? (e.g. a file doesn't have red squigglies)
- [x] `updateClassifiers` is not attaching the `intellij-community` sources or javadocs. These would be extremely useful.
- [ ] is there a way to detect if a method has been declared `implicit`?
- [x] I have had to parse the annotation manually with `.getText` because the parameters are not showing up when I use `.getParameterList` and look at the attributes. Is this a bug or am I just not using it correctly?
- [ ] how do I get the FQN of a `ScExpression` on a `ScAnnotation.annotationExpr.getAnnotationParameters` ?

BTW, anyone using ensime to hack on this repo can use this setup in `ensime.sbt`

```
// ;packagePluginCommunity ;testOnly *Stalactite*
// packagePluginCommunityZip
scalacOptions += "-Ywarn-unused-import"
ensimeIgnoreScalaMismatch in ThisBuild := true
ensimeScalaVersion in ThisBuild := "2.11.11"
// https://github.com/ensime/ensime-sbt/issues/353
ensimeUnmanagedSourceArchives in LocalProject("scalaCommunity") :=
   Seq(file(s"./SDK/ideaSDK/${ideaBuild.value}/sources.zip"))
```

I've made a prerelease of this plugin available at https://github.com/fommil/stalactite/releases/tag/v0.0.2

### DIFF
--- a/src/org/jetbrains/plugins/scala/lang/psi/impl/toplevel/typedef/StalactiteInjector.scala
+++ b/src/org/jetbrains/plugins/scala/lang/psi/impl/toplevel/typedef/StalactiteInjector.scala
@@ -1,0 +1,57 @@
+package org.jetbrains.plugins.scala.lang.psi.impl.toplevel.typedef
+
+import org.jetbrains.plugins.scala.lang.psi.api.toplevel.typedef._
+import org.jetbrains.plugins.scala.lang.psi.api.expr.ScAnnotation
+
+/**
+ * Support for https://github.com/fommil/stalactite
+ *
+ * Chooses to skip typeclass derivation (and requirements) in order to
+ * provide a much faster editing experience. Users may experience
+ * deriving failures when they run the real compiler.
+ *
+ * @author Sam Halliday
+ * @since  24/08/2017
+ */
+class StalactiteInjector extends SyntheticMembersInjector {
+  // fast check
+  def hasStalactite(source: ScTypeDefinition): Boolean =
+    source.findAnnotationNoAliases("stalactite.deriving") != null
+
+  // slower, more correct
+  def stalactite(source: ScTypeDefinition): ScAnnotation = {
+    val candidates = source.annotations("stalactite.deriving")
+    if (candidates.size != 1)
+      throw new IllegalArgumentException("expected a single @deriving annotation")
+    candidates.head
+  }
+
+  // so annotated sealed traits will generate a companion
+  override def needsCompanionObject(source: ScTypeDefinition): Boolean =
+    hasStalactite(source)
+
+  // add implicits to case object / case class companions
+  override def injectFunctions(source: ScTypeDefinition): Seq[String] =
+    source match {
+      case cob: ScObject if hasStalactite(cob) =>
+        genImplicits(cob.name + ".type", stalactite(cob))
+      case obj: ScObject =>
+        obj.fakeCompanionClassOrCompanionClass match {
+          case clazz: ScTypeDefinition if hasStalactite(clazz) =>
+            genImplicits(clazz.name, stalactite(clazz))
+          case _ => Nil
+        }
+      case _ => Nil
+    }
+
+  def genImplicits(clazz: String, stalactite: ScAnnotation) = {
+    stalactite
+      .annotationExpr
+      .getAnnotationParameters
+      .map { param =>
+        val typeclass = param.getText
+        val gen = param.getText // would be more correct to use the FQN
+        s"implicit def `$gen`: $typeclass[${clazz}] = _root_.scala.Predef.???"
+      }
+  }
+}
--- a/src/org/jetbrains/plugins/scala/lang/psi/impl/toplevel/typedef/StalactiteInjector.scala
+++ b/src/org/jetbrains/plugins/scala/lang/psi/impl/toplevel/typedef/StalactiteInjector.scala
@@ -19,12 +19,8 @@ class StalactiteInjector extends SyntheticMembersInjector {
     source.findAnnotationNoAliases("stalactite.deriving") != null
 
   // slower, more correct
-  def stalactite(source: ScTypeDefinition): ScAnnotation = {
-    val candidates = source.annotations("stalactite.deriving")
-    if (candidates.size != 1)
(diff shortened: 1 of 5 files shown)


fommil:
I found https://github.com/joewalnes/idea-community/blob/master/java/openapi/src/com/intellij/psi/PsiAnnotation.java

But I'm really struggling with testing this... it would be exceptionally useful to be able to run this code against some source code.

fommil:
FYI, I created a zip of the sources and added them to ensime like this

```scala
ensimeIgnoreScalaMismatch in ThisBuild := true
ensimeScalaVersion in ThisBuild := "2.11.11"
ensimeUnmanagedSourceArchives in LocalProject("scalaCommunity") := Seq(file("../intellij-community/java/java-psi-api/src/psi.zip"))
```

fommil:
I found the monocle tests!

src/org/jetbrains/plugins/scala/lang/psi/impl/toplevel/typedef/StalactiteInjector.scala
```diff
@@ -0,0 +1,53 @@
...
+ * Chooses to skip typeclass derivation (and requirements) in order to
+ * provide a much faster editing experience. Users may experience
+ * deriving failures when they run the real compiler.
+ *
+ * @author Sam Halliday
+ * @since  24/08/2017
+ */
+class StalactiteInjector extends SyntheticMembersInjector {
+  def stalactite(source: ScTypeDefinition): PsiAnnotation =
+    source.findAnnotationNoAliases("stalactite.deriving")
```
ilinum: I think you can do `source.annotations("stalactite.deriving")`, which returns a `Seq[ScAnnotation]`

Using `ScAnnotation` (stands for Scala Annotation) instead of `PsiAnnotation` should help you with your problems in `genImplicits`.

jastice:
> updateClassifiers is not attaching the intellij-community sources or javadocs. These would be extremely useful

The IntelliJ platform artifacts are published in a kind of non-standard way. To attach sources, open whatever class, choose the "attach sources" dialog, navigate to <project dir>/SDK/ideaSDK/<sdk version>/sources.zip and choose "all". Usually works.

fommil:
@jastice LOL, like I'm using IntelliJ...

```
find sources.zip
find: ‘sources.zip’: No such file or directory
```

How can I get the build to download this?

jastice:
Maybe you'll make an exception for writing an IntelliJ plugin? :P 
The sdk should be automatically downloaded and extracted. 
Try `find -name sources.zip` (and never trust `find`, I always need several iterations to get its syntax right)

fommil:
omg, I haven't had my caffeine yet... I don't know how to use unix commands.

fommil:
The ensime equivalent of pointy clicky above:

```
ensimeUnmanagedSourceArchives in LocalProject("scalaCommunity") := Seq(file("./SDK/ideaSDK/172.3757.52/sources.zip"))
```

is there a setting to give me the directory, or at least the version of the ideasdk that is being used?

jastice:
`ideaBuild` is the setting for the idea version used in the build
`ideaBaseDirectory` is where the sdk gets put
These settings are provided by the sbt-idea-plugin

fommil:
awesome. Unfortunately, this infinite trail of bugs is leading me to a bug in ensime... https://github.com/ensime/ensime-sbt/issues/353

So i'll use the hardcoded one for now

jastice:
`ideaBuild` is a setting so you can base other settings on it

src/org/jetbrains/plugins/scala/lang/psi/impl/toplevel/typedef/StalactiteInjector.scala
```diff
@@ -0,0 +1,57 @@
...
+class StalactiteInjector extends SyntheticMembersInjector {
+  // fast check
+  def hasStalactite(source: ScTypeDefinition): Boolean =
+    source.findAnnotationNoAliases("stalactite.deriving") != null
+
+  // slower, more correct
+  def stalactite(source: ScTypeDefinition): ScAnnotation = {
+    val candidates = source.annotations("stalactite.deriving")
+    if (candidates.size != 1)
+      throw new IllegalArgumentException("expected a single @deriving annotation")
```
ilinum: What should happen in the following scenario (two annotations)?
```
import stalactite._

@deriving(Encoder, Decoder)
@deriving(Encoder, Decoder)
case class Bar(s: String, b: Boolean)
```

I'm not sure what the right behavior should be but throwing an error doesn't seem like the right solution.
 
If the code above would cause a compile error, it would be better to report it but it might be too hard for this PR.
If the code above would not cause a compile error, we should try to mimic stalactite's behavior as much as we can.

fommil: That's not allowed. One `@deriving` only... although I'm not 100% sure what macro-paradise does when multiple macro annotations are added to a class.

ilinum: So I think the best thing would be to highlight it as an error but I don't think there's an easy way to do this here. For now, I just would not throw an error

fommil: I'll just take the first one, how about that?

niktrop:
Hi, Sam! I merged your PR manually with some changes:
https://github.com/JetBrains/intellij-scala/commit/8dd259c0dbca3ef8819112108bedaeb8465dc6f1

It is available in nightly build 2017.2.612.

PR #389, opened 2017-08-25
define build parameters (java version, heap size, sbt version)

fommil:
actual values are up for discussion, but I think they should be defined because it OOMs on `compile` on a fresh checkout using the 1g default.


jastice:
As discussed on gitter, the settings might have to differ based on whether someone is using 32bit or 64bit jdk, and as the maximum for 32bit is 2G, and for 64bit 2G might not be enough to compile. Will need a little experimentation

fommil:
also copying gitter thoughs here for others... if your CI to publish needs more, then maybe a simple step to delete `.jvmopts` would be all that is needed. I think it's important for a fresh checkout of this repo to compile without OOM for contributors.

ijuma:
Do developers still use 32 bit JDKs? In any case, the overhead of 64 bits is limited due to compressed references

niktrop:
2G is enough on 64 bit WIndows. I merged this manually:
https://github.com/JetBrains/intellij-scala/commit/4fca2bfe7b0821e350b3cd8638494a374a8564b9

PR #391, opened 2017-08-28
Add support for highlighting implicit usages

loskutov:

PR #392, opened 2017-08-30
Add support for implicit usages search

loskutov:
Requires #391 to be applied first.


niktrop:
This feature will be reimplemented with bytecode analysis.

PR #394, opened 2017-09-12
fixes typo in ScalaTypeAnnotationSettings

rladstaetter:

PR #395, opened 2017-09-16
memeber → member

ForNeVeR:
Fixed a couple of small typos: one in the user-visible message, and other ones in the variable names (were found accidentally).


ForNeVeR:
(Almost) the same has been suggested in #394, but I'm not sure I'm following your branch management scheme. Please close if not appropriate.

niktrop:
I've merged #394 in both idea172 and idea173 branches, so your fix is not necessary.

ForNeVeR:
Okay, thanks!

PR #400, opened 2017-10-02
SCL-12301: upgrade embedded scalastyle to 1.0.0

kitbellew:


niktrop:
We do not plan to make new releases of scala plugin for IDEA 2017.2.

PR #407, opened 2017-10-27
SCL-8320 Collection inspection: replace map.lift(n) with map.get(n)

t-kameyama:
https://youtrack.jetbrains.com/issue/SCL-8320


niktrop:
Thank you for contribution! Your PR is already in master:
https://github.com/jetbrains/intellij-scala/commit/e557c0157a240d8a3edc5967290daf9027b65650

PR #408, opened 2017-10-30
CBT fixes

darthorimar:
* SCL-12668, using project Scala version on build class create
* Fix strange behavior of CbtExtraModule and CbtTestModule in project structure
* Add link to CBT doc in a new CBT project description
* Show run CBT task markers only on main CBT Build class


jastice:
thanks again!

PR #410, opened 2017-11-06
Add support for Hydra compiler

maris123:
We added support for Hydra Compiler and now the user can select if he wants to use the Hydra compiler instead of the vanilla compiler.
Hydra Settings Panel is hidden by default and we added an action that enables/disables the Hydra Settings Panel:
action
This is how the Hydra Settings panel looks like:
settingspanel

Thank you Pavel and Nikolay for your support.


dragos:
We should probably squash all commits on merge, but we kept history in case it's interesting for discussion.

scala/compiler-jps/src/org/jetbrains/jps/incremental/scala/data/CompilationData.scala
```diff
@@ -77,11 +79,25 @@ object CompilationData {
 
       val canonicalSources = sources.map(_.getCanonicalFile)
```
niktrop: It would be better to extract hydraOptions logic to a separate method.

scala/compiler-jps/src/org/jetbrains/jps/incremental/scala/data/CompilerData.scala
```diff
@@ -19,16 +20,46 @@ import scala.collection.JavaConverters._
 case class CompilerData(compilerJars: Option[CompilerJars], javaHome: Option[File], incrementalType: IncrementalityType)
 
 object CompilerData {
+  private val Log: JpsLogger = JpsLogger.getInstance(CompilerData.getClass.getName)
+
   def from(context: CompileContext, chunk: ModuleChunk): Either[String, CompilerData] = {
     val project = context.getProjectDescriptor
     val target = chunk.representativeTarget
     val module = target.getModule
```
niktrop, line 30: Same here, I think hydra logic may be separated better.

scala/scala-impl/src/org/jetbrains/plugins/hydra/actions/EnableHydraSettingsAction.scala
```diff
@@ -0,0 +1,31 @@
...
+
+import com.intellij.openapi.actionSystem.{AnActionEvent, CommonDataKeys, ToggleAction}
+import com.intellij.openapi.project.Project
+import org.jetbrains.plugins.hydra.compiler.HydraCompilerSettings
+
+/**
+  * @author Maris Alexandru
+  */
+class EnableHydraSettingsAction extends ToggleAction{
+
```
niktrop: After trying this out, I think that this action should open Hydra settings, and not toggle the flag. I suggest to have `isHydraSettingsEnabled` to be false by default, and set it to true when this action is called for the first time.

niktrop: Also, this setting may be application-wide.

scala/scala-impl/src/org/jetbrains/plugins/scala/compiler/CompileServerManager.scala
```diff
@@ -206,4 +206,8 @@ object CompileServerManager {
   def showCompileServerSettingsDialog(project: Project): Unit = {
     ShowSettingsUtil.getInstance().showSettingsDialog(project, "Scala Compile Server")
   }
+
```
niktrop: This method may be moved to `hydra` package

scala/scala-impl/src/org/jetbrains/plugins/scala/project/Versions.scala
```diff
@@ -25,37 +28,67 @@ object Versions  {
...
   private def loadVersionsFrom(url: String, filter: PartialFunction[String, String]): Try[Seq[String]] = {
     loadLinesFrom(url).map { lines => lines.collect(filter) }
   }
 
-  private def loadLinesFrom(url: String): Try[Seq[String]] = {
+  private def loadHydraVersionsFrom(url: String, login:String, password: String, filter: PartialFunction[String, String]): Try[Seq[String]] = {
+    val loadedLines = loadLinesFrom(url){ connection => connection.setRequestProperty("Authorization", "Basic " + HydraCredentialsManager.getBasicAuthEncoding()) }
+    loadedLines.map { lines => lines.collect(filter) }
+  }
+
```
niktrop: Implicit parameter is redundant here, default value is used on call site.

scala/scala-impl/src/org/jetbrains/plugins/scala/project/template/Downloader.scala
```diff
@@ -62,6 +70,16 @@ object Downloader {
       s"""set libraryDependencies := Seq("ch.epfl.lamp" % "dotty_2.11" % "$version" % "scala-tool")""",
       "updateClassifiers")
   }
+
+  private def sbtCommandsForHydra(repositorySettings: HydraRepositorySettings)(platform: Platform, version: String) = {
+    Seq(
+      s"""set scalaVersion := "${version.split("_")(0)}"""",
```
niktrop: Password containing `"` breaks parsing. 😄

niktrop (commented):
You did a very good job separating hydra-related logic. I suggest also to move it from `Versions` and `Downloader` objects. You may make methods you need from them public, thats not a problem.

scala/scala-impl/src/org/jetbrains/plugins/hydra/actions/EnableHydraSettingsAction.scala
maris123: How can I open the settings panel from the Action? Do I have to use Dialog Wrapper?

niktrop: I think `showHydraCompileSettingsDialog` method should work.

maris123: yes. I forgot that I used that for the notifcation also. Thanks

dragos:
@niktrop Please have another look, I think @maris123 addressed all comments.

niktrop:
LGTM
I've added a couple of minor fixes:
https://github.com/JetBrains/intellij-scala/commits/idea173.x-hydra-integration
It should be merged after passing tests on our CI.

dragos:
@niktrop It looks like one of the changes (your last commit) failed compilation 😢

PR #411, opened 2017-11-08
SCL-9625 Inspection: Replace Some(a) with Option(a)

t-kameyama:
Fix for [SCL-9625](https://youtrack.jetbrains.com/issue/SCL-9625)


SrTobi:
I merged this in 27e17f9f7caddd72830cb22b9277829054c84579, but I disabled it by default in c898104ff4dde74ee26652cad58f2f932d8b06ff. Thanks @t-kameyama

PR #414, opened 2017-11-09
Fix IfElseToFilteredOptionInspection shortName

t-kameyama:


SrTobi:
Hmm... I can merge this. The question is, if it is worth fixing the typo in comparison to that we would destroy every disabled warning comment for users downstream...

niktrop:
@SrTobi Good catch, I think it's better to keep it as is. But you also have to make the same typo in `inspectionDescriptions/IfElseToFilteredOption.html` :)

PR #416, opened 2017-11-10
SCL-11892 Collection inspection: convert seq.exists(_ => true) to seq.nonEmpty

t-kameyama:
https://youtrack.jetbrains.com/issue/SCL-11892


SrTobi:
@t-kameyama thanks!

PR #417, opened 2017-11-11
#SCL-11156 fixed

evgenynacu:
See fix and provided SingleAbstractMethodTestBase.testSCL11156
after some research I came to conclusion there is no correct ScSubstitutor for ScDesignatorType's. No substitutor is provided, but it has to be provided. (for example, for ScDesignatorType Specific)


evgenynacu:
Unfortunately, my change breaks many tests.
Pls, review my findings: https://youtrack.jetbrains.com/issue/SCL-11156

PR #418, opened 2017-11-12
SCL-12907 fix

piter75:
This PR is an attempt at fixing broken Ivy cache indexing by:

- not filtering ivy-x.y.z.xml files based on modification date but rather on existence in plugin's cache
- indexing classes in artifacts' jar files only when the artifact has not been indexed yet
- **merging** newly found versions with those that are already stored in the index or putting it in when there was none stored yet

Additionally in this PR implicit ProjectContext is removed from ResolverIndex to make the interface cleaner and easier to use.
ProjectContext is now passed through concrete constructor where applicable


mutcianm (approved):
Thanks for your contribution! I'll merge it manually as a patch

mutcianm:
Merged as 5953afd71a0c65490e252010698f66c8560a7fac

PR #420, opened 2017-11-15
Cbt process interopt improvements

darthorimar:
Use IJ task api for doing all CBT stuff now
Should fix SCL-12870, SCL-12829
Also, temporary disables giter8 template engine support
@jastice, please, take a look


cbt/src/org/jetbrains/plugins/cbt/process/CbtProcess.scala
```diff
@@ -1,76 +1,109 @@
...
+    finished.down()
+
+    val listener = new CbtProcessListener {
+      val textBuilder = new StringBuilder
+      var success = false
+
+      override def onComplete(exitCode: Int): Unit = {
+        if (exitCode == 0)
+          success = true
+        Thread.sleep(500)
```
jastice: what is the reason for the sleep? can this be achieved in a more robust way?

jastice (approved):
lgtm, but we can probably improve the Thread.sleep

darthorimar:
@jastice you are right, the is no reason to sleep here now, fixed :)

PR #422, opened 2017-11-21
prepare for new scalaz-deriving update

fommil:
I am planning a large update to `scalaz-deriving` (formerly known as `stalactite`) and this adds support for both the legacy annotation `@stalactite.deriving`, the new annotation `@scalaz.deriving`, and the upcoming compiler plugin annotation `@deriving` (I would rather use `@scalaz.deriving` but as you may know, compiler plugins do not perform naming of annotations until rather late in the game so such plugins tend to use annotations as keywords. *sigh*)

FYI, hacking on this codebase was again a pleasure and made easier with the following `ensime.sbt` config

```
// ;packagePluginCommunity ;testOnly *ScalazDeriving*
// packagePluginCommunityZip
scalacOptions += "-Ywarn-unused-import"
ensimeIgnoreScalaMismatch in ThisBuild := true
ensimeScalaVersion in ThisBuild := (scalaVersion in LocalProject("scalaCommunity")).value
ensimeUnmanagedSourceArchives in LocalProject("scalaCommunity") :=
   Seq(file(s"./SDK/ideaSDK/${ideaBuild.value}/sources.zip"))
```

### DIFF
--- a/scala/scala-impl/test/org/jetbrains/plugins/scala/lang/macros/ScalazDerivingTest.scala
+++ b/scala/scala-impl/test/org/jetbrains/plugins/scala/lang/macros/ScalazDerivingTest.scala
@@ -16,20 +16,20 @@ import org.junit.Assert._
 import org.junit.experimental.categories.Category
 
 /**
- * IntelliJ's equivalent of stalactite's built-in PresentationCompilerTest
+ * IntelliJ's equivalent of scalaz-deriving's built-in PresentationCompilerTest
  *
  * @author Sam Halliday
  * @since  24/08/2017
  */
 @Category(Array(classOf[SlowTests]))
-class StalactiteTest extends ScalaLightCodeInsightFixtureTestAdapter {
+class ScalazDerivingTest extends ScalaLightCodeInsightFixtureTestAdapter {
 
   override implicit val version: ScalaVersion = Scala_2_11
 
   override def librariesLoaders =
     super.librariesLoaders :+
-      StalactiteTest.StalactiteLoader() :+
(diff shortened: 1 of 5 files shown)


scala/scala-impl/test/org/jetbrains/plugins/scala/lang/macros/ScalazDerivingTest.scala
```diff
@@ -126,17 +126,17 @@ case object <caret>Caz
 
 }
 
-object StalactiteTest {
+object ScalazDerivingTest {
 
-  case class StalactiteLoader() extends IvyLibraryLoaderAdapter {
+  case class ScalazDerivingLoader() extends IvyLibraryLoaderAdapter {
```
jastice, line 131: we can now load test dependencies directly instead of using the testDownloader sbt config. eg in ShapelessTest:
```
override protected def loadIvyDependencies(): Unit =
    DependencyManager("com.chuusai" %% "shapeless" % "2.3.2").loadAll
```

fommil: do you want me to change it?

jastice: just informational for now, but you might as well :)

fommil: I had a quick look and it's not obvious to me how to make the change, if you could just add this to the cleanup backlog I'd really appreciate it. The new form tests look good!

PR #423, opened 2017-11-23
#SCL-12987 - extension point for def macros on StringContexts

janlindemann:
Implementation for [SCL-12987](https://youtrack.jetbrains.com/issue/SCL-12987)


mutcianm:
Thanks for your contribution!
Having this type provider also provide expected types on interpolated parts would also be great.
However, if you don't have enough time for it, I can merge this PR as is.

avesloguzova:
Hi @mutcianm, it would be great to merge it as is.

PR #424, opened 2017-11-27
SCL-9807 Collection inspection, Seq: don’t emulate "map"

t-kameyama:
https://youtrack.jetbrains.com/issue/SCL-9807


SrTobi:
@t-kameyama thanks one last time for your PR!

I looked into it and decided that it was too specific. There are too many similar code snippets that don't work and only changing the formatting slightly prevents the inspection from working.

```
seq.foldLeft(Seq.empty) { (acc, x) => acc :+ f(x) }
```

```
seq.foldLeft(Seq.empty)((acc, x) => acc :+ x)
```

```
seq.foldLeft(Seq.empty)((acc, x) => acc :+ (x + "blub"))
```

So at the moment I am not going to merge it. If you like to make it more general, please go ahead :+1: , otherwise we can close this PR

PR #425, opened 2017-11-28
SCL-9972: Add inspection for typical incorrect usages of Source

liff:
Missing features:

 1. `Source.fromInputStream(new FooInputStream(???)).getLines()`,
 2. quick fix / conversion to `try`/`finally` and
 3. cannot handle renamed method (as in `import scala.io.Source.{fromFile => problemSolved}`.

For 1, what kinds of sources of `InputStream` should be marked? New instances created explicitly with `new` and functions/methods returning an `InputStream`?

2 can perhaps be added later?

3 could maybe be solved by adding a case for renamed methods to `org.jetbrains.plugins.scala.codeInspection.collections.{Qualified, Unqualified}` but perhaps such a case is intentionally left out?


SrTobi:
I merged this in a094191 and reworked it in 13ba01f. Thank you @liff

PR #428, opened 2018-02-02
Allow injection of custom resolvers into test framework.

jameskoch:
First time contributor. Happy to file a YouTrack for this.

Context:

I contribute to an IntelliJ plugin which depends on the Scala plugin (to offer assistance in using a particular DSL). During tests I would like to be able to resolve libraries from alternate repositories, such as an Artifactory instance.

Notes:

1. I considered passing an entire DependencyManager rather than a Seq[Resolver], but that also encodes other aspects like a blacklist which feel best left orthogonal.
1. Parameterization vs. overrides - I've attempted to follow the localized patterns I saw, such as ScalaSdkLoader accepting arguments to control its behavior, as contrasted with DependencyManagerBase which uses subclassing+overrides to control its behavior. I don't love that resolver-mgmt is handled in different ways across these.

### DIFF
--- a/scala/scala-impl/test/org/jetbrains/plugins/scala/DependencyManager.scala
+++ b/scala/scala-impl/test/org/jetbrains/plugins/scala/DependencyManager.scala
@@ -22,7 +22,7 @@ abstract class DependencyManagerBase {
 
   protected val artifactBlackList = Set("scala-library", "scala-reflect", "scala-compiler")
 
-  protected val resolvers = Seq(
+  protected val resolvers: Seq[Resolver] = Seq(
     MavenResolver("central", "http://repo1.maven.org/maven2"),
     MavenResolver("scalaz-releases", "http://dl.bintray.com/scalaz/releases"),
     IvyResolver("typesafe-releases",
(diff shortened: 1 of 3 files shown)


mutcianm:
If I'm understanding correctly, you wold like to provide custom repositories to `ScalaSdkLoader`?
If it's just about third-party libraries in another plugin, why not simply subclass `IvyManagedLoader` and redefine `DepencyManagerBase` inside it?

jameskoch:
You are correct that I need this capability for the `ScalaSdkLoader` as well.

I'm not sure `DependencyManager` can be redefined in a subclass of `IvyManagedLoader` like you proposed because it's a reference to a singleton. I had to double check that, though. I think I've recreated the key elements here - https://scastie.scala-lang.org/noofcigGT9m8GP1q2ph1WQ

`IvyManagedLoader` is only a few lines of code, so it's not that hard to reimplement the whole thing in my project. The `ScalaSdkLoader` is a bit more involved though, so I'd rather get rid of a reimplementation I've added to my code.

mutcianm:
I see your point.
However I'd stick with overriding members instead of changing existing method signatures, especially if this code has external reverse dependencies.

1. Extract `DepencyManagerBase` instance provider into a protected def or val inside `IvyManagedLoader` and `ScalaSDKLoader`
2. In your code define your custom `DepencyManagerBase` implementation with overridden `resolvers` - it's an abstract class btw not an object
3. In your code subclass  `IvyManagedLoader` and `ScalaSDKLoader` overriding the provider with instance[s] of `DepencyManagerBase` you implemented in the previous step

This will save up on a few classes as well as reduce code complexity since we're not planning to dynamically inject custom resolvers in our own test's codebase

jameskoch:
Take another look.

mutcianm:
Great, thanks!

PR #429, opened 2018-02-07
#SCL-13305: Move Refactoring: support moving object members to another

Uria1:


scala/scala-impl/resources/META-INF/scala-plugin-common.xml
```diff
@@ -14,10 +14,12 @@
         <extensionPoint name="evaluatorCompileHelper" interface="org.jetbrains.plugins.scala.debugger.evaluation.EvaluatorCompileHelper"/>
         <extensionPoint name="serverWidgetEP" interface="org.jetbrains.plugins.scala.compiler.ServerWidgetEP"/>
         <extensionPoint name="syntheticMemberInjector" interface="org.jetbrains.plugins.scala.lang.psi.impl.toplevel.typedef.SyntheticMembersInjector"/>
-        <extensionPoint name="scalaTestDefaultWorkingDirectoryProvider" interface="org.jetbrains.plugins.scala.project.maven.ScalaTestDefaultWorkingDirectoryProvider"/>
```
niktrop: Please avoid reformatting or do it in a separate commit to make actual changes more visible.

scala/scala-impl/src/org/jetbrains/plugins/scala/lang/refactoring/move/moveMembers/ScalaMoveMembersHandler.scala
```diff
@@ -0,0 +1,64 @@
...
+          }
+        case _ => false
+      }
+    }
+  }
+
+}
+
+private[moveMembers] object ScalaMoveMembersHandler {
+
```
niktrop, line 40: You can use `PsiTreeUtil.getParentOfType(PsiElement, Class<T>)`  utility method instead.

scala/scala-impl/src/org/jetbrains/plugins/scala/lang/refactoring/move/moveMembers/ScalaMoveMembersHandler.scala
```diff
@@ -0,0 +1,64 @@
...
+      case _ => ancestorObjectOf(element.getParent)
+  }
+
+  @tailrec
+  def ancestorMemberOf(element: PsiElement): Option[ScMember] = element match {
+    case null => None
+    case _: ScMember => Some(element.asInstanceOf[ScMember])
+    case _ => ancestorMemberOf(element.getParent)
+  }
+
```
niktrop, line 54: I think you may just work with `ScMember` here and move the whole `ScPatternDefinition` even if it contains several `ScReferencePattern`s. This way you can get rid of `movedScMembers` map.

scala/scala-impl/src/org/jetbrains/plugins/scala/lang/refactoring/move/moveMembers/ScalaMoveMembersDialog.scala
```diff
@@ -0,0 +1,95 @@
...
+    JavaPsiFacade.getInstance(myProject).findClass(className,  GlobalSearchScope.projectScope(myProject))
+
+  override def createCenterPanel(): JComponent = {
+    val panel = new JPanel(new BorderLayout)
+
+    val box = Box.createVerticalBox
+
+    val _panel1 = new JPanel(new BorderLayout)
+    val sourceClassField = new JTextField
+    sourceClassField.setText(sourceObject.getName)
```
niktrop, line 70: `ScObject.name` will return name without `$`

scala/scala-impl/src/org/jetbrains/plugins/scala/lang/refactoring/move/moveMembers/ScalaMoveMembersDialog.scala
```diff
@@ -0,0 +1,95 @@
...
+
+          override def getSelectedMembers: Array[PsiMember] = findReferencePatterns(initialTargetMember).toArray
+
+          override def getTargetClassName: String = className
+        }))
+      case _ => Messages.showErrorDialog(ScalaBundle.message("move.members.target.must.be.object"), RefactoringBundle.message("error.title"))
+    }
+  }
+
+  private def findClass(className: String) =
```
niktrop, line 60: I think we should avoid `$` at the end of `ScObject` name. You may create `org.jetbrains.plugins.scala.lang.psi.ElementScope` and use `getCachedObject` method to find an object by name without `$`.

scala/scala-impl/src/org/jetbrains/plugins/scala/lang/refactoring/move/moveMembers/ScalaMoveMembersDialog.scala
```diff
@@ -0,0 +1,95 @@
...
+import com.intellij.refactoring.ui.RefactoringDialog
+import com.intellij.ui.{EditorComboBox, JavaReferenceEditorUtil}
+import org.jetbrains.plugins.scala.ScalaBundle
+import org.jetbrains.plugins.scala.lang.psi.api.toplevel.typedef.ScObject
+
+class ScalaMoveMembersDialog(project: Project, canBeParent: Boolean, sourceObject: ScObject, initialTargetMember: PsiMember) extends RefactoringDialog(project, canBeParent) {
+
+  import ScalaMoveMembersHandler._
+
+  private val RECENTS_KEY = "ScalaMoveMembersDialog.RECENTS_KEY"
```
niktrop, line 22: This val is unused.

scala/scala-impl/src/org/jetbrains/plugins/scala/lang/refactoring/move/moveMembers/ScalaMoveMemberHandler.scala
```diff
@@ -0,0 +1,134 @@
...
+        val newMemberInTarget = movedScMembers(memberName(psiMemberToMove))
+        findEquivalentReferencePatternFrom(newMemberInTarget)
+    }
+  }
+
+  def init(): Unit = {
+    movedScMembers.clear()
+  }
+
+  override def decodeContextInfo(scope: PsiElement): Unit = ChangeContextUtil.decodeContextInfo(scope, null, null)
```
niktrop, line 126: `ChangeContextUtil` works only for java elements. You may use `ScalaChangeContextUtil` with similar methods. But `decode` method should be invoked on the copy of element which was used for `encode`. These methods are required to update references inside the moved members. Currently your implementation breaks for this example: 
```
object Source {
  val x = 1

  def xx = x //move it
}

object Target
```

niktrop:
It would be nice to have some tests for this feature. I can write a test class, can you please create a number of test cases in the same format as in `community/scala/scala-impl/testdata/move`?

genrym:
@niktrop thanks for the comments, we are on it

scala/scala-impl/resources/META-INF/scala-plugin-common.xml
genrym: I am not sure why it happened, probably some formatting occurred. Will revert with only the relevant changes

genrym:
@niktrop regarding test cases in `community/scala/scala-impl/testdata/move` - is there a notion of _caret_ like there is in _inline_ tests?

niktrop:
@genrym Tests for Move Class doesn't need it, because a class may be specified by a fully qualified name. It will be useful though if you are going to test moving of overloaded methods.

genrym:
@niktrop understood, in our case I think it is needed. Consider the following case:

object Source {
   val a = 5
   val b = 6
}

object Target {
   val c = 7
}

Moving member _Source.a_ to _Target_ will be available if caret is on _a_

So should I add the notion of `/*caret*/` in test cases?

niktrop:
Both ways are possible: passing a qualified member name as a parameter for a test case method, or marking a member with a /*caret*/ in testdata. The latter approach will allow to test overloaded methods in the same way:
```
object Source {
  def a(s: String) = 5
  def a(i: Int) = 6
}

object Target {
}
```

niktrop:
Updated PR : https://github.com/JetBrains/intellij-scala/pull/437

PR #430, opened 2018-02-08
#SCL-8951: Inline refactoring: support inline function with arguments

genrym:


niktrop:
Great! I've created a branch with a couple of my additions:
https://github.com/JetBrains/intellij-scala/tree/idea181.x-scl-8951
I'm going to add it for 2018.1 release after a bit of testing.

PR #433, opened 2018-03-11
support a new scalaz-deriving annotation

fommil:
v1.0 of https://gitlab.com/fommil/scalaz-deriving/ will introduce a new annotation `@xderiving` that behaves identically to `@deriving` as far as IntelliJ is concerned.

The FQN is `scalaz.xderiving` but because it is implemented as a compiler plugin, `_root_.xderiving` is equally valid (unfortunately).


niktrop:
Merged: https://github.com/JetBrains/intellij-scala/commit/6570572168ce00a705a665ab70ccfa85162a6680

PR #434, opened 2018-03-15
SCL-9859, SCL-9860, SCL-12507: unnecessary parentheses in types, patterns and lambda parameters

oowekyala:
Fixes [SCL-9860](https://youtrack.jetbrains.com/issue/SCL-9860), [SCL-9859](https://youtrack.jetbrains.com/issue/SCL-9859),  [SCL-12507](https://youtrack.jetbrains.com/issue/SCL-12507)

Some implementation "issues" I ran into:
* Is there a way to set the "ignoreClarifyingParentheses" to false during tests?
* ~Code duplication is nasty.~ Edit: b83a2d0 fixes that problem by abstracting the logic for infix nodes and parenthesised nodes into generic traits. Implementors of these traits still have duplicate methods, because they weren't named the same (eg ScParenthesisedPattern has subNode and subpattern). I think the old ones should be removed, or deprecated, I don't know your policy about that. I also removed the usages of UnnecessaryParenthesisUtil, which is superseded by the more general ScGenericParenthesisedNode trait. Again, it should probably be removed or deprecated.
 ~The algorithm to strip parentheses from a type and a pattern are exact copies, except for the types... I tried using a type class but it makes pattern matching impossible because of type erasure AFAIK. I think a better solution would be to make ScParenthesisedType, ScParenthesisedExpr, and ScParenthesisedPattern extend a common trait but the changeset would have become quite big, and probably API breaking. Same for infix types, patterns and expressions, they could probably benefit from having a common trait, and a common extractor~

I didn't add too many tests but here's a set of test cases you can use to test highlighting and the quickfix:
```scala

object RedundantParenthesesInTypes {

  class <<:[A, B]
  class op[A, B]
  class A { type a = B }
  class B extends A
  class C
  class annot extends StaticAnnotation

  val (a): (Int) = (3)

  val _: ((A, A)) => B = _ // redundant
  val _: ((A), A) => B = _ // redundant

  val _: A <<: (B <<: A) = _ // redundant
  val _: (A <<: B) <<: A = _ // required
  val _: (A op B) <<: A = _ // required
  val _: A op (B <<: A) = _ // required
  val _: (A op B) op A = _ // redundant
  val _: A op (B op A) = _ // required
  val _: (A op B op A) = _ // redundant

  val _: A => (A <<: B) => B = _ // clarifying
  val _: A => (A <<: B => B) = _ // redundant
  val _: (A => A) <<: B => B = _ // required
  val _: (A) => (B) => C = _ // redundant
  val _: A => (B => C) = _ // redundant
  val _: (A => B) => C = _ // required
  val _: () => B = _ // required
  val _: (()) => B = _ // redundant
  val _: (() => B) = _ // redundant

  val _: op[A, (_ >: B <: A)] => B = _ // redundant
  val _: op[A, _ <: (B)] => B = _ // redundant
  val _: op[A, (T forSome {type T <: B})] => B = _ // redundant
  val _: (op[A, (T) forSome {type T <: (B)}]) => B = _ // redundant

  val _: A => (A#a) = _ // redundant -- could be clarifying
  val _: A => ((A#a)#a)#a = _ // redundant
  val _: A => (this.type) = _ // redundant

  val _: (A@annot)@annot = _ // redundant
  val _: A op (B@annot) = _ // clarifying

  val _: A with (B@annot) = _ // clarifying
  val _: (A with B)@annot = _ // required

  val _: (A with B@annot) forSome {type T <: B} = _ // clarifying
  val _: A with (B@annot forSome {type T <: B}) = _ // required

  val _: _ <: (A with B) = _ // clarifying
  val _: (_ <: A) with B = _ // required
}

object RedundantParenthesesInPatterns {

  class <<:[A, B]
  class op[A, B]
  trait InfxTest
  case class +!(a: InfxTest, b: InfxTest) extends InfxTest
  case class *!(a: InfxTest, b: InfxTest) extends InfxTest
  case class *!:(a: InfxTest, b: InfxTest) extends InfxTest
  class A extends InfxTest
  class B extends A
  object C
  case class Foo(as: A*)

  val _ = {
    case (a +! b) +! c => _   // redundant
    case a *!: (b *!: c) => _ // redundant
    case a +!: (b *!: c) => _ // clarifying
    case (a *!: b) *!: c => _ // required
    case a +! (b *!: c) => _  // clarifying
    case (a +! b) *! c => _   // required
    case (a +! b *! c) => _   // redundant
    case _: A | (b: B) => _   // redundant
    case (()) => _            // redundant
    case (C, (_: B)) => _     // redundant
    case a @ (Foo(b, c, _*)) => _     // redundant
    case a @ (_ +! _) => _    // redundant
    case (a @ _) +! _ => _    // required
    case (a @ _) | _ +! _ => _    // redundant
    case (null) => _          // redundant
    case s" de d${(de)}" => _ // redundant
  }

  {
    val (a) = ""              // redundant
    val ((a), b) = ((1), 2)   // redundant
  }

}

object RedundantParenthesesInParameters {

  Seq(1) map { (i) => i + 1 }      // redundant
  Seq(1) map { (i: Int) => i + 1 } // redundant
  Seq(1) map ((i) => i + 1)        // redundant
  Seq(1) map ((i: Int) => i + 1)   // required

}
```


niktrop:
You did a great job, thank you! Documentation and tests are especially appreciated.

I understand your aversion to the code duplication, but I don't think changing class hierarchy is a way to go.
It is already quite complicated, and proposed changes to the API are going to be used in very few places.
As an alternative you may just create common extractors for parenthesized and infix elements, and `getPrecedence` may be a method in a type class.

This will give you an instance of an inspection class in tests:
```
InspectionProfileManager.getInstance(project)
  .getCurrentProfile
  .getInspectionTool("ScalaUnnecessaryParentheses", project)
  .getTool
```


When I've been reviewing your code I realized that `isParenthesesNeeded` method may reuse parser instead of duplicating it's logic.
Parentheses are required if and only if removing them changes semantics of the program. In case of types and patterns it may be formulated in terms of psi trees (for expressions there are probably a couple of exceptions). For example, let some psi element have following children:
```
|ParenthesizedPsi
|  PsiA
|PsiB
|PsiC
```
If parentheses are redundant, after removing them and reparsing we should get
```
|PsiA'
|PsiB'
|PsiC'
```
and corresponding elements should be equivalent (`getText` comparison should be enough in this context).

If you want to try this approach, you will need to use `createTypeElementFromText` and `createPatternFromText` from `ScalaPsiElementFactory`. Or just refactor existing logic from psi interfaces.

oowekyala:
>  I don't think changing class hierarchy is a way to go.
It is already quite complicated, and proposed changes to the API are going to be used in very few places.

I understand, though I feel like using type classes would look like a workaround. Especially for ScGenericInfixNode, which I think cuts down much boilerplate without being especially invasive. Plus, there were some usages of UnnecessaryParenthesesUtil around the codebase so that functionality may be more useful than not. One liners like `getParentOfType(element, classOf[AnyParenthesisedNode])` wouldn't be possible without a trait. I think it would be a shame, especially since it's done, but if you're certain...

I fiddled with your parser-based approach, and it works for the largest part but there are some hacky corner cases. For example, the type `(A@annot)@annot` is parsed like the following
```
AnnotatedType
  | TypeInParenthesis
    | AnnotatedType
      | A
      | @annot
  | @annot
```
Which with a naive approach we would simplify as
``` 
AnnotatedType
  | AnnotatedType
    | A
    | @annot
  | @annot
``` 
Even though the inner AnnotatedType should not be here. Same for compound types for example. I think type classes are a better option if we are to drop the traits

> This will give you an instance of an inspection class in tests:

Thanks for the tip!

niktrop:
Annotated and compound types is a good catch, didn't thought about them. 

I'm not 100% against changes for the hierarchy, but your proposal does seem a bit overcomplicated. In particular, I think `TreeMember` trait is unnecessary. 
You may have something like this instead:
```
trait ScGenericInfixNode[E <: ScalaPsiElement] extends ScalaPsiElement {

  def leftOperand: E

  def operation: ScReferenceElement

  def rightOperand: Option[E]

  def isRightAssoc: Boolean = associativity == -1

  def isLeftAssoc: Boolean = !isRightAssoc

  def associativity: Int = InfixExpr.associate(operation.refName)
}


object ScGenericInfixNode {
  def unapply[T <: ScalaPsiElement](arg: ScGenericInfixNode[T]): Option[(T, ScReferenceElement, Option[T])] =
    Some(arg.leftOperand, arg.operation, arg.rightOperand)
}
```

I don't mind renaming existing methods of infix and parenthesized nodes either.

niktrop:
Also you need to sign the contributor license agreement:

oowekyala:
I signed it, thanks :)

About tests, the following returns null:
```scala
InspectionProfileManager.getInstance(project)
  .getCurrentProfile
  .getInspectionTool("ScalaUnnecessaryParentheses", project)
```
I tried to look at 
```scala
InspectionProfileManager.getInstance(project)
  .getCurrentProfile
  .getAllEnabledInspectionTools(project)
```
 and the result is an empty array. Same for `getAllTools`. I would have liked to add a test case for infix precedence, like `val a +: (b *: c) = _ `, though it's not absolutely essential

Shall I rename the members of infix and parenthesised nodes too?

niktrop:
It should work after `setUp`, e.g. in test methods bodies.

I will do it myself. I still think that some of the methods in new traits should be removed from API, but I can fix it later.

oowekyala:
Well after adding an explicit `setUp()` call it fails in the `tearDown()` with
```
java.lang.IllegalArgumentException: Argument for @NotNull parameter 'module' of com/intellij/openapi/roots/ModuleRootModificationUtil.setModuleSdk must not be null
	at com.intellij.openapi.roots.ModuleRootModificationUtil.$$$reportNull$$$0(ModuleRootModificationUtil.java)
	at com.intellij.openapi.roots.ModuleRootModificationUtil.setModuleSdk(ModuleRootModificationUtil.java)
	at org.jetbrains.plugins.scala.base.libraryLoaders.SmartJDKLoader.clean(SmartJDKLoader.scala:41)
	at org.jetbrains.plugins.scala.debugger.ScalaSdkOwner.$anonfun$disposeLibraries$1(ScalaVersion.scala:56)
	at org.jetbrains.plugins.scala.debugger.ScalaSdkOwner.$anonfun$disposeLibraries$1$adapted(ScalaVersion.scala:56)
	at scala.collection.immutable.List.foreach(List.scala:389)
	at scala.collection.generic.TraversableForwarder.foreach(TraversableForwarder.scala:35)
	at scala.collection.generic.TraversableForwarder.foreach$(TraversableForwarder.scala:35)
	at scala.collection.mutable.ListBuffer.foreach(ListBuffer.scala:44)
	at org.jetbrains.plugins.scala.debugger.ScalaSdkOwner.disposeLibraries(ScalaVersion.scala:56)
	at org.jetbrains.plugins.scala.debugger.ScalaSdkOwner.disposeLibraries$(ScalaVersion.scala:55)
	at org.jetbrains.plugins.scala.base.ScalaLightCodeInsightFixtureTestAdapter.disposeLibraries(ScalaLightCodeInsightFixtureTestAdapter.scala:21)
	at org.jetbrains.plugins.scala.base.ScalaLightCodeInsightFixtureTestAdapter.tearDown(ScalaLightCodeInsightFixtureTestAdapter.scala:48)
        ...
```

Here's the test code I'm trying to run:
```scala
def test_infxPatternClarifying(): Unit = {
    setUp()

    {
      val tool = InspectionProfileManager.getInstance(project)
                 .getCurrentProfile
                 .getInspectionTool("ScalaUnnecessaryParentheses", project)
                 .getTool

      tool match {
        case check: ScalaUnnecessaryParenthesesInspection => check setIgnoreClarifying false
        case _ =>
      }
    }

    val selected = s"val a +: $START(b *: c)$END = _ "
    checkTextHasError(selected)

    val text = s"val a +: ($CARET_MARKER b *: c) = _ "
    val result = "val a +: b *: c = _ "
    val hint = hintBeginning + " (b *: c)"
    testQuickFix(text, result, hint)

  }
```

niktrop:
Ah, it's actually registered inside methods like `checkTextHasError`, not in `setUp`. Probably we have no tests modifying inspection settings. You may override `setUp` method like this:
```
  override def setUp(): Unit = {
    super.setUp()
    getFixture.enableInspections(classOfInspection)
  }
```

oowekyala:
No error anymore, but the ScalaUnnecessaryParenthesesInspection instance that does the actual highlighting is different from the one that's returned by `getInspectionTool`, and it's created with the default parameters... I tried to enable it explicitly with `getFixture.enableInspections(tool)` but it's the same thing. Maybe it's too much trouble for what it's worth

niktrop:
You may leave your test commented, I'll have a look.

oowekyala:
Ok I pushed it

niktrop:
I've merged your PR manually with some changes:
https://github.com/JetBrains/intellij-scala/commits/4ffe6c02021a3ac8f16e880aaedc106cec9c1e1b

oowekyala:
Thanks, great work on the refactoring btw :)

PR #436, opened 2018-03-22
SCL-9720, SCL-13537: fix lexing issues for scaladoc

oowekyala:
[9720](https://youtrack.jetbrains.com/issue/SCL-9720), [13537](https://youtrack.jetbrains.com/issue/SCL-13537)

With that fixed, the highlighting of the `@define` tag param and tag names work properly


niktrop:
Your changes break some tests in `org.jetbrains.plugins.scala.lang.parser.ParserTest` and `org.jetbrains.plugins.scala.lang.scaladoclexer.ScaladocLexerTest`. Probably you just need to update testdata.

Also please add tests for new token types.

oowekyala:
The last commit fixes the tests of ParserTest. The tests failures in ScaladocLexerTest are weird, it looks like the `[[...]]` link lexing fails, even though I didn't make change anything to that part of the lexer. The thing is that `[[^some.code.link^ badText]]` is now parsed as
```
identifier {^some}
. {.}
identifier {code}
. {.}
identifier {Link}
```
instead of 
```
identifier {^}
DOC_COMMENT_BAD_CHARACTER {s}
DOC_COMMENT_BAD_CHARACTER {o}
DOC_COMMENT_BAD_CHARACTER {m}
DOC_COMMENT_BAD_CHARACTER {e}
DOC_COMMENT_BAD_CHARACTER {.}
DOC_COMMENT_BAD_CHARACTER {c}
DOC_COMMENT_BAD_CHARACTER {o}
DOC_COMMENT_BAD_CHARACTER {d}
DOC_COMMENT_BAD_CHARACTER {e}
DOC_COMMENT_BAD_CHARACTER {.}
DOC_COMMENT_BAD_CHARACTER {L}
DOC_COMMENT_BAD_CHARACTER {i}
DOC_COMMENT_BAD_CHARACTER {n}
DOC_COMMENT_BAD_CHARACTER {k}
```   
That new parsing is inconsistent with the flex file as far as I've seen ([here's the relevant part](https://github.com/JetBrains/intellij-scala/blob/idea181.x/scala/scala-impl/src/org/jetbrains/plugins/scala/lang/scaladoc/lexer/scaladoc.flex#L182-L205)). It may be an artifact of having regenerated the lexer, since there were some modifications to do by hand after generation anyway (it didn't compile).

I'm not sure changing the test is a good idea, and I don't understand how this doesn't work

I'll add some tests asap

niktrop:
Your PR was finally merged into a branch for 2019.1
https://github.com/jetbrains/intellij-scala/commit/3192d3946be55bdebcc1a0108d87d8e38094be7e
Thank you for contribution!

PR #437, opened 2018-03-22
#SCL-13305: Move Refactoring: support moving members between objects

Uria1:
This PR replaces PR https://github.com/JetBrains/intellij-scala/pull/429:

Fixed all code review comments except the comment regarding ChangeContextUtil by @niktrop


niktrop:
Can you please add some more advanced tests? I think for this refactoring it's necessary to check that
1) all usages of the moved member are updated (in expressions and imports)
2) all references inside the moved member are updated (should be handled by `ScalaChangeContextUtil`)

drhumlen:
Wouldn't it be much easier to reason about the unit tests if the Scala code was placed _inside_ the test-file instead of inside the testdata/-folder as individual files?

Like this:
```scala
testRefactoring(
 codeBefore = """
   object A {
     val foo = "bar"
   }

   object B
  """,

  expectedCodeAfter = """
    object A

    object B {
      val foo = "bar"
    }
  """
)
```

That makes the threshold for adding another test-case much lower too imo.

This isn't specific to this PR; more of a general observation.

niktrop:
@drhumlen This is possible, and actually we use this approach for most of new tests. I refactored tests for Move Member in this way: https://github.com/JetBrains/intellij-scala/commit/a582d766e087924572c20a1226e12b4deb6c84f7

niktrop:
I've just merged this PR into master with some additional improvements:
https://github.com/JetBrains/intellij-scala/tree/28b65f885b6e5f532e8e7d488e65ed44eb8c7550

It will be available in 2018.2. Thank you for contribution!

Uria1:
Thanks for the update!

PR #439, opened 2018-04-05
SCL-13572: Extract method refactoring fails to collect parameters for…

jauniusb:
… local scope references in inner classes
https://youtrack.jetbrains.com/issue/SCL-13572


niktrop:
It was merged manually quite a while ago, I just forgot to close it.
f54aa6937078666dc5ccca21631a895f0731ec0f

PR #441, opened 2018-04-15
Disable verbose boxing/unboxing desugaring by default

ShaneDelmore:
When teaching new users Scala I have found the "desugar scala code" action very useful with the exception of the fact that they find the unboxing/boxing code confusing.  This pr defaults boxing desugaring to disabled as I believe by the time users need to know about boxing and unboxing they will no longer find the dialogue confusing.


ShaneDelmore:
@jastice This disabled desugaring boxing/unboxing by default but does not persist the last used settings.

ShaneDelmore:
@jastice Is there anyone besides @pavelfatin who can review this?

pavelfatin:
That makes sense. Thanks!

PR #443, opened 2018-05-24
README Updated

chamodshehanka:
I've correct your some mistakes and put intellij idea badge.


jastice:
Thanks! What mistakes did you correct? I only see the badge

PR #448, opened 2018-10-06
Support circe JsonCodec/ConfiguredJsonCodec macro annotations

tkroman:
Based on @fommil's deriving plugin (same limitations apply: no implicit resolution etc => false positives happen)

Screenshots for with/without this PR.
Notice that `ConfiguredJsonCodec`-based derivation isn't red even though it should be (no implicit `Confifuration` in scope), but I would assume it's fine with most users.

screen shot 2018-10-06 at 23 48 18
screen shot 2018-10-07 at 00 00 18

### DIFF
--- a/.idea/intellij-scala.iml
+++ b/.idea/intellij-scala.iml
@@ -1,9 +1,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<module type="JAVA_MODULE" version="4">
-  <component name="NewModuleRootManager" inherit-compiler-output="true">
-    <exclude-output />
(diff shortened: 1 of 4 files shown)


jastice:
Thank you! We will have a look at this soon.

mutcianm:
Should this also work when annotated definition already has a companion object defined in the source?

tkroman:
@mutcianm i think this is covered by the call to `fakeCompanionClassOrCompanionClass`. Do you want a test for this?

mutcianm:
Thanks for your contribution! Look forward to the next nightly build

PR #450, opened 2018-11-06
Externalized Hydra jps support into its own module

dotta:
* Introduced two services classes `DataFactoryService` and
  `CompilerFactoryService` which allow to:
  1. Configure the used Scala compiler (e.g., setting extra compiler options),
and
  2. Setting up Zinc to use a different compiler-bridge.
* Thanks to the above, almost all Hydra specific code was moved out of the
  scala/compiler-jps module. The only remaining bit to figure out is how to add
  the hydra-compiler-jps.jar to the JPS classpath.
* The `hydra.data.CompilerDataFactory` currently duplicates the implementation
  of `data.CompilerData` because on the one hand, I wasn't convinced you wanted
  to offer an API for it, and on the other hand we will need to inject a
  diffent AnalysisCallback as the current one is not thread-safe. That being said,
  I'm certainly open to suggestions that reduce code duplication.

This commit continued the work started in 27f946d to externalize the Hydra
support into a separate plugin.


scala/compiler-jps/src/org/jetbrains/jps/incremental/scala/local/worksheet/ILoopWrapperFactoryHandler.scala
```diff
@@ -108,7 +108,7 @@ object ILoopWrapperFactoryHandler {
 
   private val JAVA_USER_CP_KEY = "java.class.path"
   private val STOP_WORDS = Set("scala-library.jar", "scala-nailgun-runner.jar", "nailgun.jar", "compiler-shared.jar",
-    "incremental-compiler.jar", "compiler-jps.jar")
+    "incremental-compiler.jar", "compiler-jps.jar", "hydra-compiler-jps.jar")
```
dotta, line 111: In theory, it would be best if we use the vanilla Scala compiler in the worksheet (it's only one source file). However, I'm not sure that's actually possible, is compilation of the worksheet file delegate to Zinc (and hence it will use the set compiler-bridge) or do you handle compilation of the worksheet file outside of Zinc?

scala/scala-impl/resources/META-INF/scala-plugin-common.xml
```diff
@@ -72,7 +72,7 @@
         <testFinder implementation="org.jetbrains.plugins.scala.testingSupport.ScalaTestFinder"/>
         <exceptionFilter implementation="org.jetbrains.plugins.scala.testingSupport.util.scalatest.ScalaTestFailureLocationFilterFactory"/>
         <exceptionFilter implementation="org.jetbrains.plugins.scala.stacktrace.ScalaPackageObjectFilterFactory" order="first"/>
-        <compileServer.plugin classpath="scala-library.jar;compiler-shared.jar;jps/nailgun.jar;jps/compiler-jps.jar;jps/sbt-interface.jar;jps/incremental-compiler.jar"/>
+        <compileServer.plugin classpath="scala-library.jar;compiler-shared.jar;jps/nailgun.jar;jps/compiler-jps.jar;jps/hydra-compiler-jps.jar;jps/sbt-interface.jar;jps/incremental-compiler.jar"/>
```
dotta, line 75: This is one of the last two pending task I believe we absolutely need to take care to externalize the Hydra support into its own plugin. How can we add the `hydra-compiler-jps` to the compilerServer classpath from *outside* the scala plugin?

Followup question: once we move the Hydra support into its own plugin, how can we obtain the `compiler-jps.jar` which is needed to build the `hydra-compiler-jps` module?

scala/scala-impl/src/org/jetbrains/plugins/scala/compiler/CompileServerLauncher.scala
```diff
@@ -200,7 +200,8 @@ object CompileServerLauncher {
       new File(jpsRoot, "nailgun.jar"),
       new File(jpsRoot, "sbt-interface.jar"),
       new File(jpsRoot, "incremental-compiler.jar"),
-      new File(jpsRoot, "compiler-jps.jar")
+      new File(jpsRoot, "compiler-jps.jar"),
+      new File(jpsRoot, "hydra-compiler-jps.jar")
```
dotta, line 204: This is the second pending task. Should I create an extension point to add the `hydra-compiler-jps` here? Also, the JARs listed here seems to be the same to the one set in the `<compileServer.plugin>` extension, so I'm wondering if there is an opportunity to reduce duplication.

dotta:
Btw, the idea183.x branch this PR is based on seem to currently have a failing test:

```
> scalaCommunity/test:testOnly org.jetbrains.plugins.scala.lang.macros.CirceCodecTest
...
[error] Test org.jetbrains.plugins.scala.lang.macros.CirceCodecTest.testCaseClassWithCompanion failed: java.lang.NullPointerException: null, took 0.299 sec
```

I just thought to let you know.

scala/scala-impl/resources/META-INF/scala-plugin-common.xml
niktrop: You may register your own `compileServer.plugin` in `hydra-plugin-common.xml` and add `hydra-compiler-jps.jar` there.

You will probably use `sbt-idea-plugin` for plugin setup. It will download scala plugin to some directory, so you will need to find and add `compiler-jps.jar` to dependencies of `hydra-compiler-jps` project in your build file. 
https://github.com/JetBrains/sbt-idea-plugin#ideaexternalplugins--settingkeyideaplugin

@mutcianm can help you with it.

scala/scala-impl/src/org/jetbrains/plugins/scala/compiler/CompileServerLauncher.scala
niktrop: It appears we already have an extension you need:
org.jetbrains.plugins.scala.compiler.NailgunServerAdditionalCp

scala/compiler-jps/src/org/jetbrains/jps/incremental/scala/local/worksheet/ILoopWrapperFactoryHandler.scala
niktrop: I suggest to leave it as is right now. Worksheet uses zinc, but I don't think it will be a problem.

dotta: Ok. But how should the `hydra-compiler-jps.jar` be added to it? Or are you suggesting we leave the code above as it is now?

scala/scala-impl/src/org/jetbrains/plugins/scala/compiler/CompileServerLauncher.scala
dotta: But this extension point is not used here, is it? Happy to do the change so that it's used, if that's what you are hinting at.

scala/scala-impl/resources/META-INF/scala-plugin-common.xml
dotta: Sounds good. I guess we will do this when moving the whole hydra support into its own plugin, which is the next thing I'm working on. But we can merge this PR and then I'll have a new PR that removes all Hydra code from the intellij-scala codebase. Ok?

scala/scala-impl/src/org/jetbrains/plugins/scala/compiler/CompileServerLauncher.scala
dotta: Oh, I'm wrong, it's used. Please, disregard my last comment.

scala/compiler-jps/src/org/jetbrains/jps/incremental/scala/local/worksheet/ILoopWrapperFactoryHandler.scala
niktrop: Yes, you may leave it here. I think this part should be refactored, but I'm not sure yet how exactly to do it.

scala/scala-impl/resources/META-INF/scala-plugin-common.xml
niktrop: @dotta This PR will be merged by our CI server after passing tests.

dotta: Oh, ok, awesome!

PR #451, opened 2018-11-08
Moved Hydra support outside of the intellij-scala repo

dotta:
The only pending aspect is figuring out how to remove the reference to the
hydra-compiler-jps.jar from the `ILoopWrapperFactoryHandler` used by the worksheet.

Here is the related PR for the Hydra idea plugin: https://github.com/triplequote/hydra-idea-plugin/pull/1. I'd be very thankful if you could take a look and confirm it's good to you. Thanks!


niktrop:
Merged manually.

PR #455, opened 2019-02-06
ScPackageImpl#findPackageObject should not cache null values #SCL-14266

ddworak:
We commonly use a wildcard import from the io.udash package object, which inherits type aliases for most commonly used classes. These very often fail to be recognized correctly, rendering `import io.udash._` unused, usually after reindexing/recompiling/updating sbt project etc. The workaround we used was to remove a definition of a method/class or an import, it worked perfectly for a while after that.

I reported this bug in https://youtrack.jetbrains.com/issue/SCL-14266 and this change finally seems to fix it.

#SCL-14266 fixed


sugakandrey:
@ddworak Thank you! I'll try to take a look at this today.

sugakandrey:
Fixed in f5219aace053b8957111c6fdb099ba422bf341f1 (thank you @ddworak for tracking the issue down)

PR #456, opened 2019-03-12
Allow compilerPlugin with GAV plus classifier

seanf:
#[SCL-15099](https://youtrack.jetbrains.com/issue/SCL-15099) fixed

### DIFF
--- a/scala/integration/maven/src/org/jetbrains/plugins/scala/project/maven/ScalaMavenImporter.scala
+++ b/scala/integration/maven/src/org/jetbrains/plugins/scala/project/maven/ScalaMavenImporter.scala
@@ -12,7 +12,7 @@ import com.intellij.openapi.roots.impl.libraries.LibraryEx
 import com.intellij.openapi.roots.impl.libraries.LibraryEx.ModifiableModelEx
 import org.jdom.Element
 import org.jetbrains.idea.maven.importing.{MavenImporter, MavenRootModelAdapter}
-import org.jetbrains.idea.maven.model.{MavenArtifact, MavenArtifactInfo, MavenId => IdeaMavenId, MavenPlugin}
+import org.jetbrains.idea.maven.model.{MavenArtifact, MavenArtifactInfo, MavenPlugin}
 import org.jetbrains.idea.maven.project._
 import org.jetbrains.idea.maven.server.{MavenEmbedderWrapper, NativeMavenProjectHolder}
 import org.jetbrains.plugins.scala.extensions._
@@ -94,8 +94,8 @@ class ScalaMavenImporter extends MavenImporter("org.scala-tools", "maven-scala-p
       val repositories = mavenProject.getRemoteRepositories
 
       def resolve(id: MavenId): MavenArtifact = {
-        embedder.resolve(new MavenArtifactInfo(id.asIdea, "pom", null), repositories)
-        embedder.resolve(new MavenArtifactInfo(id.asIdea, "jar", id.classifier.orNull), repositories)
+        embedder.resolve(new MavenArtifactInfo(id.groupId, id.artifactId, id.version, "pom", null), repositories)
+        embedder.resolve(new MavenArtifactInfo(id.groupId, id.artifactId, id.version, "jar", id.classifier.orNull), repositories)
       }
 
       // Scala Maven plugin can add scala-library to compilation classpath, without listing it as a project dependency.
@@ -194,7 +194,5 @@ private object ScalaMavenImporter {
     * Represents a Maven artifact by group, artifact, version and optional classifier.
     * Similar to [[org.jetbrains.idea.maven.model.MavenId]], but supports classifier.
     */
-  private case class MavenId(groupId: String, artifactId: String, version: String, classifier: Option[String] = None) {
-    def asIdea = new IdeaMavenId(groupId, artifactId, version)
-  }
+  private case class MavenId(groupId: String, artifactId: String, version: String, classifier: Option[String] = None)
 }
--- a/scala/integration/maven/src/org/jetbrains/plugins/scala/project/maven/ScalaMavenImporter.scala
+++ b/scala/integration/maven/src/org/jetbrains/plugins/scala/project/maven/ScalaMavenImporter.scala
@@ -12,7 +12,7 @@ import com.intellij.openapi.roots.impl.libraries.LibraryEx
 import com.intellij.openapi.roots.impl.libraries.LibraryEx.ModifiableModelEx
 import org.jdom.Element
 import org.jetbrains.idea.maven.importing.{MavenImporter, MavenRootModelAdapter}
-import org.jetbrains.idea.maven.model.{MavenArtifact, MavenArtifactInfo, MavenId, MavenPlugin}
+import org.jetbrains.idea.maven.model.{MavenArtifact, MavenArtifactInfo, MavenId => IdeaMavenId, MavenPlugin}
 import org.jetbrains.idea.maven.project._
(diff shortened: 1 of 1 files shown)


scala/integration/maven/src/org/jetbrains/plugins/scala/project/maven/ScalaMavenImporter.scala
```diff
@@ -159,12 +163,17 @@ private object ScalaMavenImporter {
...
         plugin.getChildTextTrim("groupId").toOption
           .zip(plugin.getChildTextTrim("artifactId").toOption)
-          .zip(plugin.getChildTextTrim("version").toOption).map {
-          case ((groupId, artifactId), version) => new MavenId(groupId, artifactId, version)
+          .zip(plugin.getChildTextTrim("version").toOption)
+        .map {
+          case ((groupId, artifactId), version) =>
+            // It's okay if classifier is absent or blank
+            val classifier = plugin.getChildTextTrim("classifier").toOption
+              .flatMap(c => if (c.isEmpty) None else Some(c))
```
pavelfatin: `.filterNot(_.isEmpty)` :)

pavelfatin:
Thanks for the PR! Looks good, though it's not clear whether that `MavenGAVC` is really needed everywhere, given that, currently, the classifier is only used in a single place, while "MavenGAVC" is a non-standard, possibly confusing abbreviation.

How about:

```scala
private type Classifier = String
def plugins: Seq[MavenId, Classifier] =
def resolve(id: MavenId, classifier: Option[Classifier] = None): MavenArtifact = 
```
(but that's up to you, the code is OK even as it is)

scala/integration/maven/src/org/jetbrains/plugins/scala/project/maven/ScalaMavenImporter.scala
```diff
@@ -180,4 +189,12 @@ private object ScalaMavenImporter {
 
     def valid: Boolean = compilerPlugin.isDefined && compilerVersion.isDefined
   }
+
+  /**
+    * Represents a Maven artifact by group, artifact, version and optional classifier.
+    * Similar to [[org.jetbrains.idea.maven.model.MavenId]], but supports classifier.
```
pavelfatin, line 195: If you define the class as `MavenGAVC(id: MavenId, classifier: Option[String] = None)`, then this fact would be apparent, and the comment would be redundant, as well as the additional `mavenId` method.

scala/integration/maven/src/org/jetbrains/plugins/scala/project/maven/ScalaMavenImporter.scala
seanf: Of course! Thanks.

scala/integration/maven/src/org/jetbrains/plugins/scala/project/maven/ScalaMavenImporter.scala
seanf: Thanks for the review @pavelfatin. 

I've added another commit which defines MavenGAVC in terms of MavenId, but I'm not sure it really improves things overall. I found that the function `localPathTo` which uses `MavenId`/`MavenGAVC` got pretty ugly (eg `id.mavenId.getArtifactId`) until I defined an `apply` method and direct accessors for `groupId`, `artifactId` and `version`.

My idea of `MavenGAVC` was really a complete replacement for `MavenId`, which surprised me by not supporting classifier. Honestly I don't think there is much to be gained by reusing a Java class (`MavenId`) which is a one-line case class in Scala. It doesn't really stand alone as a concept once you need classifier, which is why I ended up using MavenGAVC all over the file.

The branch is open for editing if I've missed your point entirely and you want to try something else. Or if you want to drop that last commit before merging. ;-)

pavelfatin: Hmm... with the accessors and the extractor it's indeed not much better :)

The main proposal was to explicitly use a classifier string, instead of the `MavenGVAC`, only where needed, because:
* The classifier is needed only in some places, yet the "Id" -> "GVAC" has to be update everywhere.
* The "GAVC" thing is not obvious.
* "GAVC" is still called "id" everywhere: `def resolve(id: MavenGAVC)`.
* The `MavenId` is accessed as `.mavenId`, even though the "GAVC" is already "Maven".

What if we use the original variant, but:
1. Import the IDEA's `MavenId` as `IdeaMavenId`.
2. Rename the `MavenGVAC` to `MavenId`.
3. Define `asIdea` method in our `MavenId` (by analogy with the `asScala`).

This will keep the familiar name and will minimize the unrelated changes.

seanf: Hi @pavelfatin I've implemented your suggestions, but I think perhaps it would be better to omit `asIdea` (it throws away the classifier, so it's a bit of a leaky abstraction).

Alternatively, I suppose `asIdea` could return a tuple: `(MavenId, Option[Classifier])` to avoid the leak, but that tuple isn't really an IDEA concept, so I don't really favour this choice.

The `resolve` method which now calls `asIdea` could simply construct the `MavenArtifactInfo` using the group/artifact/version (and classifier) directly from the case class, in which case `asIdea` isn't needed.


I would normally try to avoid calling two different things by the same name (`MavenId`), but they are in different packages and this one's private, so it's just a matter of style. I don't suppose you like the name `MavenCoord` for the case class formerly known as `MavenGAVC`?

pavelfatin: > The resolve method which now calls `asIdea` could simply construct the `MavenArtifactInfo` using the group/artifact/version (and classifier) directly from the case class, in which case `asIdea` isn't needed.

True, that would be even better (because `asIdea` is indeed a leaky abstraction).

> I would normally try to avoid calling two different things by the same name (MavenId), but they are in different packages and this one's private, so it's just a matter of style. I don't suppose you like the name MavenCoord for the case class formerly known as MavenGAVC?

I also don't typically advocate reusing a name for two different things, but comparing to the "GAVC" (or "Coord", for that matter) abbreviation, reusing the name seems like a lesser evil :) BTW, it's not that uncommon - the Scala standard library re-uses the (publicly accessible) names of generic / mutable / immutable collections (though, that's why we have to add the `mutable.` prefix).

Feel free to add the `MavenArtifactInfo` creation, and we'll merge the changes.

seanf:
@pavelfatin Okay, I've removed `asIdea` as we discussed. I'd recommend a squash merge!

seanf:
@pavelfatin Is there anything else you need me to do on this one?

pavelfatin:
@seanf Thanks for the PR and the useful suggestions!

pavelfatin:
BTW, do you have a Twitter account? - we may mention your help in https://twitter.com/IntelliJScala (if you don't mind).

seanf:
Hi Pavel,

This is me:
- s_flanigan
- https://mobile.twitter.com/s_flanigan

Thanks!

seanf:
@pavelfatin Did my PR target the wrong branch or something? It looks like these changes still haven't made it into a release.

UPDATE: These changes were included in [version 2019.2.6](https://plugins.jetbrains.com/plugin/1347-scala/update/63997) (released on 14 June for IDEA 2019.2), according to GitHub's tag info for https://github.com/JetBrains/intellij-scala/commit/6fe52f63f771057f35bd84309f2075542bfcf69b.

PR #457, opened 2019-03-19
Root project now aggregates all projects

romanowski:


mutcianm:
Unfortunately, we have intentionally disabled root aggregation because it messes up many of our custom tasks and disabling aggregation just for them is as even more verbose than defining aggregation for all necessary projects.
In our case it's more concise to use classpath dependencies to achieve task delegation

PR #460, opened 2019-04-01
support for case class + case object in file create dialog

domdorn:


domdorn:
any reason why this is not merged into master as well? 
I just updated to the newest version (v2019.1.8.21) of the plugin in IDEA and this change is still not there?

SrTobi:
This was merged and I have it in the nightly... not sure at the moment if it is in the normal stable version

SrTobi:
image

PR #462, opened 2019-04-09
Fix caching compiler options. Now it is per target not per module name

lukaszwawrzyk:
Fix caching compiler options. Now it is per target not per module name (module name may have multiple targets)


niktrop:
Thank you for the contribution! It was manually merged to the 2019.2 branch: https://github.com/JetBrains/intellij-scala/commit/bd78cabe5aa8d24b95e001f9e4631a82ca03fb7d

PR #464, opened 2019-04-24
Suggest to add `override` modifier 

aradchykov:
Quick fix which allows to add `override` modifier 

```
trait A { def sample(): Int }
class B extends A { def sample(): Int }
```

into

```
trait A { def sample(): Int }
class B extends A { **override** def sample(): Int }
```

### DIFF
--- a/scala/scala-impl/resources/META-INF/scala-plugin-common.xml
+++ b/scala/scala-impl/resources/META-INF/scala-plugin-common.xml
@@ -1018,6 +1018,14 @@
               language="Scala" level="WARNING"
               enabledByDefault="true"/>
       <localInspection
+            implementationClass="org.jetbrains.plugins.scala.codeInspection.methodSignature.ScalaOverrideModifierMissingInspection"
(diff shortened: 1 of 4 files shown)


sugakandrey:
@comchi Thanks for taking your time to contribute this! There's already a very similar functionality implemented in [OverridingAnnotator](https://github.com/JetBrains/intellij-scala/blob/faa112914a940e8355dbca571af400c45ff4b2c0/scala/scala-impl/src/org/jetbrains/plugins/scala/annotator/OverridingAnnotator.scala#L143), which is also a bit more general (your implementation will fail to add `val/var` when overriding member is a case class parameter and does not not support type members). It is currently only triggered for non-abstract overriding members, but I think we can generalise it to also include this quick fix along with an info message, when overriding abstract `super`.

aradchykov:
cool, thanks @sugakandrey. I will redo this via OverridingAnnotator in ~1-2 weeks and open a new PR.

niktrop:
This inspection is implemented for 2019.2, see 
https://youtrack.jetbrains.com/issue/SCL-15502

PR #465, opened 2019-04-24
Propose to pull up members if they override nothing

aradchykov:
```
trait A
class A { override def sample(): int = 1 }
```

after right mouse click on `sample` to --->

```
trait A { def sample(): Int }
class A { override def sample(): int = 1 }
```


niktrop:
@comchi Thank you for your contribution! Your commit was merged to the branch for IDEA 2019.2: https://github.com/JetBrains/intellij-scala/commit/9b212b453e50fa351926067c4b85d7e338964635

PR #466, opened 2019-05-07
use current document version when formatting with scalafmt

ingarabr:
The reformat on save reads the content from the source file
when it should have used the current document that's in memory.

SCL-15402 fixed

### DIFF
--- a/scala/scala-impl/src/org/jetbrains/plugins/scala/lang/formatting/processors/ScalaFmtPreFormatProcessor.scala
+++ b/scala/scala-impl/src/org/jetbrains/plugins/scala/lang/formatting/processors/ScalaFmtPreFormatProcessor.scala
@@ -307,17 +307,16 @@ object ScalaFmtPreFormatProcessor {
     if (config == null || respectProjectMatcher && !configManager.isIncludedInProject(file, config))
       return
 
-    formatWithoutCommit(file, Some(document), config) match {
+    formatWithoutCommit(document, config) match {
       case Left(error: ScalafmtFormatError) =>
         reportInvalidCodeFailure(file, Some(error))(file.getProject)
       case _ =>
     }
   }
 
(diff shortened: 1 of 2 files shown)


ingarabr:
I have signed the CLA

unkarjedy:
@ingarabr Thanks for the PR! I will have a look within the next 24 hours.

scala/scala-impl/src/org/jetbrains/plugins/scala/lang/formatting/processors/ScalaFmtPreFormatProcessor.scala
```diff
@@ -300,24 +301,24 @@ object ScalaFmtPreFormatProcessor {
...
     }
   }
 
-  private def formatWithoutCommit(file: PsiFile, config: ScalafmtDynamicConfig): Either[FormattingError, Unit] = {
+  private def formatWithoutCommit(file: PsiFile, optDocument: Option[Document], config: ScalafmtDynamicConfig): Either[FormattingError, Unit] = {
     val scalaFmt: ScalafmtReflect = config.fmtReflect
     for {
-      document <- Option(PsiDocumentManager.getInstance(file.getProject).getDocument(file)).toRight(DocumentNotFoundError)
-      formattedText <- scalaFmt.tryFormat(file.getText, config)
+      document <- optDocument.orElse(Option(PsiDocumentManager.getInstance(file.getProject).getDocument(file))).toRight(DocumentNotFoundError)
```
unkarjedy: @ingarabr Looks like in all (2) places where `private def formatWithoutCommit` is used `Document` instance is already available. So I think we can remove `file: PsiFile` parameter, make `document: Document` parameter instead of `optDocument: Option[Document]` and get dir of `getOrElse/toRight` stuff.
Also looks like `DocumentNotFoundError` will not be needed anymore.

PR #468, opened 2019-07-02
Re-publish functions needed by Hydra in compiler-jps

CMoH:


CMoH:
The hydra plugin relies on those two functions (`javaHome` and `compilerJarsIn`) to inject itself into the compilation to do its work. Can you re-publish these from the compiler-jps plugin?

I think we can get away with migrating `compilerVersion` into our plugin (needed to select the appropriate hydra compiler version), but I re-added it here since it originated here. Please advise if it is acceptable for you.

mutcianm:
@adkozlov please see 78b533e1973c1dc19d3bca7a37525c00a4738709

scala/compiler-jps/src/org/jetbrains/jps/incremental/scala/data/CompilerData.scala
```diff
@@ -155,5 +155,11 @@ object CompilerData extends CompilerDataFactory {
     versions.exists(version.startsWith)
   }
 
+  def compilerVersion(module: JpsModule): Option[String] = compilerJarsIn(module) match {
+    case Some(CompilerJars(_, compiler, _)) => version(compiler)
+    case None => None
+  }
+  def version(compiler: File): Option[String] = readProperty(compiler, "compiler.properties", "version.number")
```
adkozlov: `readProperty(compiler, "compiler.properties", "version.number")` is being used in the `versionIn` method. Now you can call `version` instead.

The rest looks fine to me.

CMoH: I have squashed the change, and given that it's published API I also renamed `version(compiler: File)` to `compilerVersion`. Did I understand your request correctly?

CMoH:
If you want to minimize the number of public functions, I can also drop `compilerVersion(module)` fom intellij-scala and only leave `compilerVersion(file)` here, and implement the indirection through `compilerJarsIn` only in our Hydra plugin.

scala/compiler-jps/src/org/jetbrains/jps/incremental/scala/data/CompilerData.scala
adkozlov: Yes. Cool, thanks.

adkozlov:
@CMoH It would be really appreciated, thanks!

CMoH:
> @CMoH It would be really appreciated, thanks!

Done!

PR #469, opened 2019-07-15
HOCON includes for Scalafmt

nadavwr:
Currently, any `include` added to .scalafmt.conf is not resolved. This limits our options for enforcing uniform style across our projects.

This PR changes how .scalafmt.conf is loaded. Instead of reading file contents then parsing it as text once, a typesafe-config `ConfigIncluder` is used to resolve includes over `VirtualFile`. The loaded `Config` is then rendered to text (`_.root.render()`).

Still missing is change detection—only changes to the main configuration file `.scalafmt.conf` trigger configuration reload.

Note that even though `include "my-defaults.conf"` resolves files relative to the including file when using `ConfigFactory.parseFile`, this behavior is currently broken for scalafmt (both cli and sbt plugin) due to similar reasons (single file read once, includes left unresolved). I'm pursuing PR's against scalafmt as well.

related scalafmt PR: https://github.com/scalameta/scalafmt/pull/1450


unkarjedy:
@nadavwr Thanks for the PR. I will have a look soon.

nadavwr:
FYI, The matching Scalafmt PR is here: https://github.com/scalameta/scalafmt/pull/1450

scala/scala-impl/src/org/jetbrains/plugins/scala/lang/formatting/scalafmt/ScalafmtDynamicConfigManager.scala
```diff
@@ -224,10 +224,35 @@ object ScalafmtDynamicConfigManager {
...
     instanceIn(file.getProject).isIncludedInProject(file)
 
-  def readVersion(configFile: VirtualFile): Either[Throwable, Option[String]] = {
+  private def parseHoconFile(project: Project, configFile: VirtualFile): Config = {
+    val fileDocumentManager = FileDocumentManager.getInstance
+    val emptyObj = ConfigValueFactory.fromMap(Map.empty[String, Any].asJava)
+    lazy val parseOpts: ConfigParseOptions = ConfigParseOptions.defaults
+      .setIncluder(new ConfigIncluder {
+        var fallbackOpt: Option[ConfigIncluder] = None
+        override def withFallback(fallback: ConfigIncluder): ConfigIncluder = {
```
unkarjedy, line 233: Can we rewrite it to conform `withIncluder` contact defined in comment?
```java
/**
   * Returns a new includer that falls back to the given includer. This is how
   * you can obtain the default includer; it will be provided as a fallback.
   * It's up to your includer to chain to it if you want to. You might want to
   * merge any files found by the fallback includer with any objects you load
   * yourself.
   *
   * It's important to handle the case where you already have the fallback
   * with a "return this", i.e. this method should not create a new object if
   * the fallback is the same one you already have. The same fallback may be
   * added repeatedly.
   *
   * @param fallback
   * @return a new includer
   */
  ConfigIncluder withFallback(ConfigIncluder fallback);
```

> Returns a new includer

(+doing this we could avoid mutable variable )

> It's important to handle the case where you already have the fallback with a "return this"

scala/scala-impl/src/org/jetbrains/plugins/scala/lang/formatting/scalafmt/ScalafmtDynamicConfigManager.scala
```diff
@@ -224,10 +224,35 @@ object ScalafmtDynamicConfigManager {
...
+          resolveScalafmtConfigFile(project, what)
+            .flatMap(file => fileDocumentManager.getDocument(file).toOption)
+            .map(_.getText)
+            .map(ConfigFactory.parseString(_, parseOpts))
+            .map(_.root)
+            .orElse(fallbackOpt.map(_.include(context, what)))
+            .getOrElse(emptyObj)
+      })
+    val document = inReadAction(fileDocumentManager.getDocument(configFile))
+    val s = ConfigFactory.parseString(document.getText, parseOpts)
```
unkarjedy, line 248: strange redundant `s`

scala/scala-impl/src/org/jetbrains/plugins/scala/lang/formatting/scalafmt/ScalafmtDynamicConfigManager.scala
```diff
@@ -224,10 +224,35 @@ object ScalafmtDynamicConfigManager {
...
+        override def withFallback(fallback: ConfigIncluder): ConfigIncluder = {
+          fallbackOpt = Some(fallback)
+          this
+        }
+        override def include(context: ConfigIncludeContext,
+                             what: String): ConfigObject =
+          resolveScalafmtConfigFile(project, what)
+            .flatMap(file => fileDocumentManager.getDocument(file).toOption)
+            .map(_.getText)
+            .map(ConfigFactory.parseString(_, parseOpts))
```
unkarjedy, line 242: We could avoid this hack with `lazy val` and cyclic dependencies by extracting options from context. 
Have you considered using these methods:
`ConfigIncludeContext.parseOptions`
`ConfigParseable.options()` ?

scala/scala-impl/src/org/jetbrains/plugins/scala/lang/formatting/scalafmt/ScalafmtDynamicConfigManager.scala
```diff
@@ -224,10 +224,35 @@ object ScalafmtDynamicConfigManager {
...
+    lazy val parseOpts: ConfigParseOptions = ConfigParseOptions.defaults
+      .setIncluder(new ConfigIncluder {
+        var fallbackOpt: Option[ConfigIncluder] = None
+        override def withFallback(fallback: ConfigIncluder): ConfigIncluder = {
+          fallbackOpt = Some(fallback)
+          this
+        }
+        override def include(context: ConfigIncludeContext,
+                             what: String): ConfigObject =
+          resolveScalafmtConfigFile(project, what)
```
unkarjedy, line 239: Quite hard to follow what is going on in this big call chain.
We could split it extracting some peaces to meaningful variables like:
```scala
val configText = ...
val configRoot = ...
configRoot.getOrElse(emptyObj)
```

scala/scala-impl/src/org/jetbrains/plugins/scala/lang/formatting/scalafmt/ScalafmtDynamicConfigManager.scala
```diff
@@ -224,10 +224,35 @@ object ScalafmtDynamicConfigManager {
   def isIncludedInProject(file: PsiFile): Boolean =
     instanceIn(file.getProject).isIncludedInProject(file)
 
-  def readVersion(configFile: VirtualFile): Either[Throwable, Option[String]] = {
+  private def parseHoconFile(project: Project, configFile: VirtualFile): Config = {
+    val fileDocumentManager = FileDocumentManager.getInstance
+    val emptyObj = ConfigValueFactory.fromMap(Map.empty[String, Any].asJava)
```
unkarjedy, line 229: Is `emptyObj` used only when some file is included? Could we move it to a method of a ConfigIncluder?

unkarjedy:
To sum up current minor issues with these changes:

1) When scalafmt config file is changed we detect it on the next format and show a notification:
`Scalafmt picked up new style configuration`
If we modify some of the included files their changes will not be detected.

2) We do not handle (erroneous) cyclic dependencies, when two files include each other, though It is a very unlikely situation.  Currently `StackOverflowError` is thrown

unkarjedy:
added a tracking YouTrack ticket: https://youtrack.jetbrains.com/issue/SCL-17726
changes with some fixes were manually applied due to too many conflicts

PR #470, opened 2019-07-15
Add trace logging for bsp server client communication

samvel1024:


bsp/src/org/jetbrains/bsp/protocol/session/BspSession.scala
```diff
@@ -88,6 +90,25 @@ class BspSession private(bspIn: InputStream,
     }
   }
 
+
+  private def bspTraceLogger: PrintWriter = {
+    val logdir = new File(PathManager.getLogPath)
+    logdir.mkdirs()
+    val logfile = new File(logdir, "bsp-protocol-trace.log")
```
jastice: I suggest a `bsp` subdirectory under the logdir, and a trace file that is unique to the session, maybe including a timestamp

PR #476, opened 2019-08-20
Bsp testing support

samvel1024:
Testing support for BSP.
Couple of words about the changes
- Include scala test classes (buildTarget/scalaTestClasses) in `BspMetadata`
- Create a run configuration for running BSP tests
- Use SMTestRunner console view to display the test tree (in `BspTestRunner`)
- Two modes for running tests. All tests in project and selected test classes (by regex)

Screen Shot 2019-08-21 at 01 10 36

Maybe we can replace the icon for bsp test with something better 😆


bsp/src/org/jetbrains/bsp/BSPCli.scala
```diff
@@ -147,33 +142,67 @@ object BSPCli extends App {
...
 
-  def testRequest(targets: BuildIds)(server: BspServer): CompletableFuture[TestResult] = {
+  def testAllRequest(targets: BuildIds)(server: BspServer): CompletableFuture[TestResult] = {
     val params = new bsp4j.TestParams(targets)
     params.setOriginId(UUID.randomUUID().toString)
     server.buildTargetTest(params)
   }
 
+
+  def testSingleRequest(targets: BuildIds, clasId: String, buildTargetUri: String)(server: BspServer): CompletableFuture[TestResult] = {
```
jastice, line 168: typo: clasId

bsp/src/org/jetbrains/bsp/project/resolver/BspProjectResolver.scala
```diff
@@ -79,8 +79,13 @@ class BspProjectResolver extends ExternalSystemProjectResolver[BspExecutionSetti
               List.empty[ScalacOptionsItem]
             }
 
+            val testClasses = data.scalaTestClasses.map(_.getItems.asScala).getOrElse {
```
jastice: I think we should find a different way to resolve test classes, since I expect doing it during import might make the import take much longer, due to potentially triggering recompilation.

Possibility: on-demand in the run config or when trying to trigger a test run for a specific class from context menu. Any other ideas?

jastice (commented):
let's consider if we can change the test resolution to have less impact

bsp/src/org/jetbrains/bsp/project/resolver/BspProjectResolver.scala
samvel1024: It makes sense to load the classes if we want to select a group of them with a regex. However should we store them in BspMetadata then (or any other way of persistence) ? If yes, then how can that be done. The existing code will not allow to do that easily, since it computes the data all together during the import.

jastice: We can store them separately in a `PersistentStateComponent`. See eg `org.jetbrains.sbt.project.module.SbtModule`

bsp/src/org/jetbrains/bsp/project/test/bspTestRunConfig.scala
```diff
@@ -0,0 +1,79 @@
...
+class TestClass(@BeanProperty var target: String, @BeanProperty var classes: j.List[String]) {
+  def this() = this("", Collections.emptyList())
+}
+
+class BspTestRunConfiguration(val project: Project, val configurationFactory: ConfigurationFactory, val name: String)
+  extends RunConfigurationBase[String](project, configurationFactory, name) {
+
+
+  @BeanProperty var testMode: TestMode = TestMode.ALL_IN_PROJECT
+  @BeanProperty var lastTestClassesResponse: j.List[TestClass] = Collections.emptyList()
```
samvel1024, line 54: I decided to cache the last response by BSP server and store it in the run configuration.

PR #477, opened 2019-08-28
SCL-16179 Support for switch expressions (no yield yet)

amaembo:
Just some contribution in the known-to-me area to warm up.

A yield statement could be supported to some extent if it's the last statement in the top-level block inside the `case`. In other cases, the conversion seems non-trivial (though I'm a complete newbie in Scala, probably I just don't know something). Also to test the yield it will be necessary to set language level to 13-preview. Otherwise, the `yield` keyword is tokenized as an identifier. I can do these if there are no objections.

### DIFF
--- a/scala/conversion/src/org/jetbrains/plugins/scala/conversion/visitors/SimplePrintVisitor.scala
+++ b/scala/conversion/src/org/jetbrains/plugins/scala/conversion/visitors/SimplePrintVisitor.scala
@@ -101,8 +101,8 @@ class SimplePrintVisitor protected() {
       visitWhile(w, initialization, condition, body, update, whileType)
     case TryCatchStatement(resourcesList, tryBlock, catchStatements, finallyStatements, arrow) =>
       visitTryCatch(resourcesList, tryBlock, catchStatements, finallyStatements, arrow)
-    case SwitchStatemtnt(expression, body) => visitSwitchStatement(expression, body)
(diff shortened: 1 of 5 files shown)


scala/conversion/src/org/jetbrains/plugins/scala/conversion/visitors/SimplePrintVisitor.scala
```diff
@@ -790,10 +790,17 @@ class SimplePrintVisitor protected() {
     }
   }
 
-  protected def visitSwitchLabelStatement(caseValue: Option[IntermediateNode], arrow: String): Unit = {
+  protected def visitSwitchLabelStatement(caseValues: Seq[IntermediateNode], arrow: String,
+                                          body: Option[IntermediateNode]): Unit = {
     printer.append("case ")
-    if (caseValue.isDefined) visit(caseValue.get)
+    var first = true
+    caseValues.foreach(e => {
```
niktrop, line 797: There is a `printWithSeparator` utility method.

niktrop: I'm going to cherry-pick it manually, so I'll fix it.

amaembo: Thank you

niktrop:
Merged manually.

PR #478, opened 2019-08-28
Cleanup ScalaLightPlatformCodeInsightTestCaseAdapter

amaembo:


scala/scala-impl/test/org/jetbrains/plugins/scala/base/ScalaLightPlatformCodeInsightTestCaseAdapter.scala
```diff
@@ -95,7 +91,7 @@ abstract class ScalaLightPlatformCodeInsightTestCaseAdapter extends LightPlatfor
     disposeLibraries(getModule)
     val allJdks = ProjectJdkTable.getInstance.getAllJdks
     WriteAction.run(() => {
-      def foo() =
+      def foo(): Unit =
```
niktrop: @SrTobi I think this method should be inlined 😄

amaembo: Seems it was an artifact of automatic JavaToScala conversion. I simplified this.

niktrop:
Merged manually.

PR #479, opened 2019-08-28
parseExtendsBlock: do not add " extends " if nothing is displayed after it

amaembo:
#SCL-15751 fixed


niktrop:
Merged manually.

PR #480, opened 2019-08-28
ScalaTypedHandler: hasPrefix method fixed

amaembo:
#SCL-15468 fixed

When the "Align if-else statements" option is switched on, the explicit formatting action works correctly. The only problem is that indent is not adjusted automatically during the typing. We cannot indent initially (before `else` is typed), because `val x = if (true) 8` is a complete declaration in Scala, so it's possible that the user starts typing a new expression like

```scala
val x = if (true) 8
elseSomething(x)
```

And new expression should be aligned with `val`, not `if`. So we should realign only when `else` and subsequent space is typed. And this was already implemented! But due to a small bug, this feature wasn't working. And it appears it wasn't tested. I fixed the small bug and wrote a test (also for `match-case` which is handled nearby).

Btw I suggest to switch "Align if-else statements" on by default. I think the code looks better this way.


niktrop:
>Btw I suggest to switch "Align if-else statements" on by default. I think the code looks better this way.

@unkarjedy I agree.

niktrop:
Merged manually.

PR #481, opened 2019-08-28
projectionTypeText: decorate element via nameFun

amaembo:
#SCL-15745 fixed

Somewhat invasive change. Hopefully, it doesn't break anything but it looks reasonable to run the whole test suite before merging (I don't know how to do this).

Note that the resulting link `<:<` is not clickable. It seems it's a separate problem, as the same problem appears even if a trait is defined separately.


amaembo:
Seems this approach doesn't work, let's try in another way

PR #482, opened 2019-09-08
Support sbt-compiler-maven-plugin

knirski:
My team works on a project that uses Maven + https://github.com/sbt-compiler-maven-plugin/sbt-compiler-maven-plugin and at the same time we use IntelliJ + Scala plugin. This very small PR would make our life a little bit easier by making the IntelliJ plugin auto-detect the aforementioned Maven plugin.

We tried the already supported `scala-maven-plugin`, but it made build times considerably longer, so we chose to continue using `sbt-compiler-maven-plugin`.

I've tested this change manually, by importing https://github.com/knirski/sbt-compiler-maven-plugin-example-project and checking that source directories and Scala SDK are set correctly.


pavelfatin:
We've cherry-picked the commit, thanks!

PR #483, opened 2019-09-13
SCL-16244 Breakpoints do not work in lambda functions when the classes are reload dynamically

Alexey-NM:
https://youtrack.jetbrains.com/issue/SCL-16244

This partially solve the bug.
Part 1
fix: filterAllClasses, sometimes the function did  not remove class name from the full name, so the package was incorrect
Part 2 (It is not implemented)
org.jetbrains.plugins.scala.debugger.ScalaPositionManager.hasLocations will not work when dynamic classes is not in dependency list of module which is used to run the program


Alexey-NM:
Sorry I have mistaken with branch

PR #484, opened 2019-09-13
SCL-16244 Breakpoints do not work in lambda functions when the classes are reload dynamically

Alexey-NM:
https://youtrack.jetbrains.com/issue/SCL-16244

This partially solve the bug.
Part 1
fix: filterAllClasses, sometimes the function did not remove class name from the full name, so the package was incorrect
Part 2 (It is not implemented)
org.jetbrains.plugins.scala.debugger.ScalaPositionManager.hasLocations will not work when dynamic classes is not in dependency list of module which is used to run the program


niktrop:
@Alexey-NM Thank you for the report and investigation. I've squashed and merged your changes, did some refactoring and fixed the second problem as well:
https://github.com/JetBrains/intellij-scala/commits/d690d4a8ea8b15992d0272ab9e570be1a8189368

It should work in nightly builds for IDEA 2019.3 EAP.

Please sign our contributors license agreement if you didn't do it already:

PR #485, opened 2019-09-15
Fix the link to PSI Viewer

neshkeev:
The **PSI Viewer** link used to lead to a wrong page. This patch sets the correct page location of **PSI Viewer** on the documentation


sugakandrey:
@neshkeev Thanks!

PR #486, opened 2019-09-16
Improve getting started experience

Alexey-NM:
add: select the use sbt shell for build and import
It is very non-obvious why there are no files in intellij-scala\target\plugin\Scala after you have compiled files in the  IDEA

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ In order to take part in Scala plugin development, you need:
   ```
 
 2. Open IntelliJ IDEA, select `File -> New -> Project from existing sources`, point to
-the directory where Scala plugin repository is and then import it as sbt project.
+the directory where Scala plugin repository is and then import it as sbt project. On the sbt settings page, select the use sbt shell for build and import option and click OK.
 
 3. In the next step, select JDK 11 as project JDK (create it from an installed JDK if necessary).
 


jastice:
Thanks! Actually the setup we currently use for development does not depend on building with the sbt shell (though that also works), but instead compiles and builds the plugin artifact from IntelliJ, with sbt only generating the configurations. Did this not work for you? Maybe it's necessary to re-open the project after generating configs?

Alexey-NM:
>Did this not work for you? 

No, It did not. 

>Maybe it's necessary to re-open the project after generating configs?

I had not managed to do it.  The intellij-scala\target\plugin\Scala just was empty. 
I checked that the build  worked correctly in sbt,  and it was simple to setup.

I do not understand how "idea" fills "intellij-scala\target\plugin\Scala". If I have time I will try to find out it. Thank you.

jastice:
In the "Before launch" actions of the "IDEA" run config, is there "Build 'scala" artifact listed? If it is not, try adding it and then building.

image

Alexey-NM:
There are not any artifacts by default. 
I can add scalaComunity artifact but it does not work.
So It seems it does not work on Comunity Edition :(
It is very confusing.

image

mutcianm:
Did you run `;createIDEAArtifactXml ;idea-runner/createIDEARunConfiguration` after importing the project?

Alexey-NM:
>Did you run ;createIDEAArtifactXml ;idea-runner/createIDEARunConfiguration after importing the project?

Yes, I did. If do not run these commands there will be no run configuration.
I have rebooted "idea" either.

Alexey-NM:
I have rebooted "idea" either.

PR #487, opened 2019-10-23
Fix classBuffer overwriting the myExtensionInstances hashmap

dispalt:
It looks like this function gets called N times for N bundles.  If you have two
or more bundles with the same interface implemented (SyntheticMemberInjector or friends) the last one wins.

This code merges it instead.

I tried to change as little as possible.


mutcianm:
Thanks for the fix! I'll cherry-pick your changes into 192.x and 193.x branches manually

dispalt:
Thanks do you think this would make it to a 192 update soon?

PR #488, opened 2019-11-04
Do not rewrite { (a: Int) => ... } to { a: Int => ... }

smarter:
The latter syntax is removed from Dotty due to its irregularity:

```scala
scala> def foo : Int => Int = { a: Int => a }
1 |def foo : Int => Int = { a: Int => a }
  |                         ^^^^^^
  |parentheses are required around the parameter of a lambda
  |This construct can be rewritten automatically under -language:Scala2 -rewrite.
```

### DIFF
--- a/scala/scala-impl/src/org/jetbrains/plugins/scala/codeInspection/parentheses/ScalaUnnecessaryParenthesesInspectionBase.scala
+++ b/scala/scala-impl/src/org/jetbrains/plugins/scala/codeInspection/parentheses/ScalaUnnecessaryParenthesesInspectionBase.scala
@@ -36,17 +36,12 @@ abstract class ScalaUnnecessaryParenthesesInspectionBase
   def setSettings(settings: UnnecessaryParenthesesSettings): Unit
 
   private def hasSingleParenthesizedParam(f: ScFunctionExpr): Boolean = {
-    // If an anonymous function (x: T) => e with a single typed parameter appears as the result expression of a block, it can be abbreviated to x: T => e.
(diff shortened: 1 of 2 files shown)


niktrop:
@smarter Thank you for the contribution, it will be included in the coming 2019.3 release. Please sign our contributor license agreement: https://www.jetbrains.com/agreements/cla/

I also created a ticket for the actual fix of deprecated syntax:
https://youtrack.jetbrains.com/issue/SCL-16540

neko-kai:
@niktrop @smarter 
This is a dotty-only change that hasn't went through a SIP yet and may well be reverted. I'd prefer to have an option to ignore the deprecation/rewrite.

dwijnand:
You mean keep the rewrite?  The deprecation you'll only see if you compile it with Dotty, so not a intellij-scala change.

niktrop:
@neko-kai 
Right now only warning for syntax with parentheses is removed. If we'll add an inspection for deprecated syntax without parentheses, it will be configurable.

neko-kai:
@niktrop Got it, thank you! 👍
@dwijnand Yeah, that's what I mean

SrTobi:
I created the inspection for SCL-16540 and turned it off by default for now. Is there an official link/github-issue for this change in dotty so I can put that into the Description of the inspection?

smarter:
>  Is there an official link/github-issue for this change in dotty so I can put that into the Description of the inspection?

PR is https://github.com/lampepfl/dotty/pull/7156, it looks like we don't have a documentation page for that change currently.

PR #490, opened 2019-11-21
SCL-2584 Suggest to import scala.collection.JavaConversions._

MotiejusJuodelis:
SCL-2584 Suggest to import scala.collection.JavaConversions._ where necessary.
Does it makes sense to have JavaConversions though, since it's deprecated?


unkarjedy:
relates to :
https://youtrack.jetbrains.com/issue/SCL-2584
https://youtrack.jetbrains.com/issue/SCL-15352

scala/scala-impl/src/org/jetbrains/plugins/scala/codeInsight/intention/types/ImplicitConvertJavaToScalaCollectionIntention.scala
```diff
@@ -0,0 +1,31 @@
...
+    "java.util.Map"
+  )
+
+  override val alreadyConvertedPrefixes: Set[String] = Set("scala.collections")
+
+  override def getText: String = ScalaBundle.message("implicit.java.to.scala.conversions.hint")
+
+  override def getFamilyName: String = ImplicitConvertScalaToJavaCollectionIntention.getFamilyName
+
+  override val importPath: String = "scala.collection.JavaConversions._"
```
unkarjedy: `scala.collection.JavaConversion._` is deprecated in 2.13.0
(though the problem applies to ConvertJavaToScalaCollectionIntention as well)

unkarjedy (commented):
Thanks for the pull request

jozic:
Aren't JavaConversions deprecated in 2.10-2.12 and removed in 2.13?

julienrf:
Indeed, you should suggest importing `scala.jdk.CollectionConverters._`

https://www.scala-lang.org/api/2.13.1/scala/jdk/CollectionConverters$.html

jozic:
> Indeed, you should suggest importing `scala.jdk.CollectionConverters._`
> 
> https://www.scala-lang.org/api/2.13.1/scala/jdk/CollectionConverters$.html

the tricky part is that for 2.11-2.12 they come in a separate lib (and not even available for 2.10)
if intellij can figure out if those are in classpath and suggest using those over current suggestion of JavaConverters, that would be neat

niktrop:
@MotiejusJuodelis I think we should not encourage implicit conversions between java and scala collections and suggest only explicit `asJava` and `asScala` method calls. 

I've created a ticket about changes in 2.13:
https://youtrack.jetbrains.com/issue/SCL-16730

Do you want to implement it? You may use an extension method `ProjectPsiElementExt#scalaLanguageLevel` to choose correct import.

MotiejusJuodelis:
> @MotiejusJuodelis I think we should not encourage implicit conversions between java and scala collections and suggest only explicit `asJava` and `asScala` method calls.
> 
> I've created a ticket about changes in 2.13:
> https://youtrack.jetbrains.com/issue/SCL-16730
> 
> Do you want to implement it? You may use an extension method `ProjectPsiElementExt#scalaLanguageLevel` to choose correct import.

@niktrop yes, agree. I'm planning to come back to this and implement https://youtrack.jetbrains.com/issue/SCL-16730
Thanks

SrTobi:
@MotiejusJuodelis hey cool, write to me you need help

PR #493, opened 2019-12-18
Make bloop import work when bloop from $PATH is used

romanowski:


jastice:
Thanks! One issue is that on I think Mac and Windows, the version of the system paththat contains bloop is not necessarily passed into the IDE. 

But either way this is obsolete with https://github.com/JetBrains/intellij-scala/commit/57130a4689d93ebf57beb5cb0a9b1864c18acd5f which uses the bloop launcher instead. I'd be interested however if that causes some issues elsewhere.

PR #494, opened 2019-12-18
BSP reports compilation end

romanowski:
Adresses  https://youtrack.jetbrains.com/issue/SCL-16630

This is really hacky implementation but I managed it to work. Not sure if this is correct direction since this is generally big hack but we can make it production-ready.

The problem was that auto-tests listen for messages on COMPILATION_STATUS topic that standard compiler produce. This PR adds skeleton implementaion for bsp to publish required messages.

auto-test


jastice:
thank you! I will either merge this directly into idea201.x or figure out something less hacky

jastice:
I merged and made some adjustments :)

PR #495, opened 2019-12-19
Fix compatible runners search in isByteCodeIndexingActive

tpasternak:
While opening a project, the ScalaCompilerReferenceService
decides whether to enable bytecode indexing. It is done via
CompilerIndicesSettings::isBytecodeIndexingActive

One of the requirements is that there must exist a ProjectTaskRunner
that can run project's modules and is either an SBT runner or a JPS
runner.

Unfortunately, the `hasCompatibleRunner` indices function loooks only
for the first runner that can run all the modules and then checks if
it is ans SBT or JPS runner. If the first runner can run the modules,
or the module list is empty, the lookup will stop. Then, if it's
neither a JPS or an SBT runner, the result of `hasCompatibleRunners`
will be false, even if there is a good runner further on the list.

With this change, we search through the whole list for a good runner.

YouTrack: #SCL-16768 fixed

### DIFF
--- a/scala/scala-impl/src/org/jetbrains/plugins/scala/findUsages/compilerReferences/settings/CompilerIndicesSettings.scala
+++ b/scala/scala-impl/src/org/jetbrains/plugins/scala/findUsages/compilerReferences/settings/CompilerIndicesSettings.scala
@@ -43,7 +43,7 @@ class CompilerIndicesSettings(project: Project) extends PersistentStateComponent
   }
 
   private[this] def hasCompatibleRunner: Boolean =
-    runners.find { runner =>
+    runners.filter { runner =>
       val task = taskManager.createAllModulesBuildTask(true, project)
       val moduleBuildTasks = task match {
         case taskList: ProjectTaskList => taskList.asScala


sugakandrey:
@tpasternak Hi, thanks for taking your time to contribute this change, but I'm not quite sure I fully understand the idea behind it. Currently `hasCompatibleRunner` works as intended, i.e. it is only enabled if SBT or JPS is used to build the project, because it requires special machinery to communicate with the build tool. The main requirement is that when you press `Build project` button a supported (JPS or SBT) runner is invoked and if I remember correctly the first runner for which `runner::canRun(task)` is `true` is always chosen.

tpasternak:
The most important thing here is that, the `hasCompatibleRunner` method returns `false` if the module list is empty, event if a JPS or SBT runner is already available. There may be some cases (for example when you import a project from pants) in which the modules are loaded a little bit later. In such a case the `find` method is just returning the first runner from the list and then only this runner is checked whether it is a JPS or SBT one.
https://github.com/JetBrains/intellij-scala/blob/f55df32c25093dce4621f20488b825a3fa9c93d6/scala/scala-impl/src/org/jetbrains/plugins/scala/findUsages/compilerReferences/settings/CompilerIndicesSettings.scala#L45-L58

This results in `isByteCodeIndexingActive` returning `false` as well https://github.com/JetBrains/intellij-scala/blob/f55df32c25093dce4621f20488b825a3fa9c93d6/scala/scala-impl/src/org/jetbrains/plugins/scala/findUsages/compilerReferences/settings/CompilerIndicesSettings.scala#L60

If `isByteCodeIndexingActive` is false, the method `ScalaCompilerReferenceService.projectOpened` is doing nothing.
https://github.com/JetBrains/intellij-scala/blob/f55df32c25093dce4621f20488b825a3fa9c93d6/scala/scala-impl/src/org/jetbrains/plugins/scala/findUsages/compilerReferences/ScalaCompilerReferenceService.scala#L193-L220

sugakandrey:
@tpasternak Am I understanding it correctly, that `pants-plugin` **does not** provide an implementation of `ProjectTaskRunner` and you want to still have the possibility to use bytecode indices by building the project with JPS?

tpasternak:
It provides, but it's not the first one on `runners` list.

sugakandrey:
Which one is the first then?

tpasternak:
In my case it's `KotlinMPPProjectTaskRunner`, but can be anything. 

If there are no modules, then the result depends only on whether the first runner on the list is a JPS/SBT one.

sugakandrey:
I see, the problem with the suggest approach though, is that it enables indexing in some undesirable cases though (e.g. Maven projects). I'll have to think a bit about how we can make it work in your case.

tpasternak:
Ok, I agree, but in the old solution (with `find`) the problem occurs as well - the indexing will be enabled if a SBT/JPS module is the first one on `runners` list

tpasternak:
@sugakandrey I just checked "Import from SBT" and the problem seems to occur in this case as well. You can't call "Find usages" on instance methods until you close & reopen the project

sugakandrey:
@tpasternak Hmm, thanks for the info. I'll try to work something out, but we have a lot of stuff going on right now, so this might take a while.

tpasternak:
Ok, would you accept a workaround solution as decribed below?

1. Extract the whlole body of `ScalaCompilerReferenceService.projectOpened` method to 'initializeReferenceService' function.
2. Ensure the `initializeReferenceService` is called only once
3. Make `initializeReferenceService` available publicly, so it can be called directly from Pants/SBT import procedures

romanowski:
@sugakandrey do you think that solution above may work?

sugakandrey:
@tpasternak @romanowski Yeah, this might work.

lukaszwawrzyk:
@sugakandrey I created the PR with these changes here https://github.com/JetBrains/intellij-scala/pull/498

PR #496, opened 2019-12-31
BSP Server status Icon in Status Bar

tpasternak:
BSP Icon
This commit adds a BSP server control icon to Status Bar. It allows
user to kill the connection (and start a new one when a new build task
is spawned)

`toggleList` method in BspServerWidgetProvider is almost the same as
in CompileServerManager, but I decided not to extract them to a single
method, as probably there will never be a single reason to change them both.

### DIFF
--- a/bsp/src/org/jetbrains/bsp/BspServerWidgetProvider.scala
+++ b/bsp/src/org/jetbrains/bsp/BspServerWidgetProvider.scala
@@ -1,0 +1,94 @@
+package org.jetbrains.bsp
+
+
+import com.intellij.openapi.wm.{StatusBar, StatusBarWidget, StatusBarWidgetProvider, WindowManager}
+import java.awt.Point
+import java.awt.event.{ActionEvent, ActionListener, MouseEvent}
+
+import com.intellij.icons.AllIcons
+import com.intellij.ide.DataManager
+import com.intellij.openapi.actionSystem.{AnAction, AnActionEvent, DefaultActionGroup, Separator}
+import com.intellij.openapi.project.{DumbAware, Project}
+import com.intellij.openapi.ui.popup.JBPopupFactory
+import com.intellij.openapi.util.IconLoader
+import com.intellij.openapi.wm.StatusBar.Anchors
+import com.intellij.ui.awt.RelativePoint
+import com.intellij.util.Consumer
+import javax.swing.{Icon, Timer}
+import org.jetbrains.bsp.protocol.BspCommunicationService
+
+class BspServerWidgetProvider extends StatusBarWidgetProvider{
+
+  private val IconRunning = org.jetbrains.bsp.Icons.BSP
+
+  private val IconStopped = IconLoader.getDisabledIcon(IconRunning)
+
+  private class Widget(project: Project) extends StatusBarWidget {
+
+    private val timer = new Timer(1000, TimerListener)
+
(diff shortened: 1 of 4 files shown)


jastice:
Great addition, thanks for this! I will merge it into the idea201.x branch next year ;)
Maybe you can figure out the merge conflicts beforehand?

tpasternak:
Thanks, I'm not fully sure if liveness probe is done correctly(`def alive`), please double-check it while reviewing.

tpasternak:
bc4d5ad - rebase onto idea201.x

bsp/src/org/jetbrains/bsp/protocol/BspCommunication.scala
```diff
@@ -99,6 +99,11 @@ class BspCommunication(base: File, executionSettings: BspExecutionSettings) exte
       s.shutdown()
   }
 
+  def alive = session.get() match {
```
romanowski, line 102: Can be a oneliner: ``session.get().exisits(_.isAlive)```

romanowski (approved):
Just a nitpick :)

bsp/src/org/jetbrains/bsp/BspServerWidgetProvider.scala
```diff
@@ -0,0 +1,94 @@
...
+    private object TimerListener extends ActionListener {
+
+      override def actionPerformed(e: ActionEvent): Unit = {
+        statusBar.foreach(_.updateWidget(ID))
+      }
+    }
+
+    def connectionActive: Boolean = {
+      val inst = BspCommunicationService.getInstance
+      inst.communicate(project).alive
```
jastice, line 41: This assumes one BSP server at the project base, but multiple BSP servers per project, and servers not at the base are supported now. Ideally the widget would track all connections associated with a project

bsp/src/org/jetbrains/bsp/protocol/session/BspSession.scala
```diff
@@ -56,6 +56,8 @@ class BspSession private(bspIn: InputStream,
   private val queueProcessor = AppExecutorUtil.getAppScheduledExecutorService
       .scheduleWithFixedDelay(() => nextQueuedCommand, queuePause.toMillis, queuePause.toMillis, TimeUnit.MILLISECONDS)
 
+  def stopServer = serverConnection.cancelable.cancel()
```
jastice, line 59: should be done via org.jetbrains.bsp.protocol.BspCommunicationService#closeCommunication
or org.jetbrains.bsp.protocol.BspCommunication#closeSession

jastice:
I used this PR as a base and improved on it a bit to allow managing individual connections over the widget.

image

Thanks so much for the inspiration and groundwork!

bsp/src/org/jetbrains/bsp/BspServerWidgetProvider.scala
julienrf: May I ask in which situation we happen to have multiple BSP servers per project?

jastice: IntelliJ allows multiple workspaces to be associated with one project. So a user can import several BSP projects (and other types as well!) into a single IntelliJ project.

There's also the use case that one BSP project depends on another. @jvican proposed implementing sbt support by running a separate bloop server in sbt's `project` directory. But I don't like that so it's not implemented ;)

PR #497, opened 2020-01-24
Upgrade simple string into interpolated earlier

mirelon:
After typing `$` instead of `${`


jastice:
I wonder if this change wouldn't make the interpolator insertion a little to trigger-happy. We already insert the interpolator `s` when autocompleting a variable after a `$`

unkarjedy:
@jastice https://youtrack.jetbrains.com/issue/SCL-16932#focus=streamItem-27-3919326.0-0

niktrop:
@mirelon Thank you for your suggestion, but this behaviour is not desired in some cases. You won't be able to type any string literal containing `$` char. Check that `$` is escaped would be a proper solution for https://youtrack.jetbrains.com/issue/SCL-16932

PR #500, opened 2020-02-14
Ask BSP for test execution environment

tpasternak:
When the project is imported from BSP, it is expected that the tests
executed with IntelliJ runners will behave exactly the same as if they
would be run by BSP Server. In order to achieve this, the runner must
download some critical environment parameters from BSP server via
jvmTestEnvironment, and use them to spawn child testing process. These
parameters are:

  - classpath
  - environment variables
  - working directory
  - JVM options

The place where the parameters are set is
`BspJvmEnvironmentProgramPatcher`. Unfortunately, the
JavaProgramPatchers are executed in the Event Dispatch Thread, so it
is forbidden to do a (always blocking) BSP call inside. In order to
overcome it, the BSP call is done in BspFetchTestEnvironmentTask,
which is executed on a pool thread before each test execution. The
environment is then saved via `putUserData`, so the patcher can read
it without blocking.

Another tricky thing here is mapping from IntelliJ module to BSP
target. On the other hand, the targetId is required to receive JVM
environment via BSP. Unfortunately, there may be more than one target that matches
the module, so in cases when it's impossible to unambiguously
determine which target is the right one, a popup is shown, so it's
possible to choose from a list of potential targets. The choice is
saved as a `BspFetchTestEnvironmentTask`'s setting.

This addresses https://youtrack.jetbrains.com/issue/SCL-16934

### DIFF
--- a/bsp/resources/META-INF/BSP.xml
+++ b/bsp/resources/META-INF/BSP.xml
@@ -1,10 +1,6 @@
 <?xml version="1.0" encoding="iso-8859-1"?>
 
 <idea-plugin>
-    <projectListeners>
-        <listener class="org.jetbrains.bsp.project.test.environment.FetchEnvironmentTaskInstaller"
-                  topic="com.intellij.execution.RunManagerListener"/>
-    </projectListeners>
     <extensions defaultExtensionNs="com.intellij">
         <moduleType id="BSP_SYNTHETIC_MODULE" implementationClass="org.jetbrains.bsp.project.BspSyntheticModuleType"/>
         <moduleConfigurationEditorProvider implementation="org.jetbrains.bsp.project.BspSyntheticModuleEditorProvider"/>
@@ -45,6 +41,11 @@
         <java.programPatcher implementation="org.jetbrains.bsp.project.test.environment.BspJvmEnvironmentProgramPatcher"/>
         <stepsBeforeRunProvider implementation="org.jetbrains.bsp.project.test.environment.BspFetchTestEnvironmentTaskProvider"/>
     </extensions>
+
+    <projectListeners>
+        <listener class="org.jetbrains.bsp.project.test.environment.BspFetchEnvironmentTaskInstaller"
+                  topic="com.intellij.execution.RunManagerListener"/>
+    </projectListeners>
 
     <actions>
     </actions>
--- a/bsp/resources/META-INF/BSP.xml
+++ b/bsp/resources/META-INF/BSP.xml
@@ -1,7 +1,10 @@
 <?xml version="1.0" encoding="iso-8859-1"?>
 
 <idea-plugin>
-
+    <projectListeners>
+        <listener class="org.jetbrains.bsp.project.test.environment.FetchEnvironmentTaskInstaller"
+                  topic="com.intellij.execution.RunManagerListener"/>
+    </projectListeners>
     <extensions defaultExtensionNs="com.intellij">
         <moduleType id="BSP_SYNTHETIC_MODULE" implementationClass="org.jetbrains.bsp.project.BspSyntheticModuleType"/>
         <moduleConfigurationEditorProvider implementation="org.jetbrains.bsp.project.BspSyntheticModuleEditorProvider"/>
@@ -17,7 +20,6 @@
         <projectService serviceImplementation="org.jetbrains.bsp.settings.BspSettings"/>
         <projectService serviceImplementation="org.jetbrains.bsp.settings.BspLocalSettings"/>
         <projectService serviceImplementation="org.jetbrains.bsp.BspBuildLoopService"/>
-        <projectService serviceImplementation="org.jetbrains.bsp.project.test.environment.BspFetchEnvironmentTaskInstaller"/>
 
         <applicationService serviceImplementation="org.jetbrains.bsp.protocol.BspCommunicationService"/>
 
--- a/bsp/resources/META-INF/BSP.xml
+++ b/bsp/resources/META-INF/BSP.xml
@@ -41,6 +41,7 @@
         <statusBarWidgetProvider implementation="org.jetbrains.bsp.BspServerWidgetProvider"/>
 
         <postStartupActivity implementation="org.jetbrains.bsp.project.test.environment.BspFetchEnvironmentTaskInstaller$Startup"/>
+        <java.programPatcher implementation="org.jetbrains.bsp.project.test.environment.BspJvmEnvironmentProgramPatcher"/>
     </extensions>
 
     <actions>
--- a/bsp/resources/META-INF/BSP.xml
+++ b/bsp/resources/META-INF/BSP.xml
@@ -17,6 +17,7 @@
         <projectService serviceImplementation="org.jetbrains.bsp.settings.BspSettings"/>
         <projectService serviceImplementation="org.jetbrains.bsp.settings.BspLocalSettings"/>
         <projectService serviceImplementation="org.jetbrains.bsp.BspBuildLoopService"/>
+        <projectService serviceImplementation="org.jetbrains.bsp.project.test.environment.BspFetchEnvironmentTaskInstaller"/>
 
         <applicationService serviceImplementation="org.jetbrains.bsp.protocol.BspCommunicationService"/>
 
@@ -39,6 +40,7 @@
 
         <statusBarWidgetProvider implementation="org.jetbrains.bsp.BspServerWidgetProvider"/>
 
+        <postStartupActivity implementation="org.jetbrains.bsp.project.test.environment.BspFetchEnvironmentTaskInstaller$Startup"/>
     </extensions>
 
     <actions>
--- a/bsp/resources/META-INF/BSP.xml
+++ b/bsp/resources/META-INF/BSP.xml
@@ -42,8 +42,8 @@
 
         <statusBarWidgetProvider implementation="org.jetbrains.bsp.BspServerWidgetProvider"/>
 
-        <postStartupActivity implementation="org.jetbrains.bsp.project.test.environment.BspFetchEnvironmentTaskInstaller$Startup"/>
         <java.programPatcher implementation="org.jetbrains.bsp.project.test.environment.BspJvmEnvironmentProgramPatcher"/>
+        <stepsBeforeRunProvider implementation="org.jetbrains.bsp.project.test.environment.BspFetchTestEnvironmentTaskProvider"/>
     </extensions>
 
     <actions>
--- a/bsp/src/org/jetbrains/bsp/BSP.scala
+++ b/bsp/src/org/jetbrains/bsp/BSP.scala
@@ -17,8 +17,4 @@ object BSP {
   val ProjectSystemId = new ProjectSystemId("BSP", Name)
 
   val balloonNotification: NotificationGroup = NotificationGroup.balloonGroup(Name)
-
-  def isBspModule(module: Module) = {
-    ExternalSystemApiUtil.isExternalSystemAwareModule(BSP.ProjectSystemId.getId, module)
-  }
 }
--- a/bsp/src/org/jetbrains/bsp/BSP.scala
+++ b/bsp/src/org/jetbrains/bsp/BSP.scala
@@ -2,6 +2,9 @@ package org.jetbrains.bsp
 
 import com.intellij.notification.NotificationGroup
 import com.intellij.openapi.externalSystem.model.ProjectSystemId
+import com.intellij.openapi.externalSystem.util.ExternalSystemApiUtil
+import com.intellij.openapi.module.Module
+import com.intellij.openapi.util.registry.Registry
 import javax.swing.Icon
 import org.jetbrains.annotations.Nls
 
@@ -14,4 +17,8 @@ object BSP {
   val ProjectSystemId = new ProjectSystemId("BSP", Name)
 
   val balloonNotification: NotificationGroup = NotificationGroup.balloonGroup(Name)
+
+  def isBspModule(module: Module) = {
+    ExternalSystemApiUtil.isExternalSystemAwareModule(BSP.ProjectSystemId.getId, module)
+  }
 }
--- a/bsp/src/org/jetbrains/bsp/project/test/environment/BspFetchEnvironmentTaskInstaller.scala
+++ b/bsp/src/org/jetbrains/bsp/project/test/environment/BspFetchEnvironmentTaskInstaller.scala
@@ -1,10 +1,11 @@
 package org.jetbrains.bsp.project.test.environment
 
 import com.intellij.execution.{RunManagerEx, RunManagerListener, RunnerAndConfigurationSettings}
+import com.intellij.openapi.project.Project
 
-class BspFetchEnvironmentTaskInstaller extends RunManagerListener {
+class BspFetchEnvironmentTaskInstaller(project: Project) extends RunManagerListener {
   override def runConfigurationAdded(settings: RunnerAndConfigurationSettings): Unit = {
-    val runManager = RunManagerEx.getInstanceEx(settings.getConfiguration.getProject)
+    val runManager = RunManagerEx.getInstanceEx(project)
     val runConfiguration = settings.getConfiguration
     if (BspTesting.isBspRunnerSupportedConfiguration(runConfiguration)) {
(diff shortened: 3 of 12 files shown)


jastice:
Thank you! I'll handle this next week

bsp/src/org/jetbrains/bsp/project/test/environment/BspFetchEnvironmentTaskInstaller.scala
```diff
@@ -0,0 +1,28 @@
+package org.jetbrains.bsp.project.test.environment
+
+import com.intellij.execution.{RunManagerEx, RunManagerListener, RunnerAndConfigurationSettings}
+import com.intellij.openapi.components.ProjectComponent
+import com.intellij.openapi.project.Project
+
+class BspFetchEnvironmentTaskInstaller(project: Project) extends ProjectComponent {
```
jastice: Components are deprecated and we're removing them from the plugin in 2020.1
you should be able to directly install this listener in the BSP.xml

bsp/src/org/jetbrains/bsp/BSP.scala
```diff
@@ -12,4 +14,8 @@ object BSP {
   val ProjectSystemId = new ProjectSystemId("BSP", Name)
 
   val balloonNotification: NotificationGroup = NotificationGroup.balloonGroup("bsp")
+
+  def isBspModule(module: Module) = {
+    ExternalSystemApiUtil.isExternalSystemAwareModule(BSP.ProjectSystemId.getId, module)
```
jastice: I recently added this same method in BspUtil

bsp/src/org/jetbrains/bsp/project/test/environment/BspFetchTestEnvironmentTaskProvider.scala
```diff
@@ -0,0 +1,222 @@
...
+      case _ => true
+    }
+
+  }
+
+  private def askUserForTargetId(targetIds: Seq[String], task: BspFetchTestEnvironmentTask): Option[String] = {
+    var chosenTarget: Option[String] = None
+    ApplicationManager.getApplication.invokeAndWait { () => {
+      chosenTarget = Option(Messages.showEditableChooseDialog(
+        "Could not infer Pants target, please choose one from the list",
```
jastice: Is this code specific to Pants? I think the text could be more general

bsp/src/org/jetbrains/bsp/project/test/environment/BspFetchTestEnvironmentTaskProvider.scala
```diff
@@ -0,0 +1,222 @@
...
+    chosenTarget
+  }
+
+  private def fetchTargetIdsFromFiles(files: Seq[PsiFile],
+                                      workspace: URI,
+                                      project: Project,
+                                      potentialTargets: Seq[BuildTargetIdentifier]): Seq[BuildTargetIdentifier] = {
+    val communication: BspCommunication = BspCommunication.forWorkspace(workspace.toFile)
+    val bspTaskId: EventId = BuildMessages.randomEventId
+    val report = new BuildToolWindowReporter(project = project, buildId = bspTaskId, title = "bsp fetch sources")
```
jastice: The message sounds a bit misleading

bsp/src/org/jetbrains/bsp/project/test/environment/BspFetchTestEnvironmentTaskProvider.scala
```diff
@@ -0,0 +1,222 @@
...
+  }
+
+  private def processLog(report: BuildToolWindowReporter): ProcessLogger = { message =>
+    report.log(message)
+  }
+
+  private def fetchJvmTestEnvironment(target: BuildTargetIdentifier, workspace: URI, project: Project): JvmTestEnvironment = {
+    val communication: BspCommunication = BspCommunication.forWorkspace(workspace.toFile)
+    val bspTaskId: EventId = BuildMessages.randomEventId
+    val report = new BuildToolWindowReporter(project, bspTaskId, "bsp build")
```
jastice: message doesn't look accurate :)

bsp/src/org/jetbrains/bsp/project/test/environment/BspFetchTestEnvironmentTaskProvider.scala
```diff
@@ -0,0 +1,222 @@
...
+    val environment = response.getItems.asScala.head
+    JvmTestEnvironment(
+      classpath = environment.getClasspath.asScala.map(x => new URI(x).getPath),
+      workdir = environment.getWorkingDirectory,
+      environmentVariables = environment.getEnvironmentVariables.asScala.toMap,
+      jvmOptions = environment.getJvmOptions.asScala.toList
+    )
+  }
+
+  private def getBspTargets(module: Module): Option[Seq[BuildTargetIdentifier]] =
```
jastice, line 238: we can reuse org.jetbrains.bsp.data.BspMetadata#get for most of this

scala/scala-impl/resources/META-INF/scala-plugin-common.xml
```diff
@@ -54,6 +54,8 @@
     </extensions>
 
     <extensions defaultExtensionNs="com.intellij">
+      <java.programPatcher implementation="org.jetbrains.bsp.project.test.environment.BspProgramPatcher"/>
```
jastice: these should be part of BSP.xml

bsp/src/org/jetbrains/bsp/project/test/environment/BspFetchTestEnvironmentTaskProvider.scala
```diff
@@ -0,0 +1,222 @@
...
+                                      project: Project,
+                                      potentialTargets: Seq[BuildTargetIdentifier]): Seq[BuildTargetIdentifier] = {
+    val communication: BspCommunication = BspCommunication.forWorkspace(workspace.toFile)
+    val bspTaskId: EventId = BuildMessages.randomEventId
+    val report = new BuildToolWindowReporter(project = project, buildId = bspTaskId, title = "bsp fetch sources")
+    val sources: util.List[SourcesItem] = Await.result(communication.run(
+      bspSessionTask = getFiles(potentialTargets)(_, _),
+      notifications = _ => (),
+      processLogger = processLog(report),
+    ).future, 300.millis).getItems
```
jastice: may fail if the server first needs to initialize something or blocks for a while for some other reason, since there is no retry happening here

bsp/src/org/jetbrains/bsp/project/test/environment/BspFetchEnvironmentTaskInstaller.scala
```diff
@@ -0,0 +1,28 @@
...
+  override def projectOpened(): Unit = {
+    val runManager = RunManagerEx.getInstanceEx(project)
+    project.getMessageBus.connect().subscribe(RunManagerListener.TOPIC,
+      new RunManagerListener {
+        override def runConfigurationAdded(settings: RunnerAndConfigurationSettings): Unit = {
+          val runConfiguration = settings.getConfiguration
+          if (BspTesting.isBspRunnerSupportedConfiguration(runConfiguration)) {
```
jastice: should also account for if the server supports `jvmTestEnvironment` request at all

bsp/src/org/jetbrains/bsp/project/test/environment/BspFetchEnvironmentTaskInstaller.scala
```diff
@@ -0,0 +1,28 @@
...
+            val beforeRunTasks = runManager.getBeforeRunTasks(runConfiguration)
+            val task = new BspFetchTestEnvironmentTask
+            task.setEnabled(true)
+            beforeRunTasks.add(task)
```
jastice: Instead of adding it before run, could we just get and save the data on project import?

jastice (changes requested):
Thanks for the PR! As to the general approach, couldn't we just request the data on project import instead of before every test? Also we need to account for the server not supporting the requests used here at all. Maybe this needs some kind of `capabilities` addition in the protocol.

Also, please rebase onto idea201.x branch

bsp/src/org/jetbrains/bsp/project/test/environment/BspFetchEnvironmentTaskInstaller.scala
tpasternak: Hi, it was already discussed here https://github.com/build-server-protocol/build-server-protocol/pull/107#discussion_r375338681 It seems it wouldn't be the right approach, as the classpath may change between compilations.

bsp/src/org/jetbrains/bsp/project/test/environment/BspFetchTestEnvironmentTaskProvider.scala
tpasternak: So what would you recommend here? Increase timeout, do a fixed number retries, or do cancellable retries? 

If cancellable, what would be best the way to trigger cancellation then?

jastice: At the other places I use `org.jetbrains.bsp.protocol.BspJob#waitForJobCancelable`. Maybe it needs a small adaption for when there is no indicator available.

bsp/src/org/jetbrains/bsp/BSP.scala
tpasternak: Done

bsp/src/org/jetbrains/bsp/project/test/environment/BspFetchTestEnvironmentTaskProvider.scala
tpasternak: Done

bsp/src/org/jetbrains/bsp/project/test/environment/BspFetchTestEnvironmentTaskProvider.scala
tpasternak: Done

scala/scala-impl/resources/META-INF/scala-plugin-common.xml
tpasternak: Done

bsp/src/org/jetbrains/bsp/project/test/environment/BspFetchEnvironmentTaskInstaller.scala
tpasternak: After some changes, the server capability is checked right before the execution

tpasternak: After some changes, the server capability is checked right before the execution

bsp/resources/META-INF/BSP.xml
```diff
@@ -17,6 +17,7 @@
         <projectService serviceImplementation="org.jetbrains.bsp.settings.BspSettings"/>
         <projectService serviceImplementation="org.jetbrains.bsp.settings.BspLocalSettings"/>
         <projectService serviceImplementation="org.jetbrains.bsp.BspBuildLoopService"/>
+        <projectService serviceImplementation="org.jetbrains.bsp.project.test.environment.BspFetchEnvironmentTaskInstaller"/>
```
jastice: Doesn't need to be a service at all as no state is maintained. Can be registered as a listener directly something like
```
    <projectListeners>
        <listener class="org.jetbrains.bsp.project.test.environment.BspFetchEnvironmentTask"
                  topic="com.intellij.execution.RunManagerListener" />
    </projectListeners>
```

bsp/src/org/jetbrains/bsp/project/test/environment/BspFetchTestEnvironmentTaskProvider.scala
```diff
@@ -0,0 +1,257 @@
...
+
+case class JvmTestEnvironment(
+                               classpath: Seq[String],
+                               workdir: String,
+                               environmentVariables: Map[String, String],
+                               jvmOptions: Seq[String]
+                             )
+
+case object JvmTestEnvironmentNotSupported extends Throwable {
+  val msg = "Build Server does not support 'buildTarget/jvmTestEnvironment' endpoint"
```
jastice: should be moved to bundle

bsp/src/org/jetbrains/bsp/project/test/environment/BspFetchTestEnvironmentTaskProvider.scala
```diff
@@ -0,0 +1,257 @@
...
+}
+
+class BspFetchTestEnvironmentTaskProvider extends BeforeRunTaskProvider[BspFetchTestEnvironmentTask] {
+  private val logger = Logger.getInstance(classOf[BspCommunication])
+
+  override def getId: Key[BspFetchTestEnvironmentTask] = BspFetchTestEnvironmentTask.runTaskKey
+
+  override def getName: String = "Use BSP environment"
```
jastice: string should be moved to bundle

bsp/src/org/jetbrains/bsp/project/test/environment/BspFetchTestEnvironmentTaskProvider.scala
```diff
@@ -0,0 +1,257 @@
...
+    }
+
+  }
+
+  private def askUserForTargetId(targetIds: Seq[String], task: BspFetchTestEnvironmentTask): Option[String] = {
+    var chosenTarget: Option[String] = None
+    ApplicationManager.getApplication.invokeAndWait { () => {
+      chosenTarget = Option(Messages.showEditableChooseDialog(
+        "Could not infer BSP target, please choose one from the list",
+        "Choose pants target",
```
jastice: should only reference pants if code only appears in pants integration

scala/scala-impl/resources/META-INF/scala-plugin-common.xml
```diff
@@ -52,6 +52,7 @@
     </extensions>
 
     <extensions defaultExtensionNs="com.intellij">
+      <stepsBeforeRunProvider implementation="org.jetbrains.bsp.project.test.environment.BspFetchTestEnvironmentTaskProvider"/>
```
jastice: should be in BSP.xml

scala/scala-impl/resources/META-INF/scala-plugin-common.xml
```diff
@@ -1426,7 +1427,7 @@
         <resolveScopeProvider implementation="org.jetbrains.plugins.scala.lang.psi.impl.ScalaOutOfSourcesResolveScopeProvider"/>
     </extensions>
 
-    <applicationListeners>
```
jastice: undo unindent :)

bsp/src/org/jetbrains/bsp/project/test/environment/BspFetchTestEnvironmentTaskProvider.scala
```diff
@@ -0,0 +1,257 @@
...
+            .orElse(if (targetsMatchingSources.length == 1) targetsMatchingSources.headOption.map(_.getUri) else None)
+            .orElse(askUserForTargetId(potentialTargets.map(_.getUri), task))
+            .toRight(BspGetEnvironmentError("Could not choose any target ID"))
+          testEnvironment <-
+            fetchJvmTestEnvironment(new BuildTargetIdentifier(chosenTargetId), workspaceUri, module.getProject)
+              .map(Right(_))
+              .recover{
+                case err: JvmTestEnvironmentNotSupported.type => Left(BspGetEnvironmentError(err.msg))
+              }
+              .getOrElse(Left(BspGetEnvironmentError("Failed to fetch test JVM environment")))
```
jastice: and all these strings should be moved to bundle if they are potentially user-facing

jastice (changes requested):
Thanks! Looks like mostly just the strings need to be moved to the bundle. There's a convenient quickfix for that if you didn't notice. And otherwise the listener doesn't need a service to install it.

jastice (approved):
Thanks for the updates! I think it's ready to be merged!

jastice:
Manually merged after rebase and some changes.
https://github.com/JetBrains/intellij-scala/commit/b30f1712c2bc3182d8c00a05eacf056724a6f55a

PR #501, opened 2020-02-16
Add support for the context-applied scalac plugin

ybasket:
Extends the support for context bounds to also virtually generate implicit parameters which mimic the syntactic sugar provided by the [context-applied](https://github.com/augustjune/context-applied) scalac plugin. This allows users of this plugin to edit their code in IntelliJ without it complaining, while it doesn't change anything for Scala projects not using the plugin.

The limitations of context-applied (no parameter with same name, no support in `AnyVal`) are respected and have their own test cases.

Example of how the internal expansion looks like for a piece of code relying on _context-applied_:
```scala
trait Log[F[_]] {
  def log(msg: String): F[Unit]
}

class Foo(val dummy: Boolean) {
  def bar[T[_]: Log: Console] = T.log("Hello")
}
```
becomes virtually (the first two implicit params are generated already by IntelliJ, only the third (`T`) is new):
```scala
// Log definiton
class Foo(val dummy: Boolean) {
  def bar[T[_]](implicit bar$Log$0: Log[T], bar$Console$0: Console[T], T: Console[T] with Log[T]) = T.log("Hello")
}
```


ybasket:
@sugakandrey did you have a chance to check this?

sugakandrey:
@ybasket I only had a quick look, sorry. I'll try to check this soon.

sugakandrey:
@ybasket lgtm, I [cherry picked](https://github.com/JetBrains/intellij-scala/commits/idea201.x-context-applied) your changes with some very minor fixes, they will be merged automatically.

PR #502, opened 2020-02-19
SCL-17067: provide title for code folding settings

Undin:
It should fix NPE on 2020.1 platform and make the corresponding settings more structured

| Before | After |
|-|-|
| image | image |

#SCL-17067 fixed
#SCL-16722 fixed


scala/scala-impl/src/org/jetbrains/plugins/scala/editor/codeFolding/ScalaCodeFoldingOptionsProvider.java
```diff
@@ -13,7 +13,7 @@
 public class ScalaCodeFoldingOptionsProvider extends BeanConfigurable<ScalaCodeFoldingSettings> implements CodeFoldingOptionsProvider {
 
   public ScalaCodeFoldingOptionsProvider() {
-    super(ScalaCodeFoldingSettings.getInstance());
+    super(ScalaCodeFoldingSettings.getInstance(), "Scala");
```
Undin, line 16: Should I put `"Scala"` literal into message bundle?

unkarjedy:
fixed in https://youtrack.jetbrains.com/issue/SCL-17067#focus=streamItem-27-3951570.0-0

PR #503, opened 2020-02-27
Set packagePrefix from BSP

lukaszwawrzyk:
See https://github.com/build-server-protocol/build-server-protocol/pull/109


jastice:
Thanks, looks good! I'll revist once merged into BSP

jastice (approved):
Thanks for the update! I will merge this today.

PR #504, opened 2020-03-02
SCL-16246: Compare substitutors in ScalaResolveResult's equals

anivaniuk:
Seems like `org.jetbrains.plugins.scala.lang.resolve.ScalaResolveResult` `equals` method should compare also substitutor and comment
```scala
  //In valid program we should not have two resolve results with the same element but different substitutor,
  // so factor by element
```
looks not relevant.

For the following code we'll have 2 resolve results with 1 element (`implicit val info: Info[A]`) with 2 different substitutors (one from `object BaseBar extends BaseBarObj[BaseBar]` - `A -> BaseBar` and one from `object Bar extends BaseBarObj[Bar]` - `A -> Bar`) which looks completely fine. And one of these resolve results should be used to find implicit for `implicitly[Info[Bar]]`. But currently only one can be in `Set` of candidates.

```scala
trait Info[A]

trait BaseBarObj[A] {
  implicit val info: Info[A] = ???
}

trait BaseBar
object BaseBar extends BaseBarObj[BaseBar]

trait Bar extends BaseBar
object Bar extends BaseBarObj[Bar]

implicitly[Info[Bar]]
```

This problem also lead to "No implicit found" for `implicitly[Factory[Int, Seq[Int]]]` and similar Factory implicits.


sugakandrey:
@AiiR42 Thanks for taking your time to look into this, that comment is indeed incorrect and I've been working on a fix to said problem. A fix similar to yours was my first idea, but unfortunately substitutors are not very well suited for equality comparison, so I'm currently investigating other possibilities.

anivaniuk:
@sugakandrey Got it. Thank you for your comment!

Another approach I was thinking on is introducing some kind of unification process for resolve results. Like unifying resolve result with empty Substitutor and result with single ThisTypeSubstitution will result to second result.
So there will be no need to do such `ThisTypeSubstitution` hack in substitutors equals. Also we'll not be dependent on order of adding resolve results.

Is that makes sense? What is your concern regarding substitutors comparison?

sugakandrey:
@AiiR42 Basically, two things: `ScalaResolveResult`s pointing to the same element don't necessarily have "equal" substitutors, since we don't really deduplicate their contents and secondly it can be quite costly to compare substitutors for every single resolve operation.

The idea is that for implicit search **only** we can probably rely solely on the enclosing type definition or lack thereof to conclude equality of resolve results.

anivaniuk:
@sugakandrey Ah, ok. As that relates to performance I guess I can close this request. Don’t think I have enough expertise for such changes.
Thank you!

PR #506, opened 2020-03-10
Correctly add BspFetchTestEnvironmentTask to run configurations

tpasternak:
The Scala plugin tries to install BspFetchTestEnvironmentTask on
`runConfigurationAdded` event. Unfortunately, it requires
RunConfigurationEx service to be installed, and the event happens
during the initialization of it.

Apart from that, the `beforeRunTasks.add` was previously called
uncoditionally. This caused multiple additions of the task.

### DIFF
--- a/bsp/src/org/jetbrains/bsp/project/test/environment/BspFetchEnvironmentTaskInstaller.scala
+++ b/bsp/src/org/jetbrains/bsp/project/test/environment/BspFetchEnvironmentTaskInstaller.scala
@@ -1,38 +1,40 @@
 package org.jetbrains.bsp.project.test.environment
 
 import com.intellij.execution.{RunManager, RunManagerEx, RunManagerListener, RunnerAndConfigurationSettings}
+import com.intellij.openapi.components.ServiceManager
+import com.intellij.openapi.project.Project
 
-class BspFetchEnvironmentTaskInstaller extends RunManagerListener {
-  private var runManagerOpt: Option[RunManagerEx] = None
 
-  // temp solution to SCL-17155 to avoid cyclid dependencies
-  // some configurations are added during RunManager initialization
+class BspFetchEnvironmentTaskInstaller(project: Project) extends RunManagerListener {
   private var settingsToInit: List[RunnerAndConfigurationSettings] = Nil
 
   override def runConfigurationAdded(settings: RunnerAndConfigurationSettings): Unit = {
-    val runManager = runManagerOpt.getOrElse {
-      settingsToInit ::= settings
-      return
-    }
+    runManagerEx.foreach(installFetchEnvironmentTask(settings, _))
+  }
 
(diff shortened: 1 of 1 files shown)


unkarjedy:
@tpasternak Please update from the master. I've pushed the fix for cyclic services declaration yesterday

tpasternak:
Thanks, rebased and adapted to your solution (which is probably more safe than my previous one)

bsp/src/org/jetbrains/bsp/project/test/environment/BspFetchEnvironmentTaskInstaller.scala
```diff
@@ -1,38 +1,40 @@
...
+        beforeRunTasks.add(task)
       }
     }
   }
 
   override def stateLoaded(runManager: RunManager, isFirstLoadState: Boolean): Unit = {
     if (isFirstLoadState) {
-      this.runManagerOpt = Some(runManager.asInstanceOf[RunManagerEx])
-      settingsToInit.foreach(runConfigurationAdded)
+      runManagerEx.foreach {
```
unkarjedy, line 30: IMO we should use the instance provided in parameters to be clear

unkarjedy: so looks like no need in `project` parameter in the Service

bsp/src/org/jetbrains/bsp/project/test/environment/BspFetchEnvironmentTaskInstaller.scala
```diff
@@ -1,38 +1,40 @@
...
+        runManager => settingsToInit.foreach(installFetchEnvironmentTask(_, runManager))
```
unkarjedy, line 31: `settingsToInit` are always empty now
Shouldn't we patch the loaded configurations?
For some reason when initializing `RunManager` platform calls `runConfigurationAdded` for the saved run configurations (not after the service initialization). In my fix commit, I postponed the initialization of saved run configurations until `stateLoaded` method.

jastice (approved):

unkarjedy:
I forgot to press "Add a review" yesterday, nevermind, we will fix ^

bsp/src/org/jetbrains/bsp/project/test/environment/BspFetchEnvironmentTaskInstaller.scala
tpasternak: Yes, sorry for that. It was clear for me, that the `settingsToInit` should be filled in `runConfigurationAdded` handler, but I broke it while refactoring code.

tpasternak:
@unkarjedy I can fix it right now, but please let me know, I wouldn't like to collide with you again

jastice:
@tpasternak I just pushed a fix

tpasternak:
Ok, thanks and sorry

PR #507, opened 2020-03-13
Don't assume ScalaTest run cunfiguration

tpasternak:
When configuring BspFetchTestEnvironmentTask, it should not be assumed
the runconfig is ScalaTest

### DIFF
--- a/bsp/src/org/jetbrains/bsp/project/test/environment/BspFetchTestEnvironmentTaskProvider.scala
+++ b/bsp/src/org/jetbrains/bsp/project/test/environment/BspFetchTestEnvironmentTaskProvider.scala
@@ -60,12 +60,16 @@ class BspFetchTestEnvironmentTaskProvider extends BeforeRunTaskProvider[BspFetch
   override def isConfigurable: Boolean = true
 
   override def configureTask(context: DataContext, configuration: RunConfiguration, task: BspFetchTestEnvironmentTask): Promise[lang.Boolean] = {
-    val module = configuration.asInstanceOf[ScalaTestRunConfiguration].getModule
-    val res = for {
-      potentialTargets <- getBspTargets(module)
-      _ <- askUserForTargetId(potentialTargets.map(_.getUri), task)
-    } yield ()
-    Promises.resolvedPromise(res.isDefined)
+    configuration match {
+      case moduleBasedConfiguration: ModuleBasedConfiguration[_,_] =>
+        val modules = moduleBasedConfiguration.getModules
+        val res = for {
+          potentialTargets <- modules.flatMap(module => getBspTargets(module))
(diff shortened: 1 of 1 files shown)


bsp/src/org/jetbrains/bsp/project/test/environment/BspFetchTestEnvironmentTaskProvider.scala
```diff
@@ -60,12 +60,16 @@ class BspFetchTestEnvironmentTaskProvider extends BeforeRunTaskProvider[BspFetch
...
 
   override def configureTask(context: DataContext, configuration: RunConfiguration, task: BspFetchTestEnvironmentTask): Promise[lang.Boolean] = {
-    val module = configuration.asInstanceOf[ScalaTestRunConfiguration].getModule
-    val res = for {
-      potentialTargets <- getBspTargets(module)
-      _ <- askUserForTargetId(potentialTargets.map(_.getUri), task)
-    } yield ()
-    Promises.resolvedPromise(res.isDefined)
+    configuration match {
+      case moduleBasedConfiguration: ModuleBasedConfiguration[_,_] =>
```
unkarjedy, line 64: If it's only about tests,  maybe we should only handle `AbstractTestRunConfiguration`, which is base for all supported test frameworks? Not to "pollute" all unrelated configurations.

jastice (approved):
thanks!

jastice:
merged/rebased via build server: https://github.com/JetBrains/intellij-scala/commit/e59062ecfbfb90804080b33d0a3c8cde2f65222b

tpasternak:
thanks!

unkarjedy:
turned out I again didn't press "Submit review"... 🤦‍♂️
who designed this GitHub?)

bsp/src/org/jetbrains/bsp/project/test/environment/BspFetchTestEnvironmentTaskProvider.scala
tpasternak: But then you couldn't support JUnit

unkarjedy: ah, I see, so maybe add `if BspTesting.isBspRunnerSupportedConfiguration`

tpasternak: Yeah, that might be good, but we already check this before adding task to the run config https://github.com/JetBrains/intellij-scala/blob/33de905217cc97759d367c2659a2ee1352baa8ed/bsp/src/org/jetbrains/bsp/project/test/environment/BspFetchTestEnvironmentTaskInstaller.scala#L22

PR #508, opened 2020-03-16
Try to guess BSP target based on JUnit configuration

tpasternak:
Before this commit, the test runner was always asking user to choose BSP target. Now, in two certain cases, it detects it automatically.

### DIFF
--- a/bsp/src/org/jetbrains/bsp/project/test/environment/BspFetchTestEnvironmentTaskProvider.scala
+++ b/bsp/src/org/jetbrains/bsp/project/test/environment/BspFetchTestEnvironmentTaskProvider.scala
@@ -8,6 +8,7 @@ import java.util.concurrent.CompletableFuture
 import ch.epfl.scala.bsp4j._
 import com.intellij.execution.BeforeRunTaskProvider
 import com.intellij.execution.configurations.{ModuleBasedConfiguration, RunConfiguration}
+import com.intellij.execution.junit.JUnitConfiguration
 import com.intellij.execution.runners.ExecutionEnvironment
 import com.intellij.openapi.actionSystem.DataContext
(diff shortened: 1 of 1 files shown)


bsp/src/org/jetbrains/bsp/project/test/environment/BspFetchTestEnvironmentTaskProvider.scala
```diff
@@ -195,12 +196,22 @@ class BspFetchTestEnvironmentTaskProvider extends BeforeRunTaskProvider[BspFetch
...
               case data: ClassTestData => List(data.testClassPath)
               case _ => List()
             }
+          case junitConfig: JUnitConfiguration => {
+            junitConfig.getTestType match {
+              case JUnitConfiguration.TEST_METHOD => {
+                List(junitConfig.getPersistentData.getMainClassName)
+              }
+              case JUnitConfiguration.TEST_CLASS => {
+                List(junitConfig.getPersistentData.getMainClassName)
```
jastice: the same code as for `TEST_METHOD`. intentional?

tpasternak: We need to extract class name here, so yes. Btw I just noticed the default case was not handled here, so I force-pushed the change.

jastice: Ok. I'd put it in the same condition then

tpasternak: right!

jastice (approved):

PR #514, opened 2020-03-25
Read jdk from bsp server

lukaszwawrzyk:
depends on PRs to bloop and bsp
For now the information about jdk from bsp is used to setup the project sdk (not for individual modules)
https://github.com/build-server-protocol/build-server-protocol/pull/113
https://github.com/scalacenter/bloop/pull/1214

### DIFF
--- a/bsp/src/org/jetbrains/bsp/project/resolver/BspResolverLogic.scala
+++ b/bsp/src/org/jetbrains/bsp/project/resolver/BspResolverLogic.scala
@@ -9,20 +9,28 @@ import ch.epfl.scala.bsp4j._
 import com.google.gson.{Gson, JsonElement}
 import com.intellij.openapi.externalSystem.model.project._
 import com.intellij.openapi.externalSystem.model.{DataNode, ProjectKeys}
+import com.intellij.openapi.externalSystem.util.ExternalSystemApiUtil
 import com.intellij.openapi.module.StdModuleTypes
 import com.intellij.openapi.roots.DependencyScope
 import com.intellij.openapi.util.io.FileUtil
 import org.jetbrains.bsp.BspUtil._
+import org.jetbrains.bsp.data.JdkData
 import org.jetbrains.bsp.data.{BspMetadata, BspProjectData, SbtBuildModuleDataBsp, ScalaSdkData}
 import org.jetbrains.bsp.project.BspSyntheticModuleType
 import org.jetbrains.bsp.project.resolver.BspResolverDescriptors._
 import org.jetbrains.bsp.{BSP, BspBundle, BspErrorMessage}
 import org.jetbrains.plugins.scala.project.Version
+import org.jetbrains.plugins.scala.project.external.JdkByHome
+import org.jetbrains.plugins.scala.project.external.JdkByVersion
+import org.jetbrains.plugins.scala.project.external.SdkReference
 
 import scala.collection.JavaConverters._
 import scala.collection.mutable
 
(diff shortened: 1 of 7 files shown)


bsp/src/org/jetbrains/bsp/project/resolver/BspResolverLogic.scala
```diff
@@ -395,17 +396,17 @@ private[resolver] object BspResolverLogic {
...
     val projectRootPath = workspace.getCanonicalPath
     val moduleFilesDirectoryPath = moduleFilesDirectory(workspace).getCanonicalPath
     val projectRoot = new File(projectRootPath)
     val projectData = new ProjectData(BSP.ProjectSystemId, projectRoot.getName, projectRootPath, projectRootPath)
     val projectNode = new DataNode[ProjectData](ProjectKeys.PROJECT, projectData, null)
-
-    // TODO have a way to set project jdk before/during import
-    val bspProjectData = new DataNode[BspProjectData](BspProjectData.Key, BspProjectData(None), projectNode)
-    projectNode.addChild(bspProjectData)
+    val bspProjectData = jvmEnvironment.map { env =>
```
jastice: This makes creation of project dependent on existens of a jvmEnvironment. But the existence of such an endpoint isn't and shouldn't be mandatory

bsp/src/org/jetbrains/bsp/project/resolver/BspProjectResolver.scala
```diff
@@ -271,6 +270,15 @@ object BspProjectResolver {
       .catchBspErrors
       .reportFinished(reporter, sourcesEventId, message, BspBundle.message("bsp.resolver.request.failed.buildtarget.sources"))
 
+    val jvmEnvironment = {
```
jastice: jvm environment should be primarily target-scoped in BSP. IDEA's "Project JDK" can be considered just a conveience setting. We could have an optional endpoint for a workspace-wise environment, but I'm not sure it's necessary. We could just infer a default Project JDK from individual modules, or use whatever is available.

jastice (changes requested):

bsp/resources/messages/ScalaBspBundle.properties
```diff
@@ -37,6 +37,8 @@ bsp.resolver.resolving.build.structure.failed=resolving BSP build structure fail
 bsp.resolver.import.failed=import failed
 bsp.resolver.creating.bloop.configuration.from.sbt=creating Bloop configuration from sbt
 bsp.resolver.sources=sources
+bsp.resolver.jvmenvironment=jvm environment
```
jastice: capitalize JVM

project/dependencies.scala
```diff
@@ -10,7 +10,7 @@ object Versions {
   val bloopVersion = "1.4.0-RC1-94-a2be783d"
   val zincVersion = "1.1.1"
   val intellijVersion = "201.6487.11"
-  val bspVersion = "2.0.0-M7"
+  val bspVersion = "2.0.0-M7+6-d2366ddb-SNAPSHOT"
```
jastice: waiting for release

jastice (changes requested):
LGTM! Just waiting for the bsp4j release

bsp/resources/messages/ScalaBspBundle.properties
lukaszwawrzyk: actually we didn't add new endpoint so I should just delete it

PR #516, opened 2020-03-26
Conditional dependency on JUnit

tpasternak:
Before this commit, the Scala Plugin was unconditionally dependent on
JUnit plugin (it was referencing JUnitConfiguration class). This could
cause some problems, for example when JUnit plugin was disabled.

This commit makes this dependency conditional. Functions referencing
JUnitConfiguration has been moved to `runConfigurationClassExtractor`
extension, that is loaded only when JUnit plugin is enabled.

Addresses https://youtrack.jetbrains.com/issue/SCL-17209

### DIFF
--- a/bsp/resources/META-INF/BSP-JUnit.xml
+++ b/bsp/resources/META-INF/BSP-JUnit.xml
@@ -1,0 +1,6 @@
+
+<idea-plugin>
+    <extensions defaultExtensionNs="com.intellij">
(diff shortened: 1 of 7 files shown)


jastice (approved):

jastice:
Thanks Manually merged: 9859d293c573c08e21cdbfb10b3883fa7236575b

PR #517, opened 2020-03-27
Workaround for lsp4j cancellation problem

tpasternak:
This is only a POC, the right fix should probably be done in lsp4j.

lsp4j's cancellation is implemented in that way, the
`CompletableFuture` returned from the client library contains
overriden `cancel` function:
https://github.com/eclipse/lsp4j/blob/ad99a3d77139351dbc00c2dacd388fe749652309/org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/RemoteEndpoint.java#L149-L152

Unfortunately, if a CompletableFuture is mapped with `applyThen`, the
information about overriding is lost, so cancel must be called on the
original future.

In BSP-related code we modify the future a few times, so as long as
it's not fixed in intellij, we have to store the original future and
call cancel on it, instead of on derived one.

### DIFF
--- a/bsp/src/org/jetbrains/bsp/project/BspTask.scala
+++ b/bsp/src/org/jetbrains/bsp/project/BspTask.scala
@@ -76,7 +76,7 @@ class BspTask[T](project: Project,
   }
 
   override def onCancel(): Unit = {
-    resultPromise.failure(new ProcessCanceledException())
+    resultPromise.tryFailure(new ProcessCanceledException())
   }
 
   override def run(indicator: ProgressIndicator): Unit = {
@@ -132,7 +132,7 @@ class BspTask[T](project: Project,
     }
     else reporter.finish(combinedMessages)
 
-    resultPromise.success(combinedMessages)
+    resultPromise.trySuccess(combinedMessages)
   }
 
   private def messagesWithStatus(reporter: BuildReporter,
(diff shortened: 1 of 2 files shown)


jastice:
Oooh good find!

bsp/src/org/jetbrains/bsp/project/BspTask.scala
```diff
@@ -76,7 +76,7 @@ class BspTask[T](project: Project,
   }
 
   override def onCancel(): Unit = {
-    resultPromise.failure(new ProcessCanceledException())
+    resultPromise.tryFailure(new ProcessCanceledException())
```
tpasternak, line 79: @jastice do we need `onCancel` handler at all?

jastice: I think so, yes. Why do you think not?

tpasternak: I noticed the resultPomise is already completed in `BuildToolWindowReporter.actionPerformed`

tpasternak: And then the `BspTask.run` completes it as well. Is it possible it's not complited at `onCancel` stage?

jastice (approved):

PR #518, opened 2020-03-29
Position BSP environment classpath items before all other

tpasternak:
When we fetch classpath items via bspTestEnvironment endpoint, we
concatenate them to the list that is already set by other IntelliJ
components. We don't override it, because they may contain some
extra classes that are required to correcly communicate with
the UI. Nevertheless, the items from BSP are more important and they
should have higher priority (co they need to be placed before all
other ones)


tpasternak:
Btw, this was my initial intention, but  I lost this `clear` call at some point during refactoring :(

jastice (approved):

PR #519, opened 2020-03-30
Update bsp connection last activity time on client api calls

lukaszwawrzyk:
The actual problem that I fix here is that on request to workspace/buildTargets, refreshProjectsCommand can be potentially long, say 15 minutes.
Currently there is a 10 min timeout in idle connections that was preventing refresh from completing.
The lastActivity time used to detect idle connection was updated on stderr and when request is sent to the bsp server. This PR also updates activity time when server calls the client.
In case of refreshProjectsCommand, logs that come from it are sent as log events, hence this fixes the issue.


jastice (approved):
Thanks!

PR #523, opened 2020-04-04
Bsp server proxy

tpasternak:
Another approach to workaround lsp4j cancellation issue. This includes:
1. New, more generic, reflection-based solution
2. Reversion commit of the previous fix (as not needed any more)
3. "Promise already completed" fix that appeared after the changes above

### DIFF
--- a/bsp/src/org/jetbrains/bsp/protocol/session/CancellableFuture.scala
+++ b/bsp/src/org/jetbrains/bsp/protocol/session/CancellableFuture.scala
@@ -3,8 +3,8 @@ package org.jetbrains.bsp.protocol.session
 import java.util.concurrent.CompletableFuture
 
 /**
- * JDK's CompletableFutre does not handle cancellation well.
- * When a future does have an overriden `cancel` method, it is lost
+ * JDK's CompletableFuture does not handle cancellation well.
+ * When a future overrides a `cancel` method, it is lost
  * when doing transformations like `thenApply` or `thenCompose`.
  *
  * In Java 9, the new CompletableFuture's method was introduced: `newIncompleteFuture`.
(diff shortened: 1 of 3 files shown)


mzarnowski:
Although I am not a fan of reflection, I think it is the best option here - especially since the `launcher` already heavily depends on reflection.

bsp/src/org/jetbrains/bsp/protocol/session/CancellableFuture.scala
```diff
@@ -0,0 +1,44 @@
...
+object CancellableFuture {
+  def from[T](original: CompletableFuture[T]): CancellableFuture[T] = {
+    val result = new CancellableFuture[T](original)
+    original.whenComplete{ (value, error) =>
+      if(value != null) {
+        result.complete(value)
+      }
+      else {
+        result.completeExceptionally(error)
+      }
```
tpasternak: @marek1840 I'm not sure if this is correct. Maybe it should be like this?

```scala
      if(error != null) {
        result.completeExceptionally(error)
      }
      else {
        result.complete(value)
      }
```

tpasternak: Ok, changed this, as null exception cannot be thrown https://stackoverflow.com/questions/17576922/why-can-i-throw-null-in-java

bsp/src/org/jetbrains/bsp/protocol/session/CancellableFuture.scala
```diff
@@ -0,0 +1,44 @@
+package org.jetbrains.bsp.protocol.session
+
+import java.util.concurrent.CompletableFuture
+
+/**
+ * JDK's CompletableFutre does not handle cancellation well.
```
jastice: typo: Futre

jastice (approved):

jastice:
rebased and merged: ee5bf9e296b06fdedaae316dc63e6782c35e0f00 ff

PR #525, opened 2020-04-07
$outer fields may be inherited

retronym:
Scalac re-uses the outer field of a base class. Use `visibleFields` to make `this` evalution consider such inherited fields.

(First part of a few debugger improvements (mostly for scala-async) that I'm working towards in https://github.com/retronym/intellij-scala/pull/1)

### DIFF
--- a/scala/scala-impl/test/org/jetbrains/plugins/scala/debugger/evaluateExpression/ScalaThisAndSuperEvaluationTest.scala
+++ b/scala/scala-impl/test/org/jetbrains/plugins/scala/debugger/evaluateExpression/ScalaThisAndSuperEvaluationTest.scala
@@ -254,4 +254,34 @@ class ScalaThisAndSuperEvaluationTestBase extends ScalaDebuggerTestCase {
       evalEquals("foo", "1")
     }
   }
+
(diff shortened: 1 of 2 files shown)


scala/scala-impl/test/org/jetbrains/plugins/scala/debugger/evaluateExpression/ScalaThisAndSuperEvaluationTest.scala
```diff
@@ -254,4 +254,34 @@ class ScalaThisAndSuperEvaluationTestBase extends ScalaDebuggerTestCase {
...
+       |  }
+       |}
+      """.stripMargin.trim()
+  )
+  def testInnerOuterInheritedOuterFieldEtc(): Unit = {
+    runDebugger() {
+      waitForBreakpoint()
+      evalEquals("foo", "1")
+    }
+  }
```
retronym, line 286: Was:

image

niktrop:
Thank you for the fix! Merged manually: 2c8672f10cc925cfa97846ac88580c45dab85301

PR #526, opened 2020-04-07
Avoid watching .bloop and .bsp directory for project change

lukaszwawrzyk:
When these files are watched it causes automatic project refresh on each startup of Intellij. Also always brings in a floating refresh project button that is confusing.


jastice (changes requested):
as discussed in chat, let's make this behavior depend on a flag in the `.bsp/tool.json`

jastice (approved):

PR #528, opened 2020-04-14
Support target selection for worksheets in bsp projects

lukaszwawrzyk:
Worksheets are compiled using compile server with zinc. The classpath passed to compiler is inferred from module selected for given worksheet. In case of bsp single intellij module can contain multiple different build targets. Currently one of the output directories is selected (selected during creating synthetic module while importing such modules with multiple targets for one content root). 
This PR adds extension point to worksheets so that classpath can be provided in alternative way. It also adds implementation for this EP that fetches classpath from BSP. In case of ambiguous targets, it prompts user to choose it and saves the choice. It also adds a new button at top bar for worksheet that allows to change selected target at any time.
As there is quite some common code with logic for fetching classpath and other data for running tests and application, I refactored that logic to a common place (BspJvmEnvironment object).


jastice (approved):

bsp/src/org/jetbrains/bsp/project/test/environment/BspJvmEnvironment.scala
```diff
@@ -0,0 +1,256 @@
...
+  def promptUserToSelectBspTargetForWorksheet(module: Module): Unit = {
+    val potentialTargets = getBspTargets(module)
+    potentialTargets.foreach { targetIds =>
+      val holder = persistentHolderForWorksheet(module)
+      promptUserToSelectBspTarget(module.getProject, targetIds, holder)
+    }
+  }
+
+  def resolveForWorksheet(module: Module): Result[JvmEnvironment] = {
+    val holder = persistentHolderForWorksheet(module)
```
tpasternak, line 82: Does it mean that it is impossible to have two worksheets of the same module, but using different targets?

lukaszwawrzyk:
I just rebased, updating java.bean.BeanProperty to scala.bean.BeanProperty

bsp/src/org/jetbrains/bsp/project/test/environment/BspJvmEnvironment.scala
lukaszwawrzyk: Yeah, if someone has 2 worksheets that use module that has multiple targets in it, they will share the same target. If someone was switching between those, would need to change the target before run. But I hope that for now it's not a common use case.

jastice:
Thanks, we'll give this a closer look next week!

jastice (approved):

lukaszwawrzyk:
@unkarjedy Could you take a look at this PR?

unkarjedy:
@lukaszwawrzyk 
It's not forgotten =)
I had a look last week but I wanted to play with your changes locally, to fully understand how it looks like (I never used BSP projects before). Right now there are some issues with BSP for Windows, which block this and which are already being fixed by @jastice.
I hope we will manage to finish it this week.

unkarjedy:
I would say that it would be better not to show (even add) the icon for non-BSP projects at all, not just disable it.
image

bsp/src/org/jetbrains/bsp/project/test/environment/BspJvmEnvironment.scala
```diff
@@ -0,0 +1,256 @@
...
+    for {
+      workspace <- workspaceUri(module)
+      potentialTargets <- getBspTargets(module)
```
unkarjedy, line 85: Maybe lets filter only JVM targets, cause only they can be properly used in worksheets anyway.
This way we will not have to choose when it's not necessary:
image

unkarjedy: So, after leaving just JVM targets there will be practically available only main/test targets choice, am I right?
Do you have a use-case when you use test target in worksheets?

bsp/src/org/jetbrains/bsp/project/test/environment/ConfigureBspTargetForWorksheet.scala
```diff
@@ -0,0 +1,38 @@
...
+import org.jetbrains.plugins.scala.worksheet.settings.WorksheetFileSettings
+
+class ConfigureBspTargetForWorksheet extends AnAction with TopComponentAction {
+  override def genericText: String = BspBundle.message("bsp.task.choose.target.title")
+
+  override def actionIcon: Icon = Icons.BSP_TARGET
+
+  override def acceptFile(file: ScalaFile): Boolean = {
+    val module = findModule(file)
+    BspUtil.isBspModule(module)
```
unkarjedy, line 24: I would say that it would be better not to show (even add) the icon for non-BSP projects at all, not just disable it.
image

bsp/src/org/jetbrains/bsp/project/test/environment/BspJvmEnvironment.scala
lukaszwawrzyk: I will see if I actually can filter it JVM/other. 

The thing with targets is that they can be defined even per each source. So you may have 10 scala files in one package, which in intellij will be mapped to one module, but each of this file can be a different bsp target and have different classpath. 
You can find example of such project here:

https://github.com/olafurpg/bsp-pants/blob/scratchfile/scratchfile/src/test/scala/foobar/BUILD

bsp/src/org/jetbrains/bsp/project/test/environment/ConfigureBspTargetForWorksheet.scala
lukaszwawrzyk: Avoid adding is dificult, as there is no access to `project` during init of the component.  I updated PR to hide it on non BSP project.

bsp/src/org/jetbrains/bsp/project/test/environment/BspJvmEnvironment.scala
unkarjedy: ok, I see...

PR #529, opened 2020-04-15
Remove noisy library data message for bsp dependencies

lukaszwawrzyk:
Currently when looking for class there is this noisy `bsp: bsp test dependencies (<jar name>)` text near each class. This PR changes the extra text to empty string which results in simply `<jar name>` being shown without any extra noise.

Screenshot 2020-04-06 at 17 55 29

### DIFF
--- a/bsp/resources/META-INF/BSP.xml
+++ b/bsp/resources/META-INF/BSP.xml
@@ -4,9 +4,11 @@
     <depends optional="true" config-file="BSP-JUnit.xml">JUnit</depends>
     <extensionPoints>
         <extensionPoint qualifiedName="com.intellij.bspEnvironmentRunnerExtension"
-                        interface="org.jetbrains.bsp.project.test.environment.BspEnvironmentRunnerExtension"/>
+                        interface="org.jetbrains.bsp.project.test.environment.BspEnvironmentRunnerExtension"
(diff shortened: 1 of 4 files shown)


jastice:
As discussed in chat, we don't want this information to be removed in general. But we should consider how to display long module and jar names better in this situation.

bsp/resources/META-INF/BSP.xml
```diff
@@ -5,6 +5,8 @@
     <extensionPoints>
         <extensionPoint qualifiedName="com.intellij.bspEnvironmentRunnerExtension"
                         interface="org.jetbrains.bsp.project.test.environment.BspEnvironmentRunnerExtension"/>
+        <extensionPoint qualifiedName="com.intellij.bspResolverNamingExtension"
```
jastice, line 9: Going forward, all extension points should be declared dynamic so that the Scala plugin can become dynamic.

See https://www.jetbrains.org/intellij/sdk/docs/basics/plugin_structure/plugin_extension_points.html#dynamic-
extension-points

https://youtrack.jetbrains.com/issue/SCL-17035

jastice (changes requested):

jastice (approved):

PR #532, opened 2020-04-23
Detect VCS root on bsp project import

lukaszwawrzyk:
Fastpass creates bsp project root in different directory than repo root, hence repository auto detection mechanism fails to find actual repository (it detects based on project.basePath).

I am not sure if this is the correct place to do it (any suggestions welcome). I implemented logic to select a non root module as base for vcs root lookup, and then later I use auto detect mechanism on it and add found root.

### DIFF
--- a/bsp/src/org/jetbrains/bsp/data/BspProjectDataService.scala
+++ b/bsp/src/org/jetbrains/bsp/data/BspProjectDataService.scala
@@ -1,5 +1,8 @@
 package org.jetbrains.bsp.data
 
+import java.io.File
+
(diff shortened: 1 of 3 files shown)


jastice:
Thanks. Can you give an example where the current root detection fails?

olafurpg (commented):
Would it be possible to declare the VCS root in `.bsp/*.json`? In our use-case, it's common to have multiple open IntelliJ projects for the same Pants build so it's not possible for us to generate `.bsp/bloop.json` in the VCS root.

jastice (approved):

PR #534, opened 2020-05-14
BSP: Fix cross modules

joan38:
This is an attempt to fix https://youtrack.jetbrains.com/issue/SCL-17551

Trying this out on https://github.com/joan38/kubernetes-client:
Before this change:
Before project structure
Before CustomResource

With this change:
After project structure
After CustomResource

This is still a WIP and seem to mostly remove the synthetic modules which leads me to question **what else is this breaking?**

A remaining point is the naming of the module suffixed `-2.12.11` whereas it's both 2.12 and 2.13. Also the cmd+click on imports has a naming issue but looks right:
Screen Shot 2020-05-14 at 00 06 52
But before I move forward with fixing this I wanted to check with people here on the previous question.

Thanks

### DIFF
--- a/bsp/src/org/jetbrains/bsp/project/resolver/BspResolverLogic.scala
+++ b/bsp/src/org/jetbrains/bsp/project/resolver/BspResolverLogic.scala
@@ -2,9 +2,8 @@ package org.jetbrains.bsp.project.resolver
 
 import java.io.File
 import java.net.URI
-import java.nio.file.{Path, Paths}
+import java.nio.file.{Files, Paths, StandardOpenOption}
 import java.util.Collections
-
 import ch.epfl.scala.bsp4j._
 import com.google.gson.{Gson, JsonElement}
 import com.intellij.openapi.externalSystem.model.project._
@@ -20,7 +19,6 @@ import org.jetbrains.bsp.project.resolver.BspResolverDescriptors._
 import org.jetbrains.bsp.{BSP, BspBundle, BspErrorMessage}
 import org.jetbrains.plugins.scala.project.Version
 import org.jetbrains.plugins.scala.project.external.{JdkByHome, JdkByVersion}
-
 import scala.collection.JavaConverters._
 import scala.collection.mutable
 
@@ -34,22 +32,6 @@ private[resolver] object BspResolverLogic {
 
   private def extractSbtData(data: JsonElement)(implicit gson: Gson): Option[SbtBuildTarget] =
     Option(gson.fromJson[SbtBuildTarget](data, classOf[SbtBuildTarget]))
-
-  /** Find common base path of all given files */
-  private[resolver] def commonBase(dirs: Iterable[File]): Option[File] = {
-    val paths = dirs.map(_.toPath)
-    if (paths.isEmpty) None
-    else {
-      val basePath = paths.foldLeft(paths.head) { case (common, it) =>
-        common.iterator().asScala.zip(it.iterator().asScala)
-          .takeWhile { case (c, p) => c == p }
-          .map(_._1)
-          .foldLeft(paths.head.getRoot) { case (base, child) => base.resolve(child) }
-      }
-
-      Some(basePath.toFile)
-    }
-  }
 
   private[resolver] def getJdkData(target: JvmBuildTarget): JdkData = {
     JdkData(new URI(target.getJavaHome), target.getJavaVersion)
@@ -86,7 +68,6 @@ private[resolver] object BspResolverLogic {
                                                     resourcesItems: Seq[ResourcesItem],
                                                     dependencySourcesItems: Seq[DependencySourcesItem]): ProjectModules = {
 
-
     val idToTarget = buildTargets.map(t => (t.getId, t)).toMap
     val idToScalacOptions = scalacOptionsItems.map(item => (item.getTarget, item)).toMap
 
@@ -115,16 +96,19 @@ private[resolver] object BspResolverLogic {
       .map(item => (item.getTarget, sourceDirectories(item)))
       .toMap
 
-    val sharedSources = sharedSourceDirs(idToSources)
+    val sharedResources = sharedSourceDirs(idToResources)
+    val sharedSources = sharedSourceDirs(idToSources.mapValues(_.filterNot(_.generated)))
+    val sharedGeneratedSources = idToSources
+      .mapValues(_.filter(_.generated))
+      .filter { case (id, src) => sharedSources.values.flatten.toSeq.contains(id) && src.nonEmpty }
 
     val moduleDescriptions = buildTargets.flatMap { target: BuildTarget =>
       val id = target.getId
       val scalacOptions = idToScalacOptions.get(id)
-      val depSources = idToDepSources.getOrElse(id, List.empty)
-      val sources = idToSources
-        .getOrElse(id, Seq.empty)
-        .filterNot(sharedSources.contains)
-      val resources = idToResources.getOrElse(id, List.empty)
+      val depSources = idToDepSources.getOrElse(id, Seq.empty)
+      val sharedSourcesAndGenerated = (sharedSources.keys ++ sharedGeneratedSources.keys).toSeq
+      val sources = idToSources.getOrElse(id, Seq.empty).filterNot(sharedSourcesAndGenerated.contains)
+      val resources = idToResources.getOrElse(id, Seq.empty).filterNot(sharedResources.contains)
       val dependencyOutputs = transitiveDependencyOutputs(target)
 
       implicit val gson: Gson = new Gson()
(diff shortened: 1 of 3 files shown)


jastice:
Synthetic modules are a workaround to support shared sources to some degree. So removing them would break any project setup that depends on them in one way or another.

I tried reproducing the problem, but mill-bsp didn't generate the Mill classpath in the project for some reason

joan38:
@jastice I forgot to put the `-i` in the reproduce steps. So instead of `mill mill.contrib.BSP/install` it should be `./mill -i mill.contrib.BSP/install`

Do you have another example of project that uses Synthetic modules?

jastice:
Indeed, I forgot about that bit. However I'm still having problems importing, I'll need to investigate.

Any sbt project that uses shared sources will generate shared synthetic modules as well, also when imported through Bloop/BSP. For example cats. It's also possible that the Mill-BSP implementation has some problem reporting these correctly.

jastice:
There seems to be a problem with the mill-bsp implementation that causes it to break completely for the kubernetes-client project, at least on my side. Importing Mill itself over BSP works for me. Does the regular import of kubernetes-client run without errors for you, @joan38 ?

joan38:
@jastice Importing Mill itself is working fine because it is not cross compiling between 2 Scala version (i.e it doesn't have `main[2.12.11]` and `main[2.13.2]`).

I have looked at the `bsp.log` and to me it's not a Mill bug since it is reporting a target for each Scala version to the same paths.

joan38:
So here is what's going on:
If you have:
- `src`
- `src-2.12`
- `src-2.13`
- `out/2.13.2/.../generatedSources/...`
- `out/2.12.11/.../generatedSources/...`

It will create:
- A project for `src-2.12` and `out/2.13.2/.../generatedSources/...`
- Another one for `src-2.13` and `out/2.12.11/.../generatedSources/...`
- And 1 for `src` and make the 2 others depend on it.

The issue is that the code in `src` is not seeing the generated sources since the dependency is the other way around. Thus making it all red.

joan38:
Alright the last commit fixes that by adding the generated files to the shared directory.

I also worked on a better (in my opinion) synthetic modules naming. It identifys common parts of the display name and replaces the chars by `_`s:
*cats*:
Screen Shot 2020-05-18 at 01 49 46
*kubernetes-scala*:
Screen Shot 2020-05-18 at 01 50 07
I think it's better than the path naming with `/` replaced by `_`s. And it was confusing to see `kubernetes_src` for the resource synthetic module.

@jastice please have a try and let me know what you think.
Please note that this was tried in conjonction with https://github.com/lihaoyi/mill/pull/885 for Mill.

jastice:
Thank you for this! Good idea with the naming. I will review and verify the changes this week.

joan38:
So taking my previous example:
If you have:
- `src`
- `src-2.12`
- `src-2.13`
- `out/2.13.2/.../generatedSources/...`
- `out/2.12.11/.../generatedSources/...`

It will create:
- A project for `src-2.12`.
- Another one for `src-2.13`.
- And 1 for `src`, `out/2.13.2/.../generatedSources/...`, `out/2.12.11/.../generatedSources/...` and make the 2 others depend on it.

jastice:
the naming strategy looks a bit awkward in zio:

image

joan38:
True, there is something weird going on here.
Thanks for testing this. I'll debug it and find what's wrong.

joan38:
@jastice I pushed a different version that yields:
Screen Shot 2020-05-19 at 01 33 05
Screen Shot 2020-05-19 at 01 35 11
Screen Shot 2020-05-19 at 02 56 26

Let me know what you think.

bsp/src/org/jetbrains/bsp/project/resolver/BspResolverLogic.scala
```diff
@@ -220,8 +210,7 @@ private[resolver] object BspResolverLogic {
     val sourceRoots = filterRoots(sourceDirs)
     val resourceRoots = filterRoots(resourceDirs)
 
-    val moduleBase = Option(target.getBaseDirectory)
-      .map(_.toURI.toFile)
+    val moduleBase = Option(new File(target.getId.getUri.toURI.getPath))
```
jastice: use `URI.toFile` extension method instead to avoid problems with windows paths

bsp/src/org/jetbrains/bsp/project/resolver/BspResolverLogic.scala
```diff
@@ -318,31 +307,42 @@ private[resolver] object BspResolverLogic {
     data
   }
 
+  private[resolver] def sharedModuleId(targets: Seq[BuildTarget]): String = {
```
jastice, line 314: can you add an explanation or better unit test for what this function does? In particular the regex

jastice (changes requested):
I really like the improved shared module naming scheme! If you can address the minor remaining things I think it's good to go!

joan38:
Done @jastice

joan38:
@jastice Can you advice on why the tests are failing?

joan38:
@jastice I fixed the build. Should we move forward with this?

bsp/test/org/jetbrains/bsp/project/resolver/BspResolverLogicTest.scala
```diff
@@ -12,7 +12,7 @@ class BspResolverLogicTest {
   @Test def testCalculateModuleDescriptionsEmptyBaseDir(): Unit = {
 
     val target = new BuildTarget(
-      new BuildTargetIdentifier("ePzqj://jqke:540/n/ius7/jDa/t/z78"),
+      new BuildTargetIdentifier("file:///jqke:540/n/ius7/jDa/t/z78"),
```
jastice: does this break now with non-`file` URIs? technically any valid URI should be a valid identifier

bsp/src/org/jetbrains/bsp/project/resolver/BspResolverLogic.scala
```diff
@@ -220,8 +213,7 @@ private[resolver] object BspResolverLogic {
     val sourceRoots = filterRoots(sourceDirs)
     val resourceRoots = filterRoots(resourceDirs)
 
-    val moduleBase = Option(target.getBaseDirectory)
-      .map(_.toURI.toFile)
+    val moduleBase = Option(target.getId.getUri.toURI.toFile)
```
jastice: Actually this fails when the `id` URI is not a file, which is valid in BSP. Why did you change it to this?

jastice (changes requested):
Found the change to `moduleBase` actually causes errors. I'd just revert it but maybe you had a reason for it?

joan38:
@jastice the `commonPath` function is not used anymore. I removed it all together.
This is ready for rereview.

jastice:
thanks, but that still leaves the issue with `moduleBase` trying to turn a target id into a file rather than using `getBaseDirectory`

joan38:
Fixed using URI

bsp/src/org/jetbrains/bsp/project/resolver/BspResolverLogic.scala
```diff
@@ -288,7 +270,7 @@ private[resolver] object BspResolverLogic {
       moduleName,
       targets,
       Seq.empty, Seq.empty,
-      moduleBase,
+      moduleBase.map(_.toFile),
```
jastice: This still won't work because you're still passing in the `id` URI, which may or may not correspond to a base directory

joan38: Why did I touch `moduleBase`?
I pushed again reverting this part.

joan38: Wait something is still wrong

bsp/test/org/jetbrains/bsp/project/resolver/BspResolverLogicTest.scala
jastice: This test should work again in the original form

bsp/src/org/jetbrains/bsp/project/resolver/BspResolverLogic.scala
joan38: Using the `getBaseDirectory` instead of the `getId` for `moduleBase` won't work for Mill because both 2.12 or 2.13 for example points to the same base dir.

joan38: SBT is not impacted for now because it gives only one `scalaVersion` not all defined in `crossScalaVersion`

jastice: So the question is: why do you need the `basePath` to be different? Are you trying to avoid merging them into the same IntelliJ module? In that case we can consider not merging modules by path, this might be preferable anyway, because test and compile modules may have some separate configurations anyway.

Another option is reporting different base paths from Mill's side.

joan38: You need to end up somehow with 1 module that points to `src`, another 1 pointing to `src-2.12` and another 1 pointing to `src-2.13`.
Different base path from Mill's side is not ok because that is not consistant with not cross compiled modules.

jastice: I see. I want to investigate this more closely, but I'm still having problems importing kubernetes-client with mill-bsp, or even building it at all with vanilla mill. Maybe we can have a chat and figure that out?

joan38: That would be great, I sent you an email.

jastice (approved):

PR #535, opened 2020-05-20
Fix link to Scalastyle website

loganrosen:
The inspection description for Scalastyle links to https://www.scalastyle.org/, which yields a security warning because the certificate is only valid for (*.)nfshost.com at the time of writing this issue. I'm not sure if it was ever valid — Google indexes only the HTTP version. We should link to http://www.scalastyle.org/ instead, at least until the Scalastyle maintainers fix their certificate.

#SCL-17594 fixed


jastice:
thanks for the fix!

PR #536, opened 2020-05-21
Automatically import Mill project with BSP

joan38:


jastice:
Cool! But will it work without adding the BSP module explicitly to build.sc?

joan38:
@jastice Now it's checking that the plugin is enabled.
I've also implemented the cancel mechanism.

jastice:
Great! do you think it's possible to inject the plugin, like with sbt?

scala/scala-impl/src/org/jetbrains/sbt/project/MillProjectImportProvider.scala
```diff
@@ -0,0 +1,23 @@
...
+  private val buildFile = "build.sc"
+
+  def canImport(vFile: VirtualFile): Boolean = vFile match {
+    case null => false
+    case directory if directory.isDirectory =>
+      val buildWithBspPlugin = directory.getChildren.exists { vBuildScript =>
+        vBuildScript.getName == "build.sc" &&
+          Source.fromInputStream(vBuildScript.getInputStream)
+            .getLines()
+            .contains("import $ivy.`com.lihaoyi::mill-contrib-bsp:$MILL_VERSION`")
```
jastice, line 17: Would it make sense to not make this detection dependent on this specific line to be in the build.sc? Are there situations where one would want to put a specific mill version instead?

scala/scala-impl/src/org/jetbrains/sbt/project/structure/MillProjectDumper.scala
```diff
@@ -0,0 +1,29 @@
...
+import java.io.File
+import java.util.concurrent.atomic.AtomicBoolean
+import scala.concurrent.duration._
+import scala.collection.JavaConverters._
+
+object MillProjectDumper {
+  private val MillProcessCheckTimeout = 100.millis
+
+  def setupBsp(baseDir: File): MillProjectDumper = {
+    val processBuilder = new ProcessBuilder(Seq("./mill", "-i", "mill.contrib.BSP/install").asJava)
```
jastice: is the mill script available in every mill project, and will it work like this on Windows?

scala/scala-impl/src/org/jetbrains/sbt/project/structure/SbtStructureDump.scala
```diff
@@ -28,7 +28,7 @@ import scala.collection.JavaConverters._
 import scala.concurrent.Future
 import scala.util.{Failure, Success, Try}
 
-class SbtStructureDump {
+class SbtStructureDump extends ProjectDumper {
```
jastice: "Dumper" maybe a it of a misnomer for mill, perhaps call it something like "PreImport"?

jastice (commented):
LGTM, can you clear up the open questions?

scala/scala-impl/src/org/jetbrains/sbt/project/structure/MillProjectDumper.scala
joan38: No, ideally we would not need to rely on this and would have an equivalent of `sbt-launcher.jar` but I didn't have time to figure out how this whole thing works.
The `canImport` checks for that script but it's true that on windows that would be named `mill.bat` I think.

scala/scala-impl/src/org/jetbrains/sbt/project/MillProjectImportProvider.scala
joan38: I can inject the plugin, that would solve the issue.

jastice: Great, hope we can still get https://github.com/lihaoyi/mill/issues/899 sooner than later

scala/scala-impl/src/org/jetbrains/sbt/project/structure/MillProjectDumper.scala
jastice: I think it's fine if we can't cover all setups for now as long as it doesn't break in situations where assumptions don't hold

scala/scala-impl/src/org/jetbrains/sbt/project/MillProjectImportProvider.scala
joan38: Injecting it sounds not as easy as adding a file with the plugin and it feels like a waist of time to implement this since eventually we will get to https://github.com/lihaoyi/mill/issues/899.

scala/scala-impl/src/org/jetbrains/sbt/project/MillProjectImportProvider.scala
```diff
@@ -0,0 +1,23 @@
...
+      }
+      val bootstrap = directory.getChildren.exists(_.getName == "mill")
```
jastice: maybe check if it's executable?

jastice (approved):

jastice (approved):

scala/scala-impl/src/org/jetbrains/sbt/project/structure/MillPreImporter.scala
```diff
@@ -0,0 +1,29 @@
...
+import java.io.File
+import java.util.concurrent.atomic.AtomicBoolean
+import scala.concurrent.duration._
+import scala.collection.JavaConverters._
+
+object MillPreImporter {
+  private val MillProcessCheckTimeout = 100.millis
+
+  def setupBsp(baseDir: File): MillPreImporter = {
+    val processBuilder = new ProcessBuilder(Seq("./mill", "-i", "mill.contrib.BSP/install").asJava)
```
lefou, line 12: Does that mean, one has to provide a mill wrapper script with name `mill` in the project? I'd rather not enforce this.

jastice: With the current implementation, it's necessary for the automatic preimport. If the script is absent, it will not run and require manual invocation instead. We could avoid this by bundling a standalone Mill launcher, like we do with bloop or sbt, or if Mill can be fully bootstrapped from Coursier, which we also bundle. IMO doing it via Coursier would be my preferred option.

joan38: That was a quick win implementation. I think the coursier route sounds way better.

PR #540, opened 2020-06-04
Ensure that loop in `findInLibSources` is interruptible

tpasternak:
In large projects, the llop in `findInLibSources` function may be very
expensive. It's executed with read permissions, so if any write action
happens before it ends, it have to wait. This may lead to UI
freezes. The freezes usually don't occur, because some of the
functions that are called inside the loop, call `ReadAction.compute`
that throws an exception when there is a pending write action. One of
the examples is `MavenProjectsManager.isMavenizedModule`.

Unfortunately this function is not executed in all cases, so in order
to reduce probability of the freeze, it would be good to ensure the
`ReadAction.compute` is called in each iteration.


jastice:
Thanks! We just switched to the next release branch. Can you please rebase on `idea202.x`?

tpasternak:
Rebased onto newest 202.x

jastice (approved):

PR #541, opened 2020-06-04
Allow injection of custom resolvers into test framework.

DzmitryMikhalap:
Restore overwritten part of fb5dadb


jastice:
Thanks for the PR! We just switched the base branch to the next release. Can you please rebase onto idea202.x

PR #544, opened 2020-06-15
Add ScalaTest 3.1.0 fqns for generate method checks

a1kemist:
# Purpose
Adds the ScalaTest `3.1.0` fqns to the generate method checks so that test case generation works when creating a new test.

# Background
I noticed that when creating new ScalaTest tests, generating test cases for methods seems to be broken for ScalaTest version `3.1.0`. After inspecting the code, I noticed that the generator is still checking for the old fqns.


unkarjedy:
Thanks for the PR.
I've created a dedicated ticket for it: https://youtrack.jetbrains.com/issue/SCL-17723
I will add some modifications and merge it manually (we need to unify it with [ScalaTestUtil](https://github.com/JetBrains/intellij-scala/blob/a9b0f4659069347f0063a7ebf6862cff4b445bab/scala/scala-impl/src/org/jetbrains/plugins/scala/testingSupport/test/scalatest/ScalaTestUtil.scala))

unkarjedy:
The manual merge is on it's way )

PR #546, opened 2020-06-17
Optimize VCS root lookup in BSP

lukaszwawrzyk:
In case of project root is outside git repo, we look from vcs root from all modules. This uses IO and in case of many modules is very slow. This is follow up on https://github.com/JetBrains/intellij-scala/pull/532

The optimization here is not to call the detector if the module directory is already under detected root. So in case there is only one repository, we will just run the detector once.

The PR target is 201.x it would be ideal to have at least a nightly release of it. I can't PR against 202.x as  the base PR I mentioned earlier is still not merged to 202.x


olafurpg:
Several of our internal BSP users have reported that IntelliJ is hanging during the "importing BSP project data" step, which we have traced down to this fix. It would be great to have a fix available in the nightly build soon so we can roll it out to our users.

jastice (approved):

jastice:
Build running for this PR.
#545 now merged in 202.x

jastice:
Cherrypicking this to 202.x

jastice:
cherrypicked to 8b10201e6bae0fd5699848cb5f8af8b53f54a45b

PR #547, opened 2020-06-18
Gather logs from BSP refresh and include in the exception

tpasternak:
Improves user experience, previously the logs were hard to find in the
UI. The exception that was thrown previously was presented in "Build
-> Sync" window did not contain any details about the failure
reason. Although the focus was set to it, so the reason of the
failure was unclear. The user had to look for the logs in the root
job, which was unintuitive. No the logs are copied to the
Bsp4JJobFailure exception and presended in the focused view.

### DIFF
--- a/bsp/src/org/jetbrains/bsp/project/resolver/BspProjectResolver.scala
+++ b/bsp/src/org/jetbrains/bsp/project/resolver/BspProjectResolver.scala
@@ -6,7 +6,7 @@ import java.util.concurrent.CompletableFuture
 
 import ch.epfl.scala.bsp4j._
 import com.intellij.build.events.impl.SuccessResultImpl
-import com.intellij.openapi.externalSystem.model.DataNode
+import com.intellij.openapi.externalSystem.model.{DataNode, ExternalSystemException}
 import com.intellij.openapi.externalSystem.model.project._
 import com.intellij.openapi.externalSystem.model.task.{ExternalSystemTaskId, ExternalSystemTaskNotificationListener}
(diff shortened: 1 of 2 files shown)


bsp/src/org/jetbrains/bsp/project/resolver/BspProjectResolver.scala
```diff
@@ -113,11 +114,14 @@ class BspProjectResolver extends ExternalSystemProjectResolver[BspExecutionSetti
 
   private def runImport(workspace: File, executionSettings: BspExecutionSettings)
                        (implicit reporter: BuildReporter) = {
-    val notifications: NotificationCallback = {
+    def notifications(implicit reporter: BuildReporter): NotificationAggregator[BuildMessages] =
+    (messages, notification) => notification match {
       case BspNotifications.LogMessage(params) =>
-        // TODO use params.id for tree structure?
         reporter.log(params.getMessage)
+        messages.addWarning(params.getMessage)
```
tpasternak, line 121: @jastice I really don't like this code (adding a warning and then reading _.warnings) but I couldn't find any better way to utilize BuildMessages in this case. The messages are streamed line-by-line here from bloop.

jastice: how about adding a `log` field to `BuildMessages` instead?

jastice: Actually, I'll just do this change

jastice:
Manually merged/rebased with some adaptions via b9cdae8e0a856e3c514cfc7f5addaf8c8c50aa37

PR #549, opened 2020-06-24
Fix test sources for synthetic test modules

joan38:


jastice:
Thanks! What exactly does this fix?

joan38:
Screen Shot 2020-06-24 at 13 00 28
The shared `kubernetes-client/test/src` folder was not tagged as a test source and was not green. The code was also all red.

jastice (approved):

PR #550, opened 2020-06-26
More logs related to BSP execution environment

tpasternak:
I observed that the "Use BSP environment" task is sometimes not added to "Before run tasks" list for tests in BSP-based modules. The problem is hard to reproduce, and there are no logs related to it. 

Interesting fact that is it's something defined at startup - it either works correctly for the entire runtime, until IntelliJ is closed, or does not work at all. 

There are two hypothesis for this:
1. The `BspFetchTestEnvironmentTaskInstaller` is not subscribed to `RunManagerListener` topic
2. The `runConfigurationAdded` callback silently fails for some unknown reason

With these logs I hope I will be able to distinguish which of these two is true.


jastice:
cc @lukaszwawrzyk @tpasternak

jastice:
is this intended to be on `idea201.x` and not `idea202.x`?

tpasternak:
I'd love to have it in 201.x

jastice (approved):

PR #551, opened 2020-06-26
SCL-17561 scalastyle: switch to beautiful-scala fork

steved:
~~When scalastyle cannot parse the file (and there are a lot of issues around that https://github.com/scalastyle/scalastyle/issues/356 https://github.com/scalastyle/scalastyle/issues/191 https://github.com/scalastyle/scalastyle/issues/292 https://github.com/scalastyle/scalastyle/issues/187 https://github.com/scalastyle/scalastyle/issues/195 https://github.com/scalastyle/scalastyle/issues/319) it raises a `StyleException` instead of a `StyleError`. I'm unsure if not reporting them was a conscious choice, so I figured I'd raise a PR to see.~~

~~Although `StyleException` has a line number and column associated, I [could not find anywhere in the current](https://github.com/scalastyle/scalastyle/blob/5c4bb3e1d6b7666c4546bbbd10c4d503455bbc9e/src/main/scala/org/scalastyle/Checker.scala#L141) `scalastyle` code that uses it so did not include them in the match (see 60ae3a9).~~

`scalastyle`'s original repository has been unmaintained for >1 year. The [beautiful-scala](https://github.com/beautiful-scala/scalastyle) project [has taken up the mantle](https://github.com/scalastyle/scalastyle/issues/327#issuecomment-570951273) and updated dependencies and merged some much needed fixes.


scala/scala-impl/src/org/jetbrains/plugins/scala/codeInspection/scalastyle/ScalastyleCodeInspection.scala
```diff
@@ -53,9 +53,7 @@ object ScalastyleCodeInspection {
       possibleConfigFileNames.flatMap(name => Option(dir.findChild(name))).headOption
 
     def findIn(project: Project, possibleConfigFileNames: Seq[String]): Option[VirtualFile] = {
-      val root = project.getBaseDir
-      if (root == null) return None
-
+      val root = ProjectUtil.guessProjectDir(project)
```
steved: Slight drive-by refactor, but I had trouble testing this change with a different plugin and `getBaseDir` is deprecated.

scala/scala-impl/src/org/jetbrains/plugins/scala/codeInspection/scalastyle/ScalastyleCodeInspection.scala
```diff
@@ -89,6 +87,9 @@ object ScalastyleCodeInspection {
     private val maybeDocument = PsiDocumentManager.getInstance(file.getProject).getDocument(file).toOption
 
     def toProblemDescriptor(msg: Message[SourceSpec]): Option[ProblemDescriptor] = msg match {
+      case StyleException(_, _, message, _, None, None) =>
+        Some(manager.createProblemDescriptor(file, message, true, Array.empty[LocalQuickFix], ProblemHighlightType.GENERIC_ERROR))
```
steved: `GENERIC_ERROR` might be harsh since there are probably folks who don't run `scalastyle` as a blocker to code entry / passing tests.

scala/scala-impl/src/org/jetbrains/plugins/scala/codeInspection/scalastyle/ScalastyleCodeInspection.scala
SrTobi: This looks good, we'll keep that :+1:
But with the null-check... or was there a specific reason you removed that?

SrTobi:
Hi. Thanks for the PR. It looks good to me in principle, but I cannot merge it the way it currently is. Without the line and column info it is more annoying than helpful. Because putting an error on the file opens this big red banner. At least for me... is it different for you? Also intellij itself should produce an error at the correct position if the file is not parseable. so there should be already an error present and fixing that should also fix the scalastyle parsing.

Also the scalastyle project doesn't seem to be maintained anymore and as far as I can see it parses Scala 2.11 this leads to errors when using trailing commas for example:

image

Is there a version/fork that is commonly used in the community? Because then maybe we have to go to that version or maybe drop support altogether... or at least for versions >2.11

On the other hand the error message seems to have the index of the error in the message string... so you could try to parse out that number and annotate it there.

SrTobi:
Ok, it seems that we are going to migrate to a fork of scalastyle eventually. See https://youtrack.jetbrains.com/issue/SCL-17561

steved:
Sure, using the fork seems great. Any other considerations to be made with using [the beautiful-scala fork](https://github.com/scalastyle/scalastyle/issues/327#issuecomment-570951273)?

> Because putting an error on the file opens this big red banner. At least for me... is it different for you? Also intellij itself should produce an error at the correct position if the file is not parseable. so there should be already an error present and fixing that should also fix the scalastyle parsing.

Nope, this is what happened with me as well. Thanks for taking a look.

scala/scala-impl/src/org/jetbrains/plugins/scala/codeInspection/scalastyle/ScalastyleCodeInspection.scala
steved: Whoops, nope added back :)

SrTobi:
> Any other considerations to be made with using the beautiful-scala fork?

No looks good so far. It was our plan too.

@steved what problems did you encounter with `project.getBaseDir` btw? Because it fails for our test system and it seems that it is not what we want here, as it prioritizes module directories. I would like to change it back (or at least to something equivalent to the deprecated getBaseDir)

steved:
Oh hm, yeah I had trouble with the [`intellij/bazel`](https://github.com/bazelbuild/intellij) module and picking up the scalastyle config. Let me test a few things locally.

jastice:
Please check the failed test: https://github.com/JetBrains/intellij-scala/pull/551/checks#step:12:4853


```
[error] Test org.jetbrains.plugins.scala.codeInspection.scalastyle.ScalastyleTest.test failed: java.lang.AssertionError: Highlights not found: Class name does not match the regular expression '[A-Z][A-Za-z]*', took 0.253 sec
### scala: 2.10.7, jdk: JDK_11 ###
[error]     at org.jetbrains.plugins.scala.codeInspection.ScalaHighlightsTestBase.checkTextHasError(ScalaInspectionTestBase.scala:81)
[error]     at org.jetbrains.plugins.scala.codeInspection.ScalaHighlightsTestBase.checkTextHasError(ScalaInspectionTestBase.scala:67)
[error]     at org.jetbrains.plugins.scala.codeInspection.scalastyle.ScalastyleTest.test(ScalastyleTest.scala:55)
[error]     at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
[error]     at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
[error]     at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
[error]     at java.lang.reflect.Method.invoke(Method.java:498)
[error]     at junit.framework.TestCase.runTest(TestCase.java:154)
[error]     at com.intellij.testFramework.UsefulTestCase.lambda$runTest$9(UsefulTestCase.java:332)
[error]     at com.intellij.testFramework.UsefulTestCase.lambda$invokeTestRunnable$10(UsefulTestCase.java:369)
[error]     at com.intellij.testFramework.EdtTestUtilKt.runInEdtAndWait(EdtTestUtil.kt:44)
```

steved:
> Oh hm, yeah I had trouble with the intellij/bazel module and picking up the scalastyle config. Let me test a few things locally.

I think this'll have to be something I figure out with that project. It's picking up the scalastyle config from the [`project-data-dir`](https://github.com/bazelbuild/intellij/blob/72110e6bbbaab7645a664d47241cc4edfda8066d/base/src/com/google/idea/blaze/base/sync/ProjectUpdateSyncTask.java#L406) module instead of `.workspace` which has the actual project files. If you know off-hand how this should be solved, please let me know, otherwise I can dig into it separately.

SrTobi:
@steved not sure if I understand it correctly. what does intellij/bazel do here? The scala plugin scalastyle inspection looks only into the base directory where the `.idea` folder is located. into the `.idea` folder and into the `project` folder... this is also because I think scalastyle doesn't define a fix location where the config file should be, right?

So it is a little hard for us to correctly locate it... not without much more efford.

steved:
The actual `scalastyle` config lookup is ok. There intellij/bazel plugin creates two modules for the project, only one has the content of the actual project:
Screen Shot 2020-07-02 at 1 56 33 PM
Screen Shot 2020-07-02 at 1 57 33 PM

The `.idea` is nested in the first one, so `project.getBaseDir` returns `.ijwb` and there's no scalastyle config in there. We could look through all modules potentially? 
Something like (potentially crap code aside)
```scala
val roots = ModuleManager.getInstance(project).getModules.flatMap(ModuleRootManager.getInstance(_).getContentRoots)
roots.flatMap({ root =>
  val dirs = possibleLocations.flatMap(name => Option(root.findChild(name))) :+ root
  dirs.flatMap(findConfigFile(_, possibleConfigFileNames)).headOption
}).headOption
```

Also happy to take this into a new issue.

SrTobi:
Yeah, good idea (creating an issue + looking through all modules)

Do you know if it is common to have multiple config files? because we might also take into account the module in which the currently inspected file is.

Let's say you have a project like this:

```
/
  module1/
    src/...
    scalastyle-config.xml
  module2/
    src/...
build.sbt
```

In this case it wouldn't make sense to apply scalastyle to `module2` with the config from `module1`, right?

steved:
Yep, I think that makes a lot of sense! It also seems to fit the IntelliJ data model better as well. I created https://github.com/JetBrains/intellij-scala/pull/557 with a (potential) solution.

PR #555, opened 2020-07-01
Add button to terminate BSP Server

tpasternak:
The button calls command defined in `.bsp/bloop.json/argvExit`
field. Typically it's just `bloop exit`. If the field is empty,
a warning is logged.


tpasternak:
stop_bsp_server

jastice:
Usually the client sends a shutdown request and `build/exit` notification to the server on disconnect. Is this mechanism separate?

bsp/src/org/jetbrains/bsp/BspServerWidgetProvider.scala
```diff
@@ -110,6 +159,20 @@ class BspServerWidgetProvider extends StatusBarWidgetProvider {
   }
 
 
+  val argvExit = "argvExit"
+
+  private def bloopExitCommands(project: Project): List[Option[util.List[String]]] = {
```
jastice: name implies this is bloop-specific, but it looks at all connection files

bsp/src/org/jetbrains/bsp/BspServerWidgetProvider.scala
```diff
@@ -89,11 +98,51 @@ class BspServerWidgetProvider extends StatusBarWidgetProvider {
...
+      }
+
+      override def actionPerformed(e: AnActionEvent): Unit = try {
+        BspCommunicationService.getInstance.closeAll
+        ApplicationManager.getApplication.executeOnPooledThread(
+          new Runnable {
+            override def run(): Unit = {
+              val commands = bloopExitCommands(project).flatten
+              if (commands.isEmpty) {
+                logger.warn(s"BSP server exit command not set in '${argvExit}' field")
```
jastice: `argvExit` isn't defined in BSP specification and not required, shouldn't really be a warning.

bsp/src/org/jetbrains/bsp/BspServerWidgetProvider.scala
```diff
@@ -89,11 +98,51 @@ class BspServerWidgetProvider extends StatusBarWidgetProvider {
       }
     }
 
+    class RestartServer extends AnAction(BspBundle.message("bsp.widget.stop.all.bsp.restart"), BspBundle.message("bsp.widget.stop.all.bsp.restart"), AllIcons.Actions.Exit) with DumbAware {
```
jastice: action name implies restart, but it's just a stop action

bsp/src/org/jetbrains/bsp/BspServerWidgetProvider.scala
```diff
@@ -89,11 +98,51 @@ class BspServerWidgetProvider extends StatusBarWidgetProvider {
...
+          new Runnable {
+            override def run(): Unit = {
+              val commands = bloopExitCommands(project).flatten
+              if (commands.isEmpty) {
+                logger.warn(s"BSP server exit command not set in '${argvExit}' field")
+              } else {
+                logger.info(s"Running BSP server restart")
+              }
+              val timeoutLength = 30
+              commands.foreach { command =>
```
jastice, line 122: note this will stop all servers that have an `argvExit`, not just a specific one

tpasternak:
> Usually the client sends a shutdown request and `build/exit` notification to the server on disconnect. Is this mechanism separate?

We'd love to have some kind of last-chance recovery mechanism, for example for cases when connection hangs.

olafurpg:
To elaborate, this feature could be used as a fallback for users when the BSP server is not responding to `build/exit` notifications. Although our specific use-case is related to Bloop, there are other build servers that keep long running deamon processes that outlive `build/exit` notifications (for example Gradle, Mill, Bazel and likely soon sbt). I agree we should document this field in the BSP server discovery spec.

jastice:
Wouldn't the fallback in case of hanging build or failed `build/exit` be simply killing the bsp server process? But you're right this wouldn't apply to daemons that are separate from the server. So from that perspective this feature makes sense.

tpasternak:
I made this similarly to `X-detectExternalProjectFiles` field, but if you both think it should be included in protocol, then I'll of course add it

https://github.com/JetBrains/intellij-scala/blob/853e62fefd555dbeffd774be073e86f432fb76ed/bsp/src/org/jetbrains/bsp/project/BspExternalSystemManager.scala#L80

jastice:
I don't think it's strictly necessary to have it in the main protocol, I just think things that aren't required by the main protocol should be logging warnings.

tpasternak:
I force-pushed changes with suggestions applied and rebased onto updated idea202.x

bsp/src/org/jetbrains/bsp/BspServerWidgetProvider.scala
```diff
@@ -89,11 +98,54 @@ class BspServerWidgetProvider extends StatusBarWidgetProvider {
       }
     }
 
+    class TerminateServer extends AnAction(BspBundle.message("bsp.widget.stop.all.bsp.restart"), BspBundle.message("bsp.widget.stop.all.bsp.restart"), AllIcons.Actions.Exit) with DumbAware {
```
jastice: bundle key should reflect action name

bsp/src/org/jetbrains/bsp/BspServerWidgetProvider.scala
```diff
@@ -89,11 +98,54 @@ class BspServerWidgetProvider extends StatusBarWidgetProvider {
...
+
+      override def update(e: AnActionEvent): Unit = {
+        val workspace = new File(project.getBasePath)
+        val exitCommandDefined  = BspCommunicationService.getInstance.exitCommands(workspace.toURI).toOption.exists(_.nonEmpty)
```
jastice, line 105: won't this do I/O on every update?

tpasternak: No, because now the commands are fetched in `BspCommunication` constructor. I'm not fully sure it's valid but I could find any better way to implement `update` method. Please also note, I'm removing `BspConnection` from `conns` map on close.

bsp/src/org/jetbrains/bsp/BspServerWidgetProvider.scala
tpasternak: Done

tpasternak:
Reopened this PR here, as I had some problems with rebase https://github.com/JetBrains/intellij-scala/pull/560

PR #557, opened 2020-07-07
scalastyle: search for config file in same content root as file

steved:
The concept of a "project root" is deprecated [1], instead a project
consists of modules with content roots. This swaps the
scalastyle-config.xml lookup from the "project root" to find it within
the same content root as the file it's being applied to.

1: https://github.com/JetBrains/intellij-community/blob/3c162933038a472003a154f62ebc3c03895afc9b/platform/core-api/src/com/intellij/openapi/project/Project.java#L41


steved:
Well... this works great for the intellij-bazel integration, but I just tested a [multi-project](https://www.scala-sbt.org/1.x/docs/Multi-Project.html) sbt repo and it appears each module would expect to use the root "aggregate" module's scalastyle-config.xml. Might need to just keep the `project.getBasePath` check as a fallback?

SrTobi (approved):
looks good to me :+1:

SrTobi:
@steved thanks!

PR #559, opened 2020-07-13
Fix BSP jobs cancellation

tpasternak:
In order to support BSP jobs cancellation, we need `newIncompleteFuture` method of CompletableFuture (introduced in JVM9) to be properly implemented. This PR makes the code  working for JVM9 and newer, but also compilable with JVM8 (but without the actual effect).


jastice (approved):

tpasternak:
Rebased and reopened here https://github.com/JetBrains/intellij-scala/pull/561

PR #561, opened 2020-07-24
Fix BSP jobs cancellation

tpasternak:
In order to support BSP jobs cancellation, we need newIncompleteFuture method of CompletableFuture (introduced in JVM9) to be properly implemented. This PR makes the code working for JVM9 and newer, but also compilable with JVM8 (but without the actual effect).


tpasternak:
@jastice I noticed you've merged it manually, so I'm closing it 2f41ac9adc089b4013519c84d16aca6948c318af

PR #562, opened 2020-07-29
Preconfigure JDK before getting java executable

mzarnowski:
Currently, importing any sbt project would fail when there is no JDK defined within IntelliJ. This can happen for example if someone installs it for the first time and does not configure the JDK manually. 

Why does this happen?
Previously, there was a `DefaultJdkConfigurator` application component which tried to preconfigure a JDK during IDE startup. In [this commit](https://github.com/JetBrains/intellij-community/commit/31e4ac762bc422f9cb84d1fe33116a06e323a182), the component was removed and the `ProjectJdkTable::preconfigure` method was introduced. This method is actually being invoked automatically [in `ProjectSdksModel`](https://github.com/JetBrains/intellij-community/commit/31e4ac762bc422f9cb84d1fe33116a06e323a182#diff-bfd0b21274ac7c7d0d673c07ff619067R99) during the project import, but it happens much later than the `SbtExternalSystemManager` needs the JDK.

How does the PR work?
It preconfigures the JDK just before the JDK is needed. The operation is idempotent, so there is no worry that it will affect subsequent import operations.

Why is there no tests to cover this case?
Sadly, the `preconfigure` method is [ignored in unit test mode](https://github.com/JetBrains/intellij-community/commit/31e4ac762bc422f9cb84d1fe33116a06e323a182#diff-c655083450ea94cc9d970fe019b1f837R32). We are currently monitoring this behavior within our own scala [test suite](https://github.com/VirtusLab/ide-probe/blob/master/extensions/scala/tests/src/test/scala/org/virtuslab/intellij/extensions/ScalaImportTest.scala#L26) which is interacting with an actual IDE and not its unit test mode. In fact, that is how we found this issue.

### DIFF
--- a/scala/scala-impl/src/org/jetbrains/sbt/project/SbtExternalSystemManager.scala
+++ b/scala/scala-impl/src/org/jetbrains/sbt/project/SbtExternalSystemManager.scala
@@ -5,7 +5,7 @@ import java.io.File
 
 import com.intellij.diagnostic.PluginException
 import com.intellij.execution.configurations.SimpleJavaParameters
-import com.intellij.openapi.application.PathManager
+import com.intellij.openapi.application.{ApplicationManager, PathManager}
 import com.intellij.openapi.externalSystem.model.{ExternalSystemException, ProjectSystemId}
 import com.intellij.openapi.externalSystem.util._
 import com.intellij.openapi.externalSystem.{ExternalSystemConfigurableAware, ExternalSystemManager}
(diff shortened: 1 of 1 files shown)


jastice:
Thank you for this, I've been meaning to look into it!

jastice (approved):

jastice:
Manually merged via 484bf55b784791b036d361b58e67261ba2cdf56b

Thank you!

PR #563, opened 2020-08-03
Fix support of non sbt bsp projects

lukaszwawrzyk:
I added a naive guard to check if the project is sbt before attempting to run sbt bloopInstall. Checking for `bloopConfigDir` is not enough, as fastpass generates .bloop dir and the project itself is pants based and it needs fastpass to set it up. 
The PR also fixes missing jdk, by ensuring it is preconfigured before attempting to access it.

### DIFF
--- a/bsp/src/org/jetbrains/bsp/project/importing/BspProjectResolver.scala
+++ b/bsp/src/org/jetbrains/bsp/project/importing/BspProjectResolver.scala
@@ -154,17 +154,18 @@ class BspProjectResolver extends ExternalSystemProjectResolver[BspExecutionSetti
 
   // special handling for sbt projects: run bloopInstall first
   // TODO support other bloop-enabled build tools as well
-  private def preImport(executionSettings: BspExecutionSettings, workspace: File)(implicit reporter: BuildReporter) = {
+  private def preImport(executionSettings: BspExecutionSettings, workspace: File)(implicit reporter: BuildReporter): Try[BuildMessages] = {
     import BspProjectSettings._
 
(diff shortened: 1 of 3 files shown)


bsp/src/org/jetbrains/bsp/project/importing/BspProjectResolver.scala
```diff
@@ -154,12 +154,13 @@ class BspProjectResolver extends ExternalSystemProjectResolver[BspExecutionSetti
...
   // TODO support other bloop-enabled build tools as well
-  private def preImport(executionSettings: BspExecutionSettings, workspace: File)(implicit reporter: BuildReporter) = {
+  private def preImport(executionSettings: BspExecutionSettings, workspace: File)(implicit reporter: BuildReporter): Try[BuildMessages] = {
     import BspProjectSettings._
 
     val emptySuccess = Success(BuildMessages.empty.status(BuildMessages.OK))
+    def isSbtProject = new File(workspace, "build.sbt").exists()
 
-    if (executionSettings.runPreImportTask) {
+    if (executionSettings.runPreImportTask && isSbtProject) {
```
jastice: I think logically this belongs more under the if in the `AutoPreImport` case

lukaszwawrzyk: what about `BloopSbtPreImport`? Is it safe to omit it there?

jastice: That option *should* only be set for sbt projects so it should be safe to omit, and indeed it should be an error if it is set for non-sbt projects.

jastice (approved):

PR #564, opened 2020-08-21
Import BSP project via Fastpass

tpasternak:
This commit implements import of Pants projects to IntelliJ by converting
them to BSP projects with fastpass.

The tricky part was that, contrary to SBT-BSP imports, fastpass creates
BSP configuration *outside* of project root. This means, if the repository
root is `~/pants-example` and one wants to import `~/pants-example/java_app`,
the BSP configuration is not created in `~/pants-example/.bsp`, but in
`~/bsp-projects/java_app/.bsp`.

In order to support this case, the field `externalBspWorkspace` is introduced,
where BSP configuration is generated and stored, even if something else was chosen
by the user to import. Despite that, the BspSetupConfigStep must still receive
the original path selected by the user, because it needs to know what exactly is
supposed to be exported.

Thanks to this behavior, a user can import a few small, independent subprojects simultaneously
without touching rest of the project.

References:
https://github.com/scalameta/fastpass


bsp/src/org/jetbrains/bsp/project/importing/setup/FastpassConfigSetup.scala
```diff
@@ -0,0 +1,75 @@
...
+  private val cancellationFlag: AtomicBoolean = new AtomicBoolean(false)
+
+  private def waitFinish(process: Process): Try[BuildMessages] = {
+    while(!process.waitFor(FastpassProcessCheckTimeout.length, FastpassProcessCheckTimeout.unit)){
+      if(cancellationFlag.get()){
+        process.destroy()
+      }
+    }
+
+    // TODO build messages
```
tpasternak: @jastice I'm not sure how I should handle build messages here

bsp/src/org/jetbrains/bsp/settings/settings.scala
```diff
@@ -11,13 +11,16 @@ import com.intellij.openapi.externalSystem.settings._
...
 import com.intellij.openapi.project.Project
+import com.intellij.openapi.vfs.LocalFileSystem
 import com.intellij.util.messages.Topic
 import com.intellij.util.xmlb.Converter
 import com.intellij.util.xmlb.annotations.{OptionTag, XCollection}
 import javax.swing.JCheckBox
+import org.jetbrains.bsp.project.importing.setup.FastpassConfigSetup
 import org.jetbrains.bsp.settings.BspProjectSettings.{AutoConfig, AutoPreImport, BspServerConfig, BspServerConfigConverter, PreImportConfig, PreImportConfigConverter}
 import org.jetbrains.bsp.{BspBundle, _}
 import org.jetbrains.plugins.scala.project.ProjectExt
```
jastice, line 19: imports unused

scala/scala-impl/src/org/jetbrains/sbt/project/FastpassProjectImportProvider.scala
```diff
@@ -0,0 +1,12 @@
+package org.jetbrains.sbt.project
+
+import com.intellij.openapi.vfs.VirtualFile
+import Iterator.iterate
+
+object FastpassProjectImportProvider {
```
jastice: why is this under the sbt package?

bsp/src/org/jetbrains/bsp/project/importing/setup/FastpassConfigSetup.scala
jastice: just giving an OK status is fine, but you could also pass in any collected log messages

bsp/src/org/jetbrains/bsp/project/importing/setup/FastpassConfigSetup.scala
```diff
@@ -0,0 +1,75 @@
...
+    if(process.exitValue() == 1) {
+      Success(BuildMessages.empty.status(BuildMessages.OK))
+    } else {
+      Failure(BspErrorMessage(s"Command ${processBuilder.command.asScala}"))
+    }
+  }
+
+  override def run(implicit reporter: BuildReporter): Try[BuildMessages] = {
+    reporter.start()
+    val process = processBuilder.start()
```
jastice, line 88: you may want to send process output to the reporter

bsp/src/org/jetbrains/bsp/project/importing/BspResolverLogic.scala
```diff
@@ -14,13 +14,15 @@ import com.intellij.openapi.externalSystem.util.ExternalSystemApiUtil
...
 import com.intellij.openapi.util.io.FileUtil
+import com.intellij.openapi.vfs.LocalFileSystem
 import org.jetbrains.bsp.BspUtil._
 import org.jetbrains.bsp.data._
 import org.jetbrains.bsp.project.BspSyntheticModuleType
 import org.jetbrains.bsp.project.importing.BspResolverDescriptors._
 import org.jetbrains.bsp.{BSP, BspBundle}
 import org.jetbrains.plugins.scala.project.Version
 import org.jetbrains.plugins.scala.project.external.{JdkByHome, JdkByVersion}
+import org.jetbrains.sbt.project.FastpassProjectImportProvider
```
jastice: unused imports

jastice (commented):
Thanks! I'll try it out next week!

scala/scala-impl/src/org/jetbrains/sbt/project/FastpassProjectImportProvider.scala
tpasternak: ok, moved it. It was there alongside Mill provider

bsp/src/org/jetbrains/bsp/settings/settings.scala
tpasternak: done

bsp/src/org/jetbrains/bsp/project/importing/BspResolverLogic.scala
tpasternak: done

bsp/src/org/jetbrains/bsp/project/importing/setup/FastpassConfigSetup.scala
tpasternak: the messages are now passed to reporter

scala/scala-impl/src/org/jetbrains/sbt/project/FastpassProjectImportProvider.scala
jastice: oh, I thought I had moved that one

bsp/src/org/jetbrains/bsp/project/importing/setup/FastpassConfigSetup.scala
```diff
@@ -50,18 +50,28 @@ class FastpassConfigSetup(processBuilder: ProcessBuilder) extends BspConfigSetup
...
-      stdoutReader.lines().forEach(reporter.info(_, None))
-      stderrReader.lines().forEach(reporter.error(_, None))
+      val stderrLines = stderrReader.lines()
+      stderrLines.forEach{ line =>
+        buildMessages = buildMessages.addError(line)
+        reporter.error(line, None)
+      }
+      stdoutReader.lines().forEach{ line =>
+        buildMessages = buildMessages.message(line)
+        reporter.info(line, None)
```
jastice: may want to just use BuildReporter.log - info may be displayed as individual nodes (if build toolwindow is used)

jastice:
Do you have a project example that I could try this with?

bsp/src/org/jetbrains/bsp/project/importing/projectImport.scala
```diff
@@ -203,14 +212,18 @@ class BspProjectImportProvider(builder: BspProjectImportBuilder)
...
   override def createSteps(context: WizardContext): Array[ModuleWizardStep] = {
     builder.reset()
     builder.autoConfigure(context.getProjectDirectory.toFile)
+    builder.setFileToImport(context.getProjectDirectory.toString)
 
-    Array(
-      new BspSetupConfigStep(context, builder),
+    // Calling super.createSteps here is essential to ensure that
+    // the `NewProjectUtil.doCreate` is called *after* resolveProjectInfo(isPreview=true)
+    super.createSteps(context) ++ Array(
```
jastice: This causes an additional dialog on import, which isn't needed most of the time. Can we avoid it?

tpasternak:
Ok, it seems this was an issue with my homebrew setup, no with Macs in general. You can try to import `src/java/org/pantsbuild/example/hello/greet` from this repo `https://github.com/scalameta/pants`

tpasternak:
BTW - there's an assumption here that there is a `fastpass exec` in your path - you can download it with `coursier bootstrap org.scalameta:fastpass_2.12:1.5.0 -r sonatype:snapshots --main scala.meta.fastpass.Fastpass -o fastpass`

jastice:
> BTW - there's an assumption here that there is a `fastpass exec` in your path - you can download it with `coursier bootstrap org.scalameta:fastpass_2.12:1.5.0 -r sonatype:snapshots --main scala.meta.fastpass.Fastpass -o fastpass`

That's alright for now, but can we avoid it by bundling coursier? We'll probably want to do this anyway for bazel-bsp.

bsp/src/org/jetbrains/bsp/project/importing/projectImport.scala
tpasternak: Ok, managed to do it without the extra step, just commited. Nevertheless, I'm really not sure if it's safe to skip `super.createSteps` here

tpasternak:
> That's alright for now, but can we avoid it by bundling coursier? We'll probably want to do this anyway for bazel-bsp.

Ideally, I'd prefer it to be somehow configurable, but of course, you're right that bundling coursier would simplify stuff a lot

jastice:
During the initial setup I'm getting a lot of lines prefixed with `ERROR:` in the indicator

image

tpasternak:
> During the initial setup I'm getting a lot of lines prefixed with `ERROR:` in the indicator

This is because I'm calling `reporter.info`and stream all logs from fastpass to to this window. As I said, I didn't know what to do with this log. I haven't tried `reporter.log` yet

jastice:
> This is because I'm calling `reporter.info`and stream all logs from fastpass to to this window. As I said, I didn't know what to do with this log. I haven't tried `reporter.log` yet

Yup, looks like you're treating every line from stderr as error, but if that's not how fastpass treats them I'd go with 
```
buildMessages = buildMessages.message(line)
reporter.log(line)
```
for both stderr and stdout

jastice:
I imported `/Users/jast/playspace/pants/examples/src/java/org/pantsbuild/example/hello/greet`

After the import is done, I'm getting an empty `.bloop` directory, and correspondingly, this output on refresh:

```
[D] Waiting for a connection at local:///var/folders/5k/4b5my29x26q_dhrys3ghf9wh0000gn/T/bsp-launcher12815398736040296871/bsp.socket...

The server is listening for incoming connections at local:///var/folders/5k/4b5my29x26q_dhrys3ghf9wh0000gn/T/bsp-launcher12815398736040296871/bsp.socket...

Starting thread that pumps stdin and redirects it to the bsp server...

Starting thread that pumps server stdout and redirects it to the client stdout...

Empty build targets. Expected at least one build target identifier.
Empty build targets. Expected at least one build target identifier.
Empty build targets. Expected at least one build target identifier.
```

tpasternak:
Oh, could have been fixed in ffab741

jastice:
Still getting the same issue

image

bsp/src/org/jetbrains/bsp/project/importing/setup/FastpassConfigSetup.scala
```diff
@@ -55,11 +55,11 @@ class FastpassConfigSetup(processBuilder: ProcessBuilder) extends BspConfigSetup
       val stderrLines = stderrReader.lines()
       stderrLines.forEach{ line =>
         buildMessages = buildMessages.addError(line)
```
jastice, line 64: use `buildMessages.message` here too if you don't expect `line` to be actual errors in general

tpasternak:
@jastice I added support for the case when is BSP project already exists in the path - it's not running fastpass, just imports that BSP project

PR #566, opened 2020-09-03
Use fastpass from sources instead of expecting it is in PATH

tpasternak:


jastice (approved):
LGTM. Would it make sense to autodetect either sources or path, or have it configurable?

tpasternak:
Perhaps it would be best to make it configurable, but for now maybe this is enough. If something goes really wrong, then the user may edit `fastpass/bin/fastpass` file.

PR #567, opened 2020-09-04
More restrictive Fastpass project detection

tpasternak:
Previously it could return return true for non-Fastpass projects in some
certain cases.


jastice (approved):
Thanks for the fix!

PR #568, opened 2020-09-10
Do not spawn ConfigSetupTask for non-BSP imports

tpasternak:
It seems that the `onWizardFinished` method of BspSetupConfigStep is called
each time, even if other then BSP import method was chosen. This means, we need
to check this inside the method.

### DIFF
--- a/bsp/src/org/jetbrains/bsp/project/importing/bspConfigSteps.scala
+++ b/bsp/src/org/jetbrains/bsp/project/importing/bspConfigSteps.scala
@@ -210,11 +210,13 @@ class BspSetupConfigStep(wizardContext: WizardContext, builder: BspProjectImport
   override def onWizardFinished(): Unit = {
     // TODO this spawns an indicator window which is not nice.
     // show a live log in the window or something?
-    updateDataModel() // without it runSetupTask is null
-    builder.prepare(wizardContext)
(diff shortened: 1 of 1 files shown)


tpasternak:
I was looking for some cleaner solution, like checking ExternalSystemId, but nothing good came to my mind. On the other hand, moving this logic outside `BspSetupConfigStep` could be harmful

jastice:
Thanks! I'll have another look into what's up with this. The whole wizard code paths are pretty convoluted

jastice:
I could imagine we could also do it by detecting if any configSetupTask is set, but I'll merge it as is. Thanks!

tpasternak:
I was also thinking about it, but it's a tricky part. `configSetupTask` is intended to be set in `updateDataModel`, but this function is not called if the step is not shown. This means you end up with unset `configSetupTask` in `onWizardFinished`. That's why I added enforced `updateDataModel` there some time ago, but then it appeared it's called for non-BSP importers too.

jastice:
Yup, that wizard has arcane ways!

tpasternak:
Thank you! Would it be possible to trigger nightly build with it, too?

jastice:
It should be up by now!

tpasternak:
Thanks 👍

PR #569, opened 2020-09-22
Fix scalatest sbt shell command parameters

bednam:
ScalaTestRunConfiguration run with sbt (`Use sbt` setting is set to true in TestConfigurationData) doesn't work as expected for package, class nor method test kind. The reason is that the `testOnly` sbt command run in the shell isn't followed by expected arguments that should be provided by `buildTestOnly` method in SbtCommandsBuilderBase implementation for ScalaTestRunConfiguration. The change included in this PR fixes the command parameters creation when `classKey` isn't present which is the case for ScalaTestRunConfiguration.


unkarjedy:
Thanks a lot!
I will manually cherry-pick it to our release branch.

PR #570, opened 2020-10-02
[idea202.x port] Do not spawn ConfigSetupTask for non-BSP imports

tpasternak:
It seems that the `onWizardFinished` method of BspSetupConfigStep is called
each time, even if other then BSP import method was chosen. This means, we need
to check this inside the method.

idea202.x PR: https://github.com/JetBrains/intellij-scala/pull/568


jastice:
Thanks, good spot!

PR #572, opened 2020-10-26
Handle upcoming mill embedded BSP and fix mill BSP install

joan38:
BSP support is now integrated in Mill by default.
This PR handle backcompat with older versions of Mill but not super proud of the code. anyway this is temporary and we'll be able to drop all of this when dropping support for older Mill version.

This is also fixing the generation of the bsp config from mill that got broken for some reason.

Thanks


jastice (approved):
Great to see Mill is going full BSP now, I didn't even notice!

bsp/src/org/jetbrains/bsp/project/importing/bspConfigSteps.scala
```diff
@@ -103,9 +103,6 @@ object bspConfigSteps {
           builder.setPreImportConfig(NoPreImport)
           // server config to be set in next step
           SbtConfigSetup(workspace)
-        case bspConfigSteps.MillSetup =>
-          builder.setPreImportConfig(NoPreImport)
-          MillConfigSetup(workspace)
```
jastice, line 108: Taking this case out causes a compile error since we have fatal warnings now. Should it just be an empty case, or the `MillSetup` object removed?
```
[error] /Users/jast/workspace/scalaUltimate/community/bsp/src/org/jetbrains/bsp/project/importing/bspConfigSteps.scala:89:12: match may not be exhaustive.
[error] It would fail on the following input: MillSetup
[error]     } else configSetup match {
```

jastice (changes requested):
noticed compile error :)

bsp/src/org/jetbrains/bsp/project/importing/bspConfigSteps.scala
joan38: I pushed a new implementation which is much better actually.

jastice (approved):

joan38:
Thanks @jastice

PR #573, opened 2020-10-29
BSP import: Set language level matching JDK

tpasternak:


jastice:
Thanks! But I think we don't necessarily want the same language level as the project jdk. I think `javaVersion` in `JvmBuildTarget` should take precedence here.

tpasternak:
Hey, this was for project-level language version. I can setup javaVersion for modules, too. Do you know how I can test this?

jastice:
Project-level is also not necessarily JDK version. In sbt we infer it from `javacOptions` of the root project, if available, then fall back to jdk version. Maybe we can use a similar strategy here?

tpasternak:
Ok, now I see I probably took the wrong field again

tpasternak:
@jastice I think it's ready now. We don't have any equivalent of root project in BSP, but we can get the version from javacOptions for each module.

jastice (approved):

PR #577, opened 2020-11-26
don't add trailing commas for block parameters not enclosed in parentheses

piton4k:
Hey, this is my first contribution here, so let me know if I need to do something differently.

This change should fix https://youtrack.jetbrains.com/issue/SCL-16515

I'm not particularly fond of the code. I didn't study the API very deeply, I just went with the simplest thing that seemed to fix the issue. I'm open to any suggestions to make it cleaner.

I'd like to use the setting `Trailing Comma: Add when multiline` but until this is fixed, it is nearly unusable.


unkarjedy:
Thanks for your contribution!
I will manually cherry-pick the fix.

niktrop:
@piton4k Please also sign our contributor license agreement: 

PR #579, opened 2021-01-22
Add an inspection for .unary_! references

chrislemaire:
This MR adds the inspection for postfix unary negations including a quickfix to turn such negations into the prefix operator version.

I could not yet find a way to test the inspection and the quick-fix properly and I couldn't find an example using `LocalInspectionTool` yet. Is there currently a good way to do this?

Solves #SCL-10335


scala/scala-impl/src/org/jetbrains/plugins/scala/lang/formatting/settings/ImportsPanel.java
```diff
@@ -264,7 +264,7 @@ private static void setValue(final JRadioButton rb, final boolean value) {
         final JPanel panel3 = new JPanel();
         panel3.setLayout(new GridLayoutManager(1, 1, new Insets(0, 0, 0, 0), -1, -1));
         panel2.add(panel3, new GridConstraints(0, 0, 1, 1, GridConstraints.ANCHOR_CENTER, GridConstraints.FILL_BOTH, GridConstraints.SIZEPOLICY_CAN_SHRINK | GridConstraints.SIZEPOLICY_CAN_GROW, GridConstraints.SIZEPOLICY_CAN_SHRINK | GridConstraints.SIZEPOLICY_CAN_GROW, null, null, null, 0, false));
-        panel3.setBorder(IdeBorderFactory.PlainSmallWithIndent.createTitledBorder(BorderFactory.createEtchedBorder(), this.$$$getMessageFromBundle$$$("messages/ScalaBundle", "imports.panel.classes.to.use.only.with.prefix"), TitledBorder.DEFAULT_JUSTIFICATION, TitledBorder.DEFAULT_POSITION, null, null));
+        panel3.setBorder(BorderFactory.createTitledBorder(BorderFactory.createEtchedBorder(), this.$$$getMessageFromBundle$$$("messages/ScalaBundle", "imports.panel.classes.to.use.only.with.prefix"), TitledBorder.DEFAULT_JUSTIFICATION, TitledBorder.DEFAULT_POSITION, null, null));
```
chrislemaire: I am not sure what happened in these lines. I didn't touch them manually, is this perhaps some automatic refactor IntelliJ did? And, if so, is it desired or should I revert it?

chrislemaire:
I am not sure whether the way I mentioned the issue number in this commit/MR is correct, should it be changed?

niktrop:
@chrislemaire Thank you for the contribution! Please sign our Contributor Licence Agreement:

chrislemaire:
I have signed the CLA, thanks for the link!

scala/scala-impl/src/org/jetbrains/plugins/scala/lang/formatting/settings/ImportsPanel.java
SrTobi: You can remove these changes for a cleaner commit. They happen automatically and are not needed here

SrTobi:
Hi, thanks for the PR, I'll have a more in-depth look at the beginning of next week.

scala/scala-impl/src/org/jetbrains/plugins/scala/codeInspection/syntacticSimplification/PostfixUnaryNegationInspection.scala
```diff
@@ -0,0 +1,52 @@
...
+  /**
+   * Creates the quick fix for postfix unary negation references. This fix creates a new prefix negation expression and
+   * replaces the old expression for it.
+   * @param ref The old reference expression containing the unary_! identifier.
+   * @return The quick-fix.
+   */
+  private def createQuickfix(ref: ScReferenceExpression) =
+    Some(new PostfixUnaryNegationQuickFix(ScalaInspectionBundle.message("use.prefix.negation"), ref))
+
+  private[syntacticSimplification] class PostfixUnaryNegationQuickFix(name: String, element: ScReferenceExpression)
```
SrTobi: Pls add an `@Nls` annotation to the `name` parameter to indicate that `name` must contain a translated string

scala/scala-impl/src/org/jetbrains/plugins/scala/codeInspection/syntacticSimplification/PostfixUnaryNegationInspection.scala
```diff
@@ -0,0 +1,52 @@
...
+
+object PostfixUnaryNegationInspection {
+  /**
+   * Checks whether the expression is a postfix unary_! expression. Just unary_! is a method reference and should still
+   * be allowed, however.
+   * @param ref The reference expression to check.
+   * @return Whether the given expression is a postfix unary negation.
+   */
+  private def isPostfixUnaryNegation(ref: ScReferenceExpression): Boolean =
+    !ref.textMatches("unary_!") && ref.getLastChild.textMatches("unary_!")
```
SrTobi: this only works for member accesses like `true.unary_!` but not for postfix expressions like `true unary_!`.
You might also want to check for ScPostfixExpr (use `ScPostfixExpr.operation.refName` for the check)

Also don't use getLastChild here as that requires knowledge about how the psi tree is structured exactly (about which you can only be certain by looking into the parser). It's better to use `ref.refName`, and to check if it is not a simple reference use `ref.isQualified`.

scala/scala-impl/src/org/jetbrains/plugins/scala/codeInspection/syntacticSimplification/PostfixUnaryNegationInspection.scala
```diff
@@ -0,0 +1,52 @@
...
+   * @return The quick-fix.
+   */
+  private def createQuickfix(ref: ScReferenceExpression) =
+    Some(new PostfixUnaryNegationQuickFix(ScalaInspectionBundle.message("use.prefix.negation"), ref))
+
+  private[syntacticSimplification] class PostfixUnaryNegationQuickFix(name: String, element: ScReferenceExpression)
+    extends AbstractFixOnPsiElement[ScReferenceExpression](name, element) {
+
+    override protected def doApplyFix(element: ScReferenceExpression)(implicit project: Project): Unit = {
+      val negationExpr = ScalaPsiElementFactory.createExpressionFromText(s"!${element.getFirstChild.getText}")
```
SrTobi: Don't use `element.getFirstChild` here. It's better to use `AbstractFixOnTwoPsiElements` and store the element that should be replaced and the element that should be preserved separately (which are `ref` and `ref.qualifier` respectively in the case of the ScReferenceExpression).

This makes it also easy when you use this very same quickfix for ScPostfixExpr (where you can use the postfix expression itself and `postfixExpr.operand` as arguments to the Quickfix)

SrTobi (changes requested):
This looks good so far, but it should also be able to handle postfix expressions and I think this inspection should not be exclusively for unary_! but also the other three unary operators (`-`, `+`, `~`). You can call it something like `PostfixUnaryOperationInspection` then.

Last but not least, it needs a test with some common examples. especially the edge case where the unary method is called without qualifier. Also a test case with underscore section would be good, like `Seq(-1, -2, 3).map(_.unary_-)` -> `Seq(-1, -2, 3).map(-_)`. You can basically copy/adapt `test/org/jetbrains/plugins/scala/codeInspection/cast/ScalaRedundantConversionInspectionTest.scala` (in your case you don't need to override `descriptionMatches` because your inspection message is static).

Use `checkTextHasError` to check if the inspection is shown at the right position, `testQuickFix` to check if the quickfix is applied correctly, and `checkTextHasNoErrors` to check that the inspection is not shown if it does not apply.

scala/scala-impl/src/org/jetbrains/plugins/scala/codeInspection/syntacticSimplification/PostfixUnaryNegationInspection.scala
```diff
@@ -0,0 +1,52 @@
...
+class PostfixUnaryNegationInspection extends AbstractRegisteredInspection {
+  override protected def problemDescriptor(element: PsiElement,
+                                           maybeQuickFix: Option[LocalQuickFix],
+                                           descriptionTemplate: String,
+                                           highlightType: ProblemHighlightType)
+                                          (implicit manager: InspectionManager, isOnTheFly: Boolean)
+  : Option[ProblemDescriptor] =
+    element match {
+      case ref: ScReferenceExpression if isPostfixUnaryNegation(ref) =>
+        super.problemDescriptor(element.getLastChild, createQuickfix(ref), descriptionTemplate, highlightType)
```
chrislemaire: Is there a way to avoid this `element.getLastChild` here as well? I wasn't able to find an alternative that returns the `PsiElement` for the operator reference yet, just the `refName`.

scala/scala-impl/src/org/jetbrains/plugins/scala/codeInspection/syntacticSimplification/PostfixUnaryNegationInspection.scala
chrislemaire: Ahh that's exactly what I was looking for! I have applied this in the latest commit.
I have added the post-fix expression check as a separate case in the case match and with a separate predicate function, is this fine?

scala/scala-impl/src/org/jetbrains/plugins/scala/codeInspection/syntacticSimplification/PostfixUnaryNegationInspection.scala
chrislemaire: I was able to implement the fix using `AbstractFixOnTwoPsiElements` and a mapping of unary operator names to their operator counterparts.
I agree it adds a nicer structure to creating the fix :)

scala/scala-impl/test/org/jetbrains/plugins/scala/codeInspection/syntacticSimplification/PostfixUnaryOperationInspectionTest.scala
```diff
@@ -0,0 +1,199 @@
...
+         |  def f(): Seq[Int] = Seq(1, 2, 3).map(A.${START}unary_~$END)
+         |}
+      """.stripMargin
+
+    val result =
+      s"""object A {
+         |  def unary_~(n: Int) = ~n
+         |  def f(): Seq[Int] = Seq(1, 2, 3).map(~A)
+         |}
+      """.stripMargin
```
chrislemaire, line 48: An interesting find I had during testing was that it is allowed to have the unary operator in front of an object containing the function. I presume there's some nice use-cases for this, but I was quite surprised when I saw it work in this example.

scala/scala-impl/src/org/jetbrains/plugins/scala/codeInspection/syntacticSimplification/PostfixUnaryNegationInspection.scala
SrTobi: Yeap, that's fine. I mean, `ScReferenceExpression` and `ScPostfixExpr` do not have a common trait that offers all the data you want, so you need the different cases anyway

scala/scala-impl/src/org/jetbrains/plugins/scala/codeInspection/syntacticSimplification/PostfixUnaryNegationInspection.scala
SrTobi: If you look into the implementation of ScReference.refName, you immediately see the answer :smile: 

```
trait ScReference extends ScalaPsiElement with PsiPolyVariantReference {
  [...]

  def nameId: PsiElement

  def refName: String = {
    assert(nameId != null, s"nameId is null for reference with text $getText; parent: ${getParent.getText}")
    nameId.getText
  }

  [...]
```

chrislemaire: Oh wow, that's quite silly of me. Thanks for the trouble! :sweat_smile:

scala/scala-impl/test/org/jetbrains/plugins/scala/codeInspection/syntacticSimplification/PostfixUnaryOperationInspectionTest.scala
SrTobi: Ahh yes, this is unexpected, indeed. I guess it's because the rewrite rule do happen on syntax level before any type checking/resolving.

chrislemaire: I guess so. On top of that, this does also compile and run with `unary_~` referenced through `~A`

SrTobi (approved):
Looks great! Thanks again for your help :pray:

PR #580, opened 2021-01-31
Run mill with -i to workaround issues with Apple Silicon support in mill

joan38:
And Upgrade SBT

We are having issues to make mill work with Apple's arm64 is client/server mode. This is a workaround to make the BSP import work.

Thanks


project/build.properties
```diff
@@ -1 +1 @@
-sbt.version=1.3.13
\ No newline at end of file
+sbt.version=1.4.7
```
jastice: The sbt version is on 1.3.x deliberately for now because a change in the error reporting breaks our internal TeamCity integration

joan38: Ok I reverted this change

jastice (approved):

jastice:
Thanks!

PR #582, opened 2021-04-03
Added support for derevo library (https://github.com/tofu-tf/derevo).

odbc:
DerevoInjector injects circe encoder/decoder if there is @derevo.derive annotation with derevo.circe.encoder/derevo.circe.decoder arguments.


SrTobi:
Hi @odbc. Thanks for the pull request.
This seems to me as if it implements only a really small aspect of the derive macro (only works for encoder, decoder).
Can we not make this more general to work for custom derivations, for example?
Also the `@phantom` annotation does not seems to be supported. is that so rarely used or why is it missing?

From what I see it shouldn't be too hard to generate implicits for custom typeclasses. We just have to get the type of the class listed in the `derive(...)` calls, see which `InstanceDef` they implement, and generate type parameters with context bounds (or not in case of `@phantom`) + the correct return type. done. With that Enocder/Decoder should work automatically, shouldn't they?
Or do I miss any major problems?

PR #586, opened 2021-05-03
Fix broken bloop launcher classpath

hamnis:
Bloop has broken its bloop launcher, so unless we include this dependency. The launcher is broken.
A possible workaround is to install bloop as mentioned in the attached issue, but this should allow the existing setup to work.

Should fix: 
https://youtrack.jetbrains.com/issue/SCL-18969


jastice:
Thanks! But I think since Bloop presumably fixes this in https://github.com/scalacenter/bloop/commit/d8ece55ab4707ff861ab1aa9928b73be8293e3a8
we better just depend on the current development release of Bloop (1.4.8-26-cbddb8ba) instead

hamnis:
Allright. Any ETA for a possible release with that fix?

jastice:
Just pushed a fix. It should make it into tomorrow's Nightly release if there's no build issues.

hamnis:
@jastice awesome. thanks a lot!

PR #587, opened 2021-05-10
Remove restriction on max parallel Scala Compile Server threads

neko-kai:
I have more than 6 cores and because of that `sbt compile` finishes faster than an IDEA build on my machine purely due to parallelism. I believe the restriction of maximum 6 threads may be a tad too low and could probably be increased or completely unrestricted. (I chose to do the latter in this change)


jastice:
Thanks! Excuse the delay please :)

PR #590, opened 2021-06-06
Upgrading to sbt 1.5.x

hmemcpy:
I upgraded the project's sources to use sbt 1.5.3, migrated to the new slash syntax, and removed `DottyPlugin_Patched` (per TODOs).

Ran `sbt compile` - everything seemed to compile successfully!

I also had to change `intellijVersion` in `dependencies.scala` locally to `212` instead of `212.3500`, because this build is 404 for me.


jastice:
Thanks! For the moment we can't merge this as sbt versions since 1.4.x aren't quite compatible with our TeamCity runner, so we'll need to fix that on side or the other...

niktrop:
Hi Igal, thanks for the contribution! I think we may merge it without the version change itself to use modern syntax. Please sign our [Contributor License Agreement](https://www.jetbrains.com/agreements/cla/).

hmemcpy:
Done!

I honestly sent this because I wanted IntelliJ to stop prompting me to upgrade sbt every time I imported the plugin sources :)

Thanks for accepting!

jastice:
Oh, you can disable that prompt by disabling "allow overriding sbt version" in the project's sbt setting. I think I also disabled it by default for new projects? At least I wanted to!

unkarjedy:
@jastice not yet, but looks like it's time to https://youtrack.jetbrains.com/issue/SCL-18944

niktrop:
Merged manually: https://github.com/JetBrains/intellij-scala/commit/bdcf8c895d93baf00301cbe43dc31ac137927d55

PR #591, opened 2021-06-15
Use command key for copying from sbt shell on mac

mzarnowski:
Currently, to copy the selection from sbt shell, one has to use `ctrl-c` even though on mac it should actually be [`Command-C`](https://support.apple.com/en-us/HT209651)


jastice:
Hi, thanks! I hadn't added this explicitly because on Mac Cmd+C should work out of the box - it does for me anyway. Does it not work for you?

mzarnowski:
Hi! Thanks for finding the time to look into it.
I have just installed presentation assistant plugin to help me capture the screen and share an example... and Cmd+C started working in sbt-shell. 
After disabling it (without restarting), the shortcut seems to be working still. 
This didn't affect `Ctrl+C` - I can still copy selection using it, so now there are two shortcuts for the same action.

All in all, since the original issue got solved automagically, I no longer need this change - please, close the PR if you deem it unnecessary.

Thanks!

jastice:
Classic 🤣 
Thanks for the feedback, and let us know if you find some condition where it doesn't work.

PR #592, opened 2021-06-15
#SCL-19157 Add integration-test to module description

github-samuel-clarenc:
The integration tests are not correctly imported in the module/*.iml file : the `isTestSource` property is set to `false`.

This PR fixes this behavior by adding the [BuildTargetTag.INTEGRATION_TEST](https://github.com/build-server-protocol/build-server-protocol/blob/master/bsp4j/src/main/java/ch/epfl/scala/bsp4j/BuildTargetTag.java#L7) to the `isTest` value.

### DIFF
--- a/bsp/src/org/jetbrains/bsp/project/importing/BspResolverLogic.scala
+++ b/bsp/src/org/jetbrains/bsp/project/importing/BspResolverLogic.scala
@@ -309,7 +309,7 @@ private[importing] object BspResolverLogic {
 
     val targetDeps = target.getDependencies.asScala.toSeq
 
-    val data = if(tags.contains(TEST))
+    val data = if(tags.contains(TEST) || tags.contains(INTEGRATION_TEST))
       dataBasic.copy(
         targetTestDependencies = targetDeps,
         testOutput = outputPath,
@@ -364,7 +364,10 @@ private[importing] object BspResolverLogic {
     val merged = mergeModules(ancestors)
     val id = sharedModuleId(targets)
     val sources = sourceRoots ++ generatedSourceRoots
-    val isTest = targets.forall(_.getTags.asScala.contains(BuildTargetTag.TEST))
+    val isTest = targets.forall{ buildTarget =>
+      val tags = buildTarget.getTags.asScala
+      tags.contains(BuildTargetTag.TEST) || tags.contains(BuildTargetTag.INTEGRATION_TEST)
+    }
 
     val inheritorData = merged.data.copy(
       id = id,
--- a/bsp/test/org/jetbrains/bsp/project/importing/BspResolverLogicTest.scala
+++ b/bsp/test/org/jetbrains/bsp/project/importing/BspResolverLogicTest.scala
@@ -1,9 +1,10 @@
 package org.jetbrains.bsp.project.importing
 
-import ch.epfl.scala.bsp4j.{BuildTarget, BuildTargetCapabilities, BuildTargetIdentifier}
+import ch.epfl.scala.bsp4j.{BuildTarget, BuildTargetCapabilities, BuildTargetIdentifier, BuildTargetTag, SourceItem, SourceItemKind, SourcesItem}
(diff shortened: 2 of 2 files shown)


bsp/test/org/jetbrains/bsp/project/importing/BspResolverLogicTest.scala
```diff
@@ -27,4 +28,38 @@ class BspResolverLogicTest {
     assert(rootModule.data.targets.head == target)
   }
 
+  @Test
+  def testCalculateModuleDescriptionsBaseDirWithTest(): Unit = {
+    val target = new BuildTarget(
+      new BuildTargetIdentifier("ePzqj://jqke:540/n/ius7/jDa/t/z78"),
```
github-samuel-clarenc, line 34: I used the same `BuildTargetIdentifier` than the previous test but I don't really know whether it was randomly generated or not.

bsp/src/org/jetbrains/bsp/project/importing/BspResolverLogic.scala
```diff
@@ -309,7 +309,7 @@ private[importing] object BspResolverLogic {
 
     val targetDeps = target.getDependencies.asScala.toSeq
 
-    val data = if(tags.contains(TEST))
+    val data = if(tags.contains(TEST) || tags.contains(INTEGRATION_TEST))
```
github-samuel-clarenc, line 312: Can't we use here the [target.getCapabilities.getCanTest](https://github.com/build-server-protocol/build-server-protocol/blob/43869cf93e28d6e05d0e9052470447c67d4e4851/bsp4j/src/main/xtend-gen/ch/epfl/scala/bsp4j/BuildTargetCapabilities.java#L14) ? 
```suggestion
    val data = if(target.getCapabilities.getCanTest)
```

samuelClarencTeads:
This PR won't work this way : the `test` tag is present for a unit test build target but no `integration-test` is present on an IntegrationTest build target : I don't know where the [BuildTarget.tag](https://github.com/build-server-protocol/build-server-protocol/blob/43869cf93e28d6e05d0e9052470447c67d4e4851/bsp4j/src/main/xtend-gen/ch/epfl/scala/bsp4j/BuildTarget.java#L23) is defined

github-samuel-clarenc:
As a workaround, we can manually add the tag in the `build.sbt` : 

```
IntegrationTest/bspBuildTarget := (IntegrationTest/bspBuildTarget).value.withTags(Vector("test"))
```

github-samuel-clarenc:
As https://github.com/sbt/sbt/pull/6551 is merged on SBT side, do you think we can go on on intellij side @jastice  ?

bsp/test/org/jetbrains/bsp/project/importing/BspResolverLogicTest.scala
jastice: I think this was extracted as a failing case from ScalaCheck generated cases once

jastice:
Hi! Thanks for this, and sorry for the delay.

This one is a bit tricky. In IntelliJ, we only support the compile and test scopes. With the current logic, compile and test targets get merged into a single IntelliJ module. If we also merge in the Integrationtest scope, you'll likely run into problems with mixed classpaths when trying to run/edit things on IntelliJ side. This is one reason we recommend not using this scope at all when working with IntelliJ, and instead configuring separate projects.

However, we could try explicitly not merging integrationtest target instead, and also optionally not merging compile and test targets at all.

github-samuel-clarenc:
Thanks for your response ! 

If the integration-tests are not merged anymore, they will probably appear in IntelliJ as a new module, right ? That's not really what we need here ! 

As we have a workaround, I guess we could probably close this PR ! 

Thanks anyway for your time.

jastice:
Thanks for the feedback! I'll close this for now, but consider how to support such targets as separate modules.

Vaguely related Scala plugin ticket: https://youtrack.jetbrains.com/issue/SCL-16115

PR #594, opened 2021-06-28
SCL-18372 Instruct BSP server to workspace/reload during refresh

retronym:


bsp/src/org/jetbrains/bsp/project/importing/BspProjectResolver.scala
```diff
@@ -72,7 +72,7 @@ class BspProjectResolver extends ExternalSystemProjectResolver[BspExecutionSetti
     reporter.startTask(structureEventId, None, BspBundle.message("bsp.resolver.resolving.build.structure"))
     val targetsEventId = BuildMessages.randomEventId
     reporter.startTask(targetsEventId, Some(structureEventId), BspBundle.message("bsp.resolver.build.targets"))
-    val targetsRequest = server.workspaceBuildTargets()
+    val targetsRequest = server.workspaceReload().thenCompose(_ => server.workspaceBuildTargets())
```
retronym: You might want to retain backwards compatibility with BSP servers that don't implement this method. We don't have [protocol version negotiation yet](https://github.com/build-server-protocol/build-server-protocol/issues/56), but I guess we could just swallow a "not-implemented" error here.

jastice: There's a flag for support of this: `BuildServerCapabilities.canReload`
But I see you're already using it :)

project/dependencies.scala
```diff
@@ -10,7 +10,7 @@ object Versions {
   val bloopVersion = "1.4.8-26-cbddb8ba"
   val zincVersion = "1.4.0-M12"
   val intellijVersion = "212.4486"
-  val bspVersion = "2.0.0-M12+27-4994bd9d-SNAPSHOT"
+  val bspVersion = "2.0.0-M14+7-56ef5d5f-SNAPSHOT"
```
jastice: We can use M14 without the latest snapshot

jastice (approved):
Thanks! I must have forgotten to add this ^^

project/dependencies.scala
retronym:
```suggestion
  val bspVersion = "2.0.0-M14"
```

jastice:
YouTrack issue: https://youtrack.jetbrains.com/issue/SCL-19230

PR #596, opened 2021-09-08
BSP import: don't add Scala library to runtime classpath in Scala SDK

lrytz:
When importing a project through bsp, the compiler classpath
(scala-library, scala-reflect, scala-compiler) obtained through
bsp should only be used for the "Compiler classpath" section of the
Scala SDK.

Before this change, the classpath was also added to the module's
`<CLASSES>` ("Standard library" section of the Scala SDK in the UI).
This is
  - slightly incorrect, because it adds the scala-compiler jar to the
    module's classpath
  - not needed, because the scala-library is added to the module's
    classpath as a separate library

https://github.com/scala/scala-dev/issues/786#issuecomment-914567436


jastice (approved):
Had to check for unintended side effects, but it seems fine :)

jastice:
Merging was causing troubles so I applied this change manually. Thank you!

https://github.com/JetBrains/intellij-scala/commit/86c9f393244b4b65230d77ad1c53a8d002ada05d

PR #601, opened 2021-12-21
update Zinc version to 1.5.9

gerald732:
Zinc 1.5.9 updates log4j to 2.16.0, which disables JNDI lookup and fixes a denial of service vulnerability (CVE-2021-45046)

### DIFF
--- a/project/dependencies.scala
+++ b/project/dependencies.scala
@@ -8,7 +8,7 @@ object Versions {
   // run runtimeDependencies/update manually
   val sbtVersion: String = Sbt.latest
   val bloopVersion = "1.4.8-81-e170cd66"
-  val zincVersion = "1.5.7"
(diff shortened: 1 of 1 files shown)


niktrop:
@gerald732 Thank you for the fix, but it is not necessary. We've removed log4j classes from the plugin distribution altogether as they are not actually used at runtime: 
https://github.com/JetBrains/intellij-scala/blob/ae3cb5fe8b1c03ed2e4d3c5c4631e8e62e5ef4be/project/dependencies.scala#L107

Plugin with the fix is already published for 2021.2 and 2021.3.

gerald732:
Thanks for the quick response @niktrop. Will it be made available for 2021.1 as well?

niktrop:
We didn't plan to do that as there is no real security problem in our case.

PR #602, opened 2021-12-22
Pass environment used in MacOS terminal to bsp server

lukaszwawrzyk:
By looking at `EnvironmentUtil.getEnvironmentMap` docs it looks like it is a known issue that by default environment variables that are default when running process are quite useless, thus it is required to use this method when starting a process. 
This is specifically targeted at bazel-bsp that is started and later starts bazel server. Currently, bazel server has different environment in command line and IntelliJ, specifically JAVA_HOME and PATH which invalidates cache.

### DIFF
--- a/bsp/src/org/jetbrains/bsp/protocol/session/GenericConnector.scala
+++ b/bsp/src/org/jetbrains/bsp/protocol/session/GenericConnector.scala
@@ -1,8 +1,8 @@
 package org.jetbrains.bsp.protocol.session
 
 import java.io.File
-
 import ch.epfl.scala.bsp4j.BspConnectionDetails
+import com.intellij.util.EnvironmentUtil
 import org.jetbrains.bsp.protocol.session.BspServerConnector.{BspCapabilities, ProcessBsp}
 import org.jetbrains.bsp.protocol.session.BspSession.Builder
 import org.jetbrains.bsp.{BspBundle, BspError, BspErrorMessage}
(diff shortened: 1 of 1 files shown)


jastice (approved):
Thanks for finding this!

jastice:
Is this to be merged into 212.x intentionally?

lukaszwawrzyk:
It would be useful to have it in 212 just in case. For testing purposes and developing bsp in general for now it is ok to just have it in 213.x

jastice:
Changing the base branch has less pretty effects. Can you rebase on idea213.x?

github-sebastien-boulet:
We have similar issue with sbt. Any chance to get that quickly in a new release?

jastice:
> We have similar issue with sbt. Any chance to get that quickly in a new release?

Made an issue for that: https://youtrack.jetbrains.com/issue/SCL-19854

github-sebastien-boulet:
@jastice I have tested this PR on my Intellij and it fixes the issue for sbt too. The environment is correctly passed.

jastice:
> @jastice I have tested this PR on my Intellij and it fixes the issue for sbt too. The environment is correctly passed.

I assume you mean using sbt over BSP, not in the direct sbt support?

github-sebastien-boulet:
> I assume you mean using sbt over BSP, not in the direct sbt support?

Yes.

github-sebastien-boulet:
@jastice would it possible to push a new 213.x release with this fix? 🙏

jastice:
> @jastice would it possible to push a new 213.x release with this fix? 🙏

The nightly should contain it now

PR #603, opened 2021-12-22
ScalaBundle: remove redundant scalafmt message

kitbellew:
Instead of `scalafmt.config.load.errors.file.not.found`, which is not very informative and referenced only once, switch to the message which is used in all other cases: `scalafmt.can.not.find.config.file`.


kitbellew:
@unkarjedy fyi

kitbellew:
should i update this?

unkarjedy:
@kitbellew 
I will look into it.
I also saw a notification about your comment
>so even though that case class has correct hashcode and equals...

Was it intentionally deleted? I cna't find it on GitHub

kitbellew:
sorry, i deleted the comment when i realized it was not about whether the key works in the map, but whether the map can be serialized -- or restored from current cache which used a string...

kitbellew:
superseded by #615.

PR #605, opened 2022-01-09
Set Xtarget when using scala 3 with gradle.

SimY4:
Using scala 3 with Gradle and targetCompatibility. Leads to plugin failing to compile with: `bad option '-target:jvm-1.8' was ignored`

Since Scala 3 compiler flag `-target` was renamed to `-Xtarget` (source: https://docs.scala-lang.org/scala3/guides/migration/options-lookup.html)

This PR takes the Scala version into consideration for calculating target flag value.


jastice (approved):

jastice:
Thank you! Sorry for the delay; since it's no longer up to date with the current main branch, I manually patched it onto idea221.x:
fbac90a757085647b445a1394e393ee72d11f40d

PR #606, opened 2022-01-21
Avoid errors when importing javac opts through bsp

lukaszwawrzyk:
IntelliJ during fastpass import always reports an error for javac options and this confuses users that maybe import didn't work.
The thing is that target is either scala or java. If it is scala target, all info is fetched using `fetchScalacOptions`. Targets with scala may contain java. Anyway, they are reported with languageIds = ['scala', 'java']. This results in attempts to fetch compiler options twice (using scala and java) while it is enough to do it once.


jastice (approved):

jastice:
cherry-picked on 221.x: dc630a9a0b70a84ed53d7892624187a0dc673e8f

PR #612, opened 2022-04-16
Ignore non http(s) resolvers when downloading scalafmt

987Nabil:
This avoids errors while downloading scalafmt, when resolvers with non
http(s) protocols are defined (e.g. gcs or s3 urls) #SCL-19488

### DIFF
--- a/scala/scala-impl/src/org/jetbrains/plugins/scala/lang/formatting/scalafmt/dynamic/ScalafmtDynamicDownloader.scala
+++ b/scala/scala-impl/src/org/jetbrains/plugins/scala/lang/formatting/scalafmt/dynamic/ScalafmtDynamicDownloader.scala
@@ -54,8 +54,16 @@ object ScalafmtDynamicDownloader {
   ) extends DependencyManagerBase {
 
     // first search in the provided resolvers, then fallback to the default ones
-    override protected def resolvers: Seq[Resolver] =
-      extraResolvers ++ super.resolvers
+    // but only use resolvers with http(s) protocol, to avoid problems with private repositories.
+    // (like gcs, s3, etc.) This is fine since scalafmt should always be published to a public repo.
+    override protected def resolvers: Seq[Resolver] = {
+      val HttpUrl = "(https?://.*)".r
+      (extraResolvers ++ super.resolvers).filter {
+        case MavenResolver(_, HttpUrl(_)) => true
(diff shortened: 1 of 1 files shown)


scala/scala-impl/src/org/jetbrains/plugins/scala/lang/formatting/scalafmt/dynamic/ScalafmtDynamicDownloader.scala
```diff
@@ -54,8 +54,18 @@ object ScalafmtDynamicDownloader {
   ) extends DependencyManagerBase {
 
     // first search in the provided resolvers, then fallback to the default ones
+    // but only use resolvers with http(s) protocol, to avoid problems with private repositories.
+    // (like gcs, s3, etc.) This is fine since scalafmt should always be published to a public repo.
```
baldram, line 58: This showstopper bug has made our life with IntelliJ harder for a long time. Thanks for the fix proposal! 
LGTM!

A side note. Probably not a big deal here however, the pattern matching allows us to simplify it as follows:
```scala
    val validUrl = "http[s]{0,1}://.*".r

    override protected def resolvers: Seq[Resolver] =
      (extraResolvers ++ super.resolvers).filter {
        case MavenResolver(_, validUrl(_*)) => true
        case IvyResolver(_, validUrl(_*)) => true
        case _ => false
      }
```

987Nabil: I slightly changed the style, but I like the idea. 👍

baldram: All right, it looks good 👍

unkarjedy:
FTR:
This is a workaround for https://youtrack.jetbrains.com/issue/SCL-19488

unkarjedy:
Thank you for the PR!
I will manually cherry-pick it and close it.
I think the fix will be available in the next 2022.1 bugfix release soon (and in 2022.1 Nightly first)

unkarjedy:
cherry-picked: https://github.com/JetBrains/intellij-scala/commit/53e2af7b5d0bcddc339cba2ca073b8e4edca3e19

PR #613, opened 2022-04-20
SCL-20133: Fix type inferencing to Nothing

dwijnand:
https://youtrack.jetbrains.com/issue/SCL-20133

### DIFF
--- a/scala/scala-impl/src/org/jetbrains/plugins/scala/lang/psi/types/ScalaBounds.scala
+++ b/scala/scala-impl/src/org/jetbrains/plugins/scala/lang/psi/types/ScalaBounds.scala
@@ -41,6 +41,8 @@ trait ScalaBounds extends api.Bounds {
         case (_, ex: ScExistentialType) => glb(t1, ex.quantified, checkWeak).unpackedType
         case (lhs: ScTypePolymorphicType, rhs: ScTypePolymorphicType) =>
           polymorphicTypesBound(lhs, rhs, BoundKind.Glb, checkWeak)
+        case (lhs, _: WildcardType) => lhs
+        case (_: WildcardType, rhs) => rhs
         case _ => ScCompoundType(Seq(t1, t2), Map.empty, Map.empty)
       }
     }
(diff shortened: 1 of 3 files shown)


dwijnand:
I tried to demonstrate that this is the right behaviour, by using the Scala 3 behaviour, but it looks like the Scala 3 compiler is accidentally not symmetric:

```scala
sbt:scala3> repl
Welcome to Scala 3.1.3-RC1-bin-SNAPSHOT-nonbootstrapped-git-ee6733a (11.0.12, Java OpenJDK 64-Bit Server VM).
Type in expressions for evaluation. Or try :help.

scala> import dotty.tools.dotc.*, core.*, Contexts.*, Denotations.*, SymDenotations.*, Symbols.*, Types.*

scala> lazy val ictx = new ContextBase().initialCtx.fresh
lazy val ictx: dotty.tools.dotc.core.Contexts.FreshContext

scala> given Context = ictx.setSetting(ictx.settings.classpath, "/Users/dnw/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/org/scala-lang/scala-library/2.13.8/scala-library-2.13.8.jar:/d/scala3/out/bootstrap/scala3-library-bootstrapped/scala-3.1.3-RC1-bin-SNAPSHOT-nonbootstrapped/classes")
lazy val given_Context: dotty.tools.dotc.core.Contexts.Context

scala> ctx.initialize()

scala> TypeComparer.glb(WildcardType, defn.LongType)
val res0: dotty.tools.dotc.core.Types.Type = TypeRef(ThisType(TypeRef(NoPrefix,module class scala)),class Long)

scala> TypeComparer.glb(defn.LongType, WildcardType)
val res1: dotty.tools.dotc.core.Types.Type = WildcardType
```

Spoke with Martin and he agrees it should be and that it should favour, using the same, Long rather than WildcardType.

dwijnand:
Merged in with https://github.com/JetBrains/intellij-scala/commit/4c26d042bc6ddaaee53713f413c712a52b0e6f96

PR #614, opened 2022-04-22
Scalafmt config: show nested parse error message

kitbellew:
Fixes [SCL-20123.](https://youtrack.jetbrains.com/issue/SCL-20123)

### DIFF
--- a/scala/scala-impl/resources/messages/ScalaBundle.properties
+++ b/scala/scala-impl/resources/messages/ScalaBundle.properties
@@ -1241,7 +1241,6 @@ scalafmt.can.not.create.config.file=Can't create configuration file
 scalafmt.config.load.errors.failed.to.load.config=Failed to load scalafmt config
 scalafmt.config.load.errors.file.not.found=file not found
 scalafmt.config.load.errors.parse.error=parse error: {0}
-scalafmt.config.load.errors.unknown.error=unknown error: {0}
 scalafmt.config.load.errors.cyclic.includes.detected=cyclic includes detected
 scalafmt.config.load.actions.open.config.file=open config file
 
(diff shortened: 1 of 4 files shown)


kitbellew:
@unkarjedy PTAL, I can't test but fairly certain it should fix. Also, is 213.x still the right branch?

unkarjedy:
Thanks!
I will look into it and manually cherry-pick the changes

>Also, is 213.x still the right branch?
It's `idea221.x` currently

unkarjedy:
@kitbellew 
It still shows an "unknown error" prefix:
image

Why `org.scalafmt.dynamic.ScalafmtReflect#parseConfigFromString` 
returns `ScalafmtDynamicError.UnknownError` 
instead of `ScalafmtDynamicError.ConfigParseError`
like in `org.scalafmt.dynamic.ScalafmtReflect#parseConfig`
?
It looks like returning `ConfigParseError` would be the proper fix, or I do I miss anything?

Before this change
https://github.com/kitbellew/intellij-scala/commit/e394ed96e4c4edb1833d706d9ee33c01bbfa07ea#diff-7d252a166bcf0c809c86fc1ab4e6777f7cb4c007fd863a264782f860a55790e5L211
we treated any `ScalafmtConfigException` (currently non existing?) as `ConfigResolveError.ConfigParseError`

kitbellew:
of course, i can try to fix that on the formatter side but from this end, we need to find a solution which will work reasonably well with most versions.

unkarjedy:
JFTR we will need to update [`scalafmtVersion`](https://github.com/kitbellew/intellij-scala/blob/c76fedf08e7f1fde621d0d12d962dd6864bd343c/project/dependencies.scala#L119) when this is fixed on the formatter side

kitbellew:
@unkarjedy if this is a direct (not dynamic) dependency, then I'll just fix the underlying exception. will ping you with pr to take a look.

kitbellew:
superseded by #615 and scalameta/scalafmt#3192 fixes.

PR #616, opened 2022-04-28
Create SDKs per-module rather than per-project

steveniemitz:
When importing a project from BSP, different modules may have different java SDKs (java 8 vs java 11 for example).  This changes the import process to resolve (and importantly create if required) the SDK for each module.

PR #618, opened 2022-05-15
Limit maximum merged module name length

steveniemitz:
The module merger can end up generating module names that are so long they exceed the maximum file name length and can't be saved.  This limits them to a saner 150 (arbitrarily chosen) characters, plus an MD5 hash of all the target names to ensure they remain unique.

### DIFF
--- a/bsp/src/org/jetbrains/bsp/project/importing/BspResolverLogic.scala
+++ b/bsp/src/org/jetbrains/bsp/project/importing/BspResolverLogic.scala
@@ -10,6 +10,7 @@ import com.intellij.openapi.roots.DependencyScope
 import com.intellij.openapi.util.io.FileUtil
 import com.intellij.pom.java.LanguageLevel
 import org.apache.commons.codec.digest.DigestUtils
+import org.apache.commons.io.FileSystem
 import org.jetbrains.bsp.BspUtil._
 import org.jetbrains.bsp.data._
 import org.jetbrains.bsp.project.BspSyntheticModuleType
@@ -20,7 +21,6 @@ import org.jetbrains.plugins.scala.project.Version
 import org.jetbrains.plugins.scala.project.external.{JdkByHome, JdkByVersion}
 import org.jetbrains.sbt.project.data.{SbtModuleData, SbtModuleNode}
 import org.jetbrains.sbt.project.module.SbtModuleType
-
 import java.io.File
 import java.net.URI
 import java.nio.file.Paths
@@ -399,8 +399,6 @@ private[importing] object BspResolverLogic {
     data
   }
 
-  private final val MaxModuleNameLength = 150
-
   private[importing] def sharedModuleId(targets: Seq[BuildTarget]): String = {
     val upperCaseWords = """(?<!(^|[A-Z]))(?=[A-Z])""".r
     val pascalCaseWords = """(?<!^)(?=[A-Z][a-z])""".r
@@ -420,9 +418,9 @@ private[importing] object BspResolverLogic {
     }
     val ret = head.map(combine).mkString +
       (if (tail.nonEmpty) tail.map(combine).mkString("(", "", ")") else tail.mkString)
-    if (ret.length > MaxModuleNameLength) {
-      val prefix = ret.substring(0, MaxModuleNameLength)
-      val suffix = DigestUtils.md5Hex(targets.map(_.getDisplayName).mkString)
+    if (ret.length > FileSystem.getCurrent.getMaxFileNameLength) {
+      val suffix = DigestUtils.md5Hex(ret)
+      val prefix = ret.substring(0, FileSystem.getCurrent.getMaxFileNameLength - suffix.length)
       prefix + suffix
     } else {
       ret
--- a/bsp/src/org/jetbrains/bsp/project/importing/BspResolverLogic.scala
+++ b/bsp/src/org/jetbrains/bsp/project/importing/BspResolverLogic.scala
@@ -9,6 +9,7 @@ import com.intellij.openapi.module.StdModuleTypes
 import com.intellij.openapi.roots.DependencyScope
 import com.intellij.openapi.util.io.FileUtil
 import com.intellij.pom.java.LanguageLevel
+import org.apache.commons.codec.digest.DigestUtils
 import org.jetbrains.bsp.BspUtil._
 import org.jetbrains.bsp.data._
 import org.jetbrains.bsp.project.BspSyntheticModuleType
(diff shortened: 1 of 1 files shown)


bsp/src/org/jetbrains/bsp/project/importing/BspResolverLogic.scala
```diff
@@ -398,6 +400,15 @@ private[importing] object BspResolverLogic {
     data
   }
 
+  private def md5(input: String): String = {
```
unkarjedy: Consider using `org.apache.commons.codec.digest.DigestUtils#md5Hex(java.lang.String)`

bsp/src/org/jetbrains/bsp/project/importing/BspResolverLogic.scala
```diff
@@ -398,6 +400,15 @@ private[importing] object BspResolverLogic {
...
+    val digest = MessageDigest.getInstance("MD5")
+    digest.update(input.getBytes(StandardCharsets.UTF_8))
+    val result = digest.digest()
+    result.map("%02X".format(_)).mkString
+  }
+
+  private final val MaxModuleNameLength = 150
```
unkarjedy: We can use `org.apache.commons.io.FileSystem.getCurrent` + `org.apache.commons.io.FileSystem#getMaxFileNameLength` (it's 255 for all OS)

(we should remember to substract suffix length when calling `substring`)

abrams27:
is it possible that names arent unique? the name is a path + suffix (names separated with `+` if im not mistaken), paths are unique in principle, so these names should also be unique

for bazel ofc xd

bsp/src/org/jetbrains/bsp/project/importing/BspResolverLogic.scala
```diff
@@ -415,8 +426,15 @@ private[importing] object BspResolverLogic {
       val nonEmptyParts = parts.filter(_.nonEmpty)
       if (nonEmptyParts.size > 1) nonEmptyParts.mkString("(", "+", ")") else nonEmptyParts.mkString
     }
-    head.map(combine).mkString +
+    val ret = head.map(combine).mkString +
       (if (tail.nonEmpty) tail.map(combine).mkString("(", "", ")") else tail.mkString)
+    if (ret.length > MaxModuleNameLength) {
+      val prefix = ret.substring(0, MaxModuleNameLength)
+      val suffix = md5(targets.map(_.getDisplayName).mkString)
```
unkarjedy: I am maybe a little out of the BSP context, but can't we just call md5 on `ret` itself?

steveniemitz:
> is it possible that names arent unique? the name is a path + suffix (names separated with `+` if im not mistaken), paths are unique in principle, so these names should also be unique
> 
> for bazel ofc xd

Sorry I'm not sure what your question is exactly?  The hashed version of the names should be just as unique as the full text, the chance of a hash collision here is incredibly small.

abrams27:
so i was wondering if there is a case where the hash would be needed (is there a case where the module name is not unique), but I forgot that you are trimming the name now, so it is possible

steveniemitz:
> so i was wondering if there is a case where the hash would be needed (is there a case where the module name is not unique), but I forgot that you are trimming the name now, so it is possible

ah, it's needed because the name is getting trimmed down yeah.

bsp/src/org/jetbrains/bsp/project/importing/BspResolverLogic.scala
jastice: Just wondering if we can make this a little smarter so as to not have 2 very similar module names, or is it a rare enough occurrence not to matter in practice?

steveniemitz: if you have any suggestions for making it smarter I'm all ears. :P

It really shouldn't be very common though, I only ran into this because we have a pretty weird edge case in our build system that causes many modules to all contain the same source root.

bsp/src/org/jetbrains/bsp/project/importing/BspResolverLogic.scala
steveniemitz: I mean at that point, does 150 vs 255 characters really matter?  It doesn't need to be exact here, once you're past like 100 the name is pretty meaningless.

bsp/src/org/jetbrains/bsp/project/importing/BspResolverLogic.scala
steveniemitz: I don't see why that wouldn't work, nor why it would be any better than what its doing now.

bsp/src/org/jetbrains/bsp/project/importing/BspResolverLogic.scala
unkarjedy: It's of cause not a major.
There is just no need in an extra constant.
Code readers and future maintainers will question: "what is this 150, why it is 150"?
Using `getMaxFileNameLength` will clearly show the intention and cause.

bsp/src/org/jetbrains/bsp/project/importing/BspResolverLogic.scala
unkarjedy: > why it would be any better than what its doing now.

1. Ret is already calculated, no extra calculation is required.
2. Again, code maintainers will question "why using `targets.map(_.getDisplayName).mkString`, is it some special magic here"? If we can avoid redundant stuff we should do it.

jastice: I could imagine truncating common substrings, but that complexity would only be worth it if it's common enough. If it's easy though, I'd suggest only adding the hash if truncating would otherwise lead to name duplications, to avoid cryptic looking module names.

steveniemitz: that seems more complicated than worth it for this kind of edge case.  If you feel like adding that in later feel free to, but it's out of the scope of the amount of effort I'm willing to put into this fix.

unkarjedy: It's already better then it was when It didn't work at all

jastice (approved):

jastice:
Thanks! I'll merge it now, unless you were going to add something, @steveniemitz ?

steveniemitz:
Nope good on my side, thank you!

jastice:
also cherrypicked to `idea222.x`: 2d84da

steveniemitz:
thank you!

steveniemitz:
I noticed this (and #616) wasn't in the 2022.1.16 release, do I need to submit a PR to add it to the idea221.release branch too?

jastice:
If a fix is important to you I can also cherrypick it on 221 :)

steveniemitz:
> If a fix is important to you I can also cherrypick it on 221 :)

That'd be amazing, thank you :D

PR #623, opened 2022-07-07
SCL-20405: Allow path of the structure file to be configured, and als…

akamali:
Allow path of the structure file to be configured, and also the flags to be externally controllable instead only through internal UI.

### DIFF
--- a/scala/scala-impl/src/org/jetbrains/sbt/project/SbtProjectResolver.scala
+++ b/scala/scala-impl/src/org/jetbrains/sbt/project/SbtProjectResolver.scala
@@ -219,26 +219,41 @@ class SbtProjectResolver extends ExternalSystemProjectResolver[SbtExecutionSetti
       Failure(new UnsupportedOperationException(message))
     }
     else {
-      val structureFileReused = new File(FileUtil.getTempDirectory) / s"sbt-structure-reused-${projectRoot.getName}.xml"
-      if (RegistryManager.getInstance().is("sbt.project.import.reuse.previous.structure.file")){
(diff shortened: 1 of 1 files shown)


jastice (approved):
Thanks, this looks useful. I'd worry that it's underdocumented, but mid-term we might deprecate the sbt-structure support anyway. So for future-proofing, I'd suggest to consider adding some kind build structure precomputation and caching to sbt and its BSP server itself.

akamali:
@jastice Thanks for taking a look and your feedback. I'm happy to add some comments to the code, I'll get to it tonight or tomorrow. 

As for future proofing I'm afraid your comments are beyond my sbt and BSP understanding, I can't say I know a ton about either. I'd imagine caching in sbt itself isn't too complicated given that most things should already be serializable, but not sure what this would really look like.

Another thing to have in mind for caching in sbt is the time itself to start up sbt itself. On our (ancient) version of sbt it takes around 2-3 minutes just to get to the sbt shell (many modules, sbt resolves +170k keys/settings). Caching in sbt would mean we would still have to pay that start up price.

jastice:
I've already started the internal merging process, no need to update this PR.

> Caching in sbt would mean we would still have to pay that start up price.

I see. In that case, maybe it's worth addressing this on BSP client side. But maybe with newer sbt versions it will be less of a problem anyway. Which version of sbt are you using?

akamali:
> Which version of sbt are you using?

0.13.18 😅

akamali:
@jastice Please cherry-pick to idea221.x if possible.

PR #624, opened 2022-07-11
Changed getCanonicalPath to custom toCanonicalPath

mikkoziel:
Using method getCanonicalPath was very aggravating for CPU and made imports last longer. I added custom toCanonicalPath implicit method that only take AbsolutePath and normalize it which I think is enough and which makes imports faster(for example: before change - ca.11min, after change- ca.3min)

### DIFF
--- a/bsp/src/org/jetbrains/bsp/project/importing/BspResolverLogic.scala
+++ b/bsp/src/org/jetbrains/bsp/project/importing/BspResolverLogic.scala
@@ -515,6 +515,13 @@ private[importing] object BspResolverLogic {
       case (first, _) => first
     }
 
+  // extension method added as for better performance than getCanonicalPath method
+  implicit class FileExtensions(val f: File){
+    def getCanonicalPathSimplified: String = {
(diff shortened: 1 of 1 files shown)


jastice (changes requested):
Thank you very much, great find!
I just worry that `toCanonicalPath` is too easily confused with `getCanonicalPath`, maybe we can find another name?
And please add a short comment with the motivation for the method

Also, can you please rebase on `idea222.x`, we'll then cherrypick to lower versions as needed.

mikkoziel:
Changed name to getCanonicalPathSimplified and added comment

mikkoziel:
I rebased the commit. All should be all right now.
And we will need the cherrypick to version idea221.x.

jastice (approved):
Thanks!

jastice:
rebased and merged:
- v221.x: dba55790304b00723b61a3c39e09ceb02f0ae725
- v222.x: 2b10cef3be2c6fd4a4805d2c1f6c7700344af8e2

PR #625, opened 2022-07-20
Add json, html, js to interpolated injection mappings

bplommer:
issue: https://youtrack.jetbrains.com/issue/SCL-16050/language-injection-via-interpolation-add-default-interpolators-for-htmljsonjsetc

I've manually tested that this works as expected for json and html - I wasn't able to manually test for JS because it isn't supported by community edition. Unfortunately I've had some difficulties with getting tests to run locally.


unkarjedy:
Thanks for the PR, I will try to merge it to the 2022.2 Nightly this week.

Could you please give more details on what difficulties you faced during local tests run?
We will try to fix/improve the process
>I've had some difficulties with getting tests to run locally

unkarjedy:
Merged manually into idea222.release & idea222.x
The Nightly version will be available tomorrow

PR #628, opened 2022-08-18
Fix off-by-one error in line numbers

retronym:
We need a zero-based line number in this context.

### DIFF
--- a/scala/scala-impl/src/org/jetbrains/plugins/scala/debugger/LocationLineManager.scala
+++ b/scala/scala-impl/src/org/jetbrains/plugins/scala/debugger/LocationLineManager.scala
@@ -30,7 +30,7 @@ trait LocationLineManager {
 
   def exactLineNumber(location: Location): Int = location match {
     case gen: GeneratedLocation =>
-      gen.lineNumber()
+      ScalaPositionManager.checkedLineNumber(gen)
     case _ =>
       checkAndUpdateCaches(location.declaringType())
       customizedLocationsCache.getOrElse(location, ScalaPositionManager.checkedLineNumber(location))


retronym:
See https://github.com/JetBrains/intellij-scala/commit/d6665c8be8e7d0ddf613c096143bcd7f063a63a7#r81569136

unkarjedy:
@retronym 
Thanks! 
Just in case, do you already have any example when it leads to bug in the debugger?
Would be nice to add it to our tests

retronym:
I was notified of the problem by someone else and all I did was find the
place in the plug-in code that treats GeneratedPosition differently.

I have asked them for any repro details, will add them here if/when they
get back to me. My hunch is that The plug-in is using GeneratedPosition for
them because it can’t find a true JDI position as it doesn’t know about the
synthetic wrapper classes added by custom compliler phases. So maybe they
see GeneratedPosition more often than other users they might run into this.
Again, this is just a hunch…

nafg:
I guess you're not talking about me, and this might be unrelated, but here's my experience:

tldr version: `idea --line $y --column $x $file` puts the cursor on column `x + 1`.

Longer story:

In Tilix I have a custom link that for regex `(\/\S*\.[\w]+):([0-9]*):([0-9]*)` it runs `/home/naftoli/bin/idea-auto --line $2 --column $3 $1` which currently basically runs the Toolbox-generated script with the same arguments.

Now if I create a an error like this:

```scala
        availabilities.collect1 {
```

the compiler prints

```text
[error] /home/naftoli/repositories/gitlab.com/chesednow/disp/request_views_shared/src/main/scala/chesednow/requests/views/shared/RequestValidations.scala:52:24: value collect1 is not a member of Seq[(chesednow.shared.common.LKEnt[chesednow.equipment.shared.EquipmentItem.Full], chesednow.equipment.shared.Availability)]
[error] did you mean collect?
[error]         availabilities.collect1 {
[error]                        ^
```

and if I ctrl-click the hyperlinked part (`/home/naftoli/repositories/gitlab.com/chesednow/disp/request_views_shared/src/main/scala/chesednow/requests/views/shared/RequestValidations.scala:52:24`) it goes to column 25, _after_ the `c` of `collect1`.

retronym:
@nafg I'm not sure if the code that handles the CLI integration is affected by this bug. I would presume it is not related, because the code I'm touching should only be relevent to the debugger, AFAIK.

@unkarjedy I've just got some additional information about the original user that observed the problem. They have a custom IntelliJ extension that is contributing async stack traces. It instantiates `GeneratedPosition`, apparently to have control over the presented method name avoiding a mangled name.

AFAICT this is valid usage of `GeneratedPosition` and this PR is still valid, it seems clear enough to me that https://github.com/JetBrains/intellij-scala/commit/d6665c8be8e7d0ddf613c096143bcd7f063a63a7#r81569136 was accidentally behaviour changing.

ghost:
Sorry, I was on vacation last week. I think this is a good change. I will use our internal CI in order to run all tests.

PR #629, opened 2022-08-26
CopyFromHistoryViewerAction: use correct shortcut set, #SCL-14369 fixed

azolotko:


unkarjedy:
Thanks for the fix!
For me it doesn't work on windows though.

I decided to completely remove SigIntAction because it's not that actual 
after https://github.com/JetBrains/intellij-scala/pull/630
We already have a dedicated action to stop the shell.
(under the hood it uses the same mechanism actually)
After the change it works on Windows as well.


Thanks for the fix!
I will manually cherry-pick the changes and include it into next 2022.2

azolotko:
@unkarjedy For me, the copy action is broken again after https://github.com/JetBrains/intellij-scala/commit/ed7a1d8cd40c2a1b060f20c316f781a0327726bc. As well as the find action.

unkarjedy:
@azolotko 
Sorry, the fixes were not merged to final branches.
New 2022.2 Nightly & EAP versions with the fix should be uploaded today or tomorrow

PR #630, opened 2022-08-26
SBT StartAction and DebugShellAction: copy shortcut sets from default actions, #SCL-15583 fixed

azolotko:


unkarjedy:
Thanks for the fix!
I will manually cherry-pick the changes and include it into next 2022.2

PR #621, opened 2022-06-28
Limit maximum merged module name length to less than MaxFileNameLength

steveniemitz:
#618 attempted to fix this, but during the PR process we changed a 150 character limit to the system max file name length.  Unfortunately this causes issues, IJ may attempt to append `(shared)` to the file name, putting it back over the max limit.

This change adds a 50 character buffer in to allow extra modifications to the file name once created.

PR #632, opened 2022-09-22
Limit maximum merged module name length to less than MaxFileNameLength

steveniemitz:
This was merged in #621 but never made it into idea222.release.  I'm not sure if this is the right way to get it in this branch, let me know if I should do something else, thanks!

### DIFF
--- a/bsp/src/org/jetbrains/bsp/project/importing/BspResolverLogic.scala
+++ b/bsp/src/org/jetbrains/bsp/project/importing/BspResolverLogic.scala
@@ -399,6 +399,8 @@ private[importing] object BspResolverLogic {
     data
   }
 
+  private final val MaxFileNameLength = FileSystem.getCurrent.getMaxFileNameLength - 50
+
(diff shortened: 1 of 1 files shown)


jastice:
The changes do appear to be in [`idea222.release`](https://github.com/JetBrains/intellij-scala/blob/b952a5b6f26733b321706a30ce47c9a1077ae985/bsp/src/org/jetbrains/bsp/project/importing/BspResolverLogic.scala#L402-L423)

steveniemitz:
Hm, you're right they are, https://github.com/JetBrains/intellij-scala/tree/2022.2.13 was created after though, shouldn't that commit be there?  Are those tags not being created off the release branch?

jastice:
The tag is on the release branch, too. I can't quite explain the situation 😓

steveniemitz:
yeah that's super weird, but its in 2022.2.16 now so I'm happy!  Thank you!

PR #633, opened 2022-09-22
Use sbt 1.5 syntax for dumpStructureTo

akozich:
With sbt 1.5+ I started receiving warnings while reloading sbt project in Idea

`[warn] sbt 0.13 shell syntax is deprecated; use slash syntax instead: Global / dumpStructureTo`

This happens because sbt 1.5 deprecated sbt 0.13 syntax with : separator

Here's the ticket [SCL-18893](https://youtrack.jetbrains.com/issue/SCL-18893)

### DIFF
--- a/scala/scala-impl/src/org/jetbrains/sbt/project/structure/SbtStructureDump.scala
+++ b/scala/scala-impl/src/org/jetbrains/sbt/project/structure/SbtStructureDump.scala
@@ -64,7 +64,7 @@ class SbtStructureDump {
         ideaPort.fold("")(port => s"; set ideaPort in Global := $port")
       } else ""
 
-    val cmd = s";reload; $setCmd ;${if (preferScala2) "preferScala2;" else ""}*/*:dumpStructureTo $structureFilePath; session clear-all $ideaPortSetting"
+    val cmd = s";reload; $setCmd ;${if (preferScala2) "preferScala2;" else ""}*/*/dumpStructureTo $structureFilePath; session clear-all $ideaPortSetting"
     val aggregator = shellMessageAggregator(EventId(s"dump:${UUID.randomUUID()}"), shell, reporter)
 
     shell.command(cmd, BuildMessages.empty, aggregator)


jastice (approved):

unkarjedy:
The change breaks interop with sbt 0.13.
This is the output when sbt shell is used:
>Extracting structure failed: Structure file ~/AppData/Local/Temp/sbt-structure.xml is empty
Expected symbol
Not a valid command: */*/
*/*/dumpStructureTo ~/AppData/Local/Temp/sbt-structure.xml
^

Note, that even though sbt 0.13 is quite old, there were no decision to drop it's support.
There are still existing/legacy projects which use sbt 0.13

unkarjedy:
So we need to conditionally select the syntax depending on the sbt version.
One way to know sbt version before launching sbt is probably too lookup `project/build.properties` and read `sbt.version` value.
If there is no explicit version we can use sbt 1.x syntax when bundled sbt-launcher is used (which is 1.6.2 in 2022.2).
Though we should remember that a custom sbt-launch can be used, which also might be some old launcher (0.13).
In this case to be safe we should fallback to the 0.13 syntax

image

jastice:
I think it would be safest to expose the sbt version known at launch time in `SbtProcessManager` OR obtain it via `sbtVersion` command at some point after launch in `SbtShellCommunication`

in particular [these lines](https://github.com/JetBrains/intellij-scala/blob/314a4052714824a4db4f639473fe70f630b3be54/scala/scala-impl/src/org/jetbrains/sbt/shell/SbtProcessManager.scala#L102-L135) would be relevant

unkarjedy:
@jastice 
It would be the safest.
Though it would require some principal changes in command sending.
AFAIU currently command is written to sbt process unconditionally.
If we detect sbt version at runtime we would need to first wait for some event which will provide us sbt version and only then write the command.

danicheg:
> [warn] sbt 0.13 shell syntax is deprecated; use slash syntax instead: Global / dumpStructureTo

Although it's a minor thing, it's 2024, but this still catches the eye. I'm using bundled sbt launcher (`v1.9.x`) and plugin `2023.2.23`.

unkarjedy:
FTR: https://youtrack.jetbrains.com/issue/SCL-18893/SBT-1.5.0-deprecates-in-syntax-in-sbt-files

PR #634, opened 2022-10-07
add comparison of test module and run configuration module

cobr123:
When there are tests with the same name in different modules, IDE uses the same run configuration, which leads to the launch of the wrong tests.

Example project https://github.com/cobr123/tests_with_same_name_in_two_modules


unkarjedy:
Thanks for the fix!
I manually merged the fix into the latest branch `idea231.x`
The fix will be available in the next 2023.1 Nightly version.

PR #635, opened 2022-10-19
Enable BSP tracing only when selected

Duhemm:
Previously, BSP tracing would only be enabled, but the tracer would not write to disk when BSP tracing is de-selected in the IDE settings.

Enabling BSP tracing when the BSP messages are very large would considerably slow down the import process, or completely prevent it, because the BSP traces would still be produced, only to be later ignored by the log writer.

This commit completely disables BSP tracing when the option is de-selected in the IDE settings.

----

While working on this patch, I've been importing via Bloop a group of 714 targets having about 1.6M classpath entries. The import was getting stuck for 30 minutes on the `scalac options` step of BSP import, or the import would outright fail. With this patch, the entire `resolving BSP build structure` group of steps finishes in 1m50s for the same import.

If possible, a back port of this patch to 2022.2 would be greatly appreciated.

### DIFF
--- a/bsp/src/org/jetbrains/bsp/protocol/session/BspSession.scala
+++ b/bsp/src/org/jetbrains/bsp/protocol/session/BspSession.scala
@@ -122,26 +122,21 @@ class BspSession private(bspIn: InputStream,
     }
   }
 
-  private def bspTraceLogger: PrintWriter = {
-    val logfile = sys.env.get("BSP_TRACE_PATH")
-      .orElse(sys.props.get("BSP_TRACE_PATH"))
-      .map(new File(_))
-      .getOrElse({
-        val dirs = Paths.get(PathManager.getLogPath, "bsp")
-        Files.createDirectories(dirs)
-        val stamp = DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH-mm-ss").format(LocalDateTime.now())
-        new File(dirs.toFile, s"bsp-protocol-trace-$stamp.log")
-      })
-    new PrintWriter(lazyFileCreateWriter(logfile)) {
-
-      override def println(x: Any): Unit =
-        if (traceLogPredicate())
-          super.println(x)
-
-      override def flush(): Unit =
-        if (traceLogPredicate())
-          super.flush()
-
+  private def bspTraceLogger: Option[PrintWriter] = {
+    if (traceLogPredicate()) {
+      val logfile = sys.env.get("BSP_TRACE_PATH")
+        .orElse(sys.props.get("BSP_TRACE_PATH"))
+        .map(new File(_))
+        .getOrElse({
+          val dirs = Paths.get(PathManager.getLogPath, "bsp")
+          Files.createDirectories(dirs)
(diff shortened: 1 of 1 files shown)


jastice:
Oh wow, good find, that's quite the impact!

I am confused though - where does the time go when tracing is disabled? It wouldn't be the actual printing and writing of the trace.

The only problem I see with this PR is that to enable or disable tracing, the BSP session would need to be restarted. Do you see a way to avoid this?

Duhemm:
Thanks!

> It wouldn't be the actual printing and writing of the trace.

I think it is. In the import I mentioned in the PR description, the response to `build/scalacOptions` is large: 289MB. The entire message is converted to a string and included into another string: 
https://github.com/eclipse/lsp4j/blob/32aac26422a93019c197f6f7d92349bee609f2d9/org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/TracingMessageConsumer.java#L178-L182
I don't know how else to explain the difference. In a debugger, the `toString` never returned. I haven't measured how long that step takes in isolation.

I'm not sure how to keep the toggle on tracing 🤔 I think this would need extending or overriding the launcher's `Builder`, the `MessageTracer`, the `TracingMessageConsumer`, ...

I'll look into it. Do you think this is a must-have?

jastice:
> I think it is. In the import I mentioned in the PR description, the response to build/scalacOptions is large: 289MB. The entire message is converted to a string and included into another string:
https://github.com/eclipse/lsp4j/blob/32aac26422a93019c197f6f7d92349bee609f2d9/org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/TracingMessageConsumer.java#L178-L182

Oh, I see, so it creates the String even before logging it. That's unfortunate.

> I'm not sure how to keep the toggle on tracing 🤔 I think this would need extending or overriding the launcher's Builder, the MessageTracer, the TracingMessageConsumer, ...

It does look like that to me, pretty inconvenient. Let's say if we don't find a viable way around it by tomorrow I'll merge this, given the large impact.

I do wonder why this particular stringification is so expensive when it also happens on the protocol itself, any idea?

Duhemm:
> I do wonder why this particular stringification is so expensive when it also happens on the protocol itself, any idea?

My guess is that it's pretty printing that makes it a lot slower (I still haven't gotten to measuring it). Unfortunately, I don't see a way to disable pretty printing without replacing the `TracingMessageConsumer`, because it's calling a static function of `MessageJsonHandler` to pretty print the message.

jastice:
I suppose that might be it, even though it's a bit surprising. Anyway, I will merge this then.

Duhemm:
Thanks! Would you like me to cherry pick that on top of `222.x`?

jastice:
I'll handle that

jastice:
What surprises me a bit, doesn't the trace logging on bazel-bsp  have a similar performance impact? Did you check that?

abrams27:
> What surprises me a bit, doesn't the trace logging on bazel-bsp have a similar performance impact? Did you check that?

>>  I've been importing via Bloop

so bazel-bsp is not involved

PR #637, opened 2022-11-10
SCL-20726: check if project is already disposed

Undin:


ghost:
Thanks for opening the PR. There might be other issues connected to this component, I will examine in more detail.

ghost:
https://youtrack.jetbrains.com/issue/SCL-20726 was resolved by reverting the commit that caused it.

PR #638, opened 2023-02-06
SCL-20936 add sbt parameters to the settings pane

azdrojowa123:
[YouTrack Issue](https://youtrack.jetbrains.com/issue/SCL-20936/Please-allow-sbt-launcher-command-line-parameters-in-addition-to-VM-parameters)
Sbt command line parameters in most cases are translated to the JVM system properties. 
Part of them that is responsible for logging level, is passed directly to the sbt-laucnher.jar.


unkarjedy:
The review is closed in favor of Space review

PR #639, opened 2023-02-07
Set Terminal start directory to VCS root for BSP projects

lukaszwawrzyk:
If the IJ project root does not match workspace git root, the terminal starting directory will be set to IJ project root, which is usually not what users want and they need to change the directory each time. They open a new terminal, or they need to change their terminal settings manually. 
Bsp project can be created outside of the project root or in a subdirectory of it, but to invoke build tool from terminal or interact with git, users want to be at their workspace root in the terminal.

### DIFF
--- a/bsp/resources/META-INF/BSP-Terminal.xml
+++ b/bsp/resources/META-INF/BSP-Terminal.xml
@@ -1,0 +1,5 @@
+<idea-plugin>
+    <extensions defaultExtensionNs="com.intellij">
+        <bspVcsRootExtension implementation="org.jetbrains.bsp.data.TerminalStartDirectoryConfig"/>
+    </extensions>
+</idea-plugin>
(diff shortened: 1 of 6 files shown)


jastice (approved):

unkarjedy:
@lukaszwawrzyk 
Are you planning to implement the newly created extension point `org.jetbrains.bsp.data.BspVcsRootExtension` somewhere outside Scala Plugin?
If not then there is no point in extension point and we should replace it with a more direct implementation.

lukaszwawrzyk:
> @lukaszwawrzyk Are you planning to implement the newly created extension point `org.jetbrains.bsp.data.BspVcsRootExtension` somewhere outside Scala Plugin? If not then there is no point in extension point and we should replace it with a more direct implementation.

The extension point requires a dependency on terminal plugin which doesn't need to be on, so to avoid errors, the EP is only added when the plugin is installed. For this I need the EP mechanism.

unkarjedy:
Ah, right, indeed.
We at least should mark such extension points as `@ApiStatus.Internal` (I will do it)

PR #646, opened 2023-04-26
Use exact source roots from BSP where possible

lukaszwawrzyk:
Reasoning is explained in the code comments.
Generally, if someone uses bazel and follows maven structure and keeps a target per standard source set, it will look more natural. The real problem that I am solving here is that something that should be 1 source set becomes 5 source sets, randomly assigned based on enclosing directory and it is inconvenient to run all tests in given module from IntelliJ.

### DIFF
--- a/bsp/src/org/jetbrains/bsp/project/importing/BspResolverLogic.scala
+++ b/bsp/src/org/jetbrains/bsp/project/importing/BspResolverLogic.scala
@@ -117,7 +117,7 @@ private[importing] object BspResolverLogic {
       .map(item => (item.getTarget, item.getOutputPaths.asScala.iterator.map(_.getUri.toURI.toFile).toSeq))
       .toMap
 
-    // Source roots (repoted by BSP) that are not shared with other targets
+    // Source roots (reported by BSP) that are not shared with other targets
     val exclusiveSourceRoots = sourcesItems
       .flatMap(item => item.getRoots.asScala.map(_ -> item.getTarget))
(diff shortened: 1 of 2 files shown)


lukaszwawrzyk:
Also, I run formatter on this file accidentally. I can revert if it is important, but I guess these formatting changes make a lot of sense.

azdrojowa123 (approved):

bsp/src/org/jetbrains/bsp/project/importing/BspResolverLogic.scala
```diff
@@ -110,15 +110,23 @@ private[importing] object BspResolverLogic {
...
     val idToResources = resourcesItems
-      .map(item => (item.getTarget, item.getResources.asScala.iterator.map(sourceDirectory(_)).toSeq))
+      .map(item => (item.getTarget, item.getResources.asScala.iterator.map(sourceDirectory(_)).distinct.toSeq))
       .toMap
 
     val idToOutputPaths = outputPathsItems
       .map(item => (item.getTarget, item.getOutputPaths.asScala.iterator.map(_.getUri.toURI.toFile).toSeq))
       .toMap
 
+    // Source roots (repoted by BSP) that are not shared with other targets
```
odisseus: Typo

lukaszwawrzyk: fixed

unkarjedy:
Merged manually into idea231.x.

PR #650, opened 2023-11-30
migrate scala sdk extension impl from `intellij-bsp`

xuansontrinh:
what were done:

- register an additional synthetic scala library and convert it to a scala sdk by internal scala plugin API.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -52,6 +52,7 @@ lazy val scalaCommunity: sbt.Project =
       debugger % "test->test;compile->compile",
       testingSupport % "test->test;compile->compile",
       devKitIntegration % "test->test;compile->compile",
+      intellijBspIntegration % "test->test;compile->compile",
       gradleIntegration % "test->test;compile->compile",
       intelliLangIntegration % "test->test;compile->compile",
       mavenIntegration % "test->test;compile->compile",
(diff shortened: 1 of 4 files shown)


scala/integration/intellij-bsp/src/org/jetbrains/plugins/scala/project/bsp/ScalaSdkGetter.scala
```diff
@@ -0,0 +1,33 @@
...
+import org.jetbrains.magicmetamodel.impl.workspacemodel.impl.updaters.transformers.{ScalaUtil => BspScalaUtil}
+import org.jetbrains.plugins.bsp.extension.points.ScalaSdkGetterExtension
+import org.jetbrains.plugins.bsp.server.tasks.ScalaSdk
+import org.jetbrains.plugins.scala.project.ScalaLanguageLevel
+import org.jetbrains.plugins.scala.project.external.ScalaSdkUtils
+
+import java.net.URI
+import java.nio.file.Paths
+
+class ScalaSdkGetter extends ScalaSdkGetterExtension {
```
unkarjedy: I haven't seen the sources of BSP plugin, but the name `ScalaSdkGetterExtension` looks confusing for something which doesn't "gets scala sdk".
The logic seems to be about "registering Scala SDK" (in the module?) or "adding scala sdk to module".
Both "register" and "add" sound kinda opposite to "get"
I would suggest renaming it in the BSP plugin

unkarjedy:
Note that you would need to duplicate the changes in idea241.x branch once you publish BSP plugin for 2024.1

azdrojowa123:
as this comment above will be corrected LGTM

azdrojowa123 (approved):

unkarjedy:
JFTR, the change will only be merged into idea233.x and won't be available in idea241.x automatically, because the BSP plugin is not published for IDEA 2024.1 yet

PR #653, opened 2024-01-21
Refactoring an anonymous class to inner class #SCL-227

leonbur:
This commit fixes this issue

https://youtrack.jetbrains.com/issue/SCL-227

**UPD**
Related thread in the discord 
https://discord.com/channels/931170831139217469/931171260824707072/1198202206260834344

### DIFF
--- a/scala/scala-impl/resources/messages/ScalaBundle.properties
+++ b/scala/scala-impl/resources/messages/ScalaBundle.properties
@@ -1804,6 +1804,8 @@ move.members=Move members
 move.members.supported.only.stable.objects=Move refactoring is supported only for stable object members
 move.members.not.supported.implicits=Move refactoring is not supported for implicit definitions
 move.members.not.supported.overridden=Move refactoring is not supported for overridden definitions
+move.anonymousToInner.name=Convert Anonymous to Inner\u2026
+move.anonymousToInner.dialog.title=Convert Anonymous to Inner
 
 ### org/jetbrains/plugins/scala/lang/refactoring/rename/RenameScalaClassProcessor.scala
 rename.companion.module=Rename companion {0}
--- a/scala/scala-impl/src/org/jetbrains/plugins/scala/lang/refactoring/move/anonymousToInner/ScalaAnonymousToInnerDialog.scala
+++ b/scala/scala-impl/src/org/jetbrains/plugins/scala/lang/refactoring/move/anonymousToInner/ScalaAnonymousToInnerDialog.scala
@@ -1,0 +1,121 @@
+package org.jetbrains.plugins.scala.lang.refactoring.move.anonymousToInner
+
+import com.intellij.java.refactoring.JavaRefactoringBundle
+import com.intellij.openapi.project.Project
+import com.intellij.openapi.ui.DialogWrapper
+import com.intellij.psi.{PsiManager, PsiNameHelper}
+import com.intellij.refactoring.HelpID
+import com.intellij.refactoring.ui.NameSuggestionsField
+import com.intellij.refactoring.util.{CommonRefactoringUtil, ParameterTablePanel, RefactoringMessageUtil, VariableData}
+import com.intellij.ui.IdeBorderFactory
+import com.intellij.util.ui.FormBuilder
+import org.jetbrains.plugins.scala.ScalaBundle
+import org.jetbrains.plugins.scala.lang.psi.api.ScFile
+import org.jetbrains.plugins.scala.lang.psi.api.toplevel.templates.ScExtendsBlock
+import org.jetbrains.plugins.scala.lang.psi.api.toplevel.typedef.ScTemplateDefinition
+import org.jetbrains.plugins.scala.lang.psi.types.ScParameterizedType
+import org.jetbrains.plugins.scala.lang.psi.types.api.designator.ScDesignatorType
+import org.jetbrains.plugins.scala.lang.refactoring.extractMethod.ScalaVariableData
+
+import java.awt.BorderLayout
+import javax.swing.{JComponent, JPanel}
+import scala.jdk.CollectionConverters._
+
+/**
+ * @note original implementation was inspired by [[com.intellij.refactoring.anonymousToInner.AnonymousToInnerDialog]]
+ */
+class ScalaAnonymousToInnerDialog(project: Project, extendsBlock: ScExtendsBlock, variables: Array[ScalaVariableData], target: Either[ScFile, ScTemplateDefinition]) extends DialogWrapper(project) {
+
+  setTitle(ScalaBundle.message("move.anonymousToInner.dialog.title"))
+
+  private val classNameField: NameSuggestionsField = new NameSuggestionsField(project)
+  classNameField.setSuggestions(suggestNewClassNames())
+  classNameField.selectNameWithoutExtension()
+
+  init()
+
+
+  private def suggestNewClassNames(): Array[String] =
+    extendsBlock.superTypes.head match {
+      case pType: ScParameterizedType =>
+        val name = pType.designator.toString
+        val typeParameters = pType.typeArguments.map(_.toString).mkString + name
+        Array(typeParameters, "My" + name)
+      case designator: ScDesignatorType =>
+        val name = designator.toString
+        Array("My" + name)
+    }
+
+  override protected def doOKAction(): Unit = {
+    var errorString: String = null
+    val innerClassName = getClassName
+    val manager = PsiManager.getInstance(project)
+    if (innerClassName.isEmpty) errorString = JavaRefactoringBundle.message("anonymousToInner.no.inner.class.name")
+    else {
+      if (!PsiNameHelper.getInstance(manager.getProject).isIdentifier(innerClassName)) {
+        errorString = RefactoringMessageUtil.getIncorrectIdentifierMessage(innerClassName)
+      }
+      else {
+        val existingClassesAndObjects = target.fold(
+          file =>
+            file.getClassNames.asScala.toSet,
+          template =>
+            template.extendsBlock
+              .getOrCreateTemplateBody
+              .getChildren
+              .collect { case c: ScTemplateDefinition => c.name }
+              .toSet
+        )
+
+        if (existingClassesAndObjects.contains(innerClassName)) {
+          errorString = JavaRefactoringBundle.message("inner.class.exists", innerClassName, target.map(_.name).getOrElse("TOP LEVEL DEFINTION"))
+        }
+      }
+    }
+
+    if (errorString != null) {
+      CommonRefactoringUtil.showErrorMessage(ScalaAnonymousToInnerHandler.getRefactoringName, errorString, HelpID.ANONYMOUS_TO_INNER, project)
+    } else {
+      super.doOKAction()
+    }
+    classNameField.requestFocusInWindow
+  }
+
+  override protected def createNorthPanel: JComponent =
+    FormBuilder
+      .createFormBuilder
+      .addLabeledComponent(JavaRefactoringBundle.message("anonymousToInner.class.name.label.text"), classNameField)
+      .getPanel
+
+  override def getPreferredFocusedComponent: JComponent = classNameField.getFocusableComponent
+
+  override def createCenterPanel(): JComponent = {
+    val panel = new JPanel(new BorderLayout)
+    panel.add(createParametersPanel, BorderLayout.CENTER)
+    panel
+  }
+
+  def getClassName: String = classNameField.getEnteredName
+
+  def getVariables: Array[ScalaVariableData] = variables
+
+
+  private def createParametersPanel = {
+    val panel = new ParameterTablePanel(project, variables.asInstanceOf[Array[VariableData]], extendsBlock) {
+      override protected def updateSignature(): Unit = {
+      }
+
+      override protected def doEnterAction(): Unit = {
+        clickDefaultButton()
+      }
+
+      override protected def doCancelAction(): Unit = {
+        this.asInstanceOf[DialogWrapper].doCancelAction()
+      }
+    }
+    panel.setBorder(IdeBorderFactory.createTitledBorder(JavaRefactoringBundle.message("anonymousToInner.parameters.panel.border.title"), false))
+    panel
+  }
+
+  override protected def getHelpId: String = HelpID.ANONYMOUS_TO_INNER
+}
--- a/scala/scala-impl/src/org/jetbrains/plugins/scala/lang/refactoring/move/anonymousToInner/ScalaAnonymousToInnerDialog.scala
+++ b/scala/scala-impl/src/org/jetbrains/plugins/scala/lang/refactoring/move/anonymousToInner/ScalaAnonymousToInnerDialog.scala
@@ -97,7 +97,7 @@ class ScalaAnonymousToInnerDialog(project: Project, extendsBlock: ScExtendsBlock
 
   def getClassName: String = classNameField.getEnteredName
 
-  def getVariables: Array[ScalaVariableData] = variables
+  def getVariables: Array[ScalaVariableData] = variables.filter(_.passAsParameter)
 
 
   private def createParametersPanel = {
--- a/scala/scala-impl/src/org/jetbrains/plugins/scala/lang/refactoring/move/anonymousToInner/ScalaAnonymousToInnerHandler.scala
+++ b/scala/scala-impl/src/org/jetbrains/plugins/scala/lang/refactoring/move/anonymousToInner/ScalaAnonymousToInnerHandler.scala
@@ -3,9 +3,11 @@ package org.jetbrains.plugins.scala.lang.refactoring.move.anonymousToInner
 import com.intellij.openapi.application.ApplicationManager
 import com.intellij.openapi.command.CommandProcessor
 import com.intellij.openapi.diagnostic.Logger
+import com.intellij.openapi.editor.Editor
 import com.intellij.openapi.project.Project
 import com.intellij.openapi.util.NlsContexts
-import com.intellij.psi.{PsiComment, PsiElement}
+import com.intellij.psi.PsiElement
+import com.intellij.refactoring.util.CommonRefactoringUtil
 import com.intellij.util.IncorrectOperationException
 import org.jetbrains.annotations.VisibleForTesting
 import org.jetbrains.plugins.scala.ScalaBundle
@@ -27,19 +29,30 @@ import org.jetbrains.plugins.scala.project.ProjectContext
 object ScalaAnonymousToInnerHandler {
   private val LOG = Logger.getInstance(ScalaAnonymousToInnerHandler.getClass)
 
-  def invoke(project: Project, element: ScNewTemplateDefinition): Unit = {
+  private val cantRefactorBecauseOfVar = """Cannot perform refactoring.
+                                           |Extraction of anonymous class with references to vars out of scope is currently unsupported""".stripMargin
+
+  private val helpId = "reafctoring.anonymousToInner"
+
+  def invoke(project: Project, editor: Editor, element: ScNewTemplateDefinition): Unit = {
     val extendsBlock = element.extendsBlock
     val (variables, targetContainer) = parseInitialExtendsBlock(extendsBlock)
 
-    for {
-      DialogResult(className, renamedVariables) <- showRefactoringDialog(project, extendsBlock, variables, targetContainer)
-    } yield performRefactoring(project, className, renamedVariables, extendsBlock, element, targetContainer)
+    if (containsVarsOutOfScope(extendsBlock, variables))
+      CommonRefactoringUtil.showErrorHint(project, editor, cantRefactorBecauseOfVar, getRefactoringName, helpId)
+    else
+      for {
+        DialogResult(className, renamedVariables) <- showRefactoringDialog(project, extendsBlock, variables, targetContainer)
+      } yield performRefactoring(project, className, renamedVariables, extendsBlock, element, targetContainer)
   }
+
+  def containsVarsOutOfScope(extendsBlock: ScExtendsBlock, variables: Array[ScalaVariableData]): Boolean =
+    variables.exists(v => v.element.isVar && !extendsBlock.isAncestorOf(v.element))
 
   @VisibleForTesting
   def parseInitialExtendsBlock(extendsBlock: ScExtendsBlock): (Array[ScalaVariableData], Either[ScFile, ScTemplateDefinition]) = {
     val targetContainer = findTargetContainer(extendsBlock)
-    val usedVariables = collectUsedVariables(extendsBlock, targetContainer)
+    val usedVariables = collectUsedVariables(extendsBlock)
     (usedVariables, targetContainer)
   }
 
@@ -48,12 +61,12 @@ object ScalaAnonymousToInnerHandler {
       val action: Runnable = () => {
         try {
           val innerClassNewTemplate = newTemplateForInnerClass(project, className, variables, originalElement)
-          val (newClassFromAnonymous, maybeWarningComment) = createClass(className, anonClass, variables, project)
-          val containerWithNewInnerClass = targetContainer
+          val newClassFromAnonymous = createClass(className, anonClass, variables, project)
+
+          targetContainer
             .fold(file => file, classOrObject => classOrObject)
             .add(newClassFromAnonymous)
 
-          maybeWarningComment.foreach(comment => containerWithNewInnerClass.addBefore(comment, newClassFromAnonymous))
           originalElement.replace(innerClassNewTemplate)
         }
         catch {
@@ -75,28 +88,28 @@ object ScalaAnonymousToInnerHandler {
     createElementFromText[ScNewTemplateDefinition](text, newTemplate)
   }
 
-  private def collectUsedVariables(anonClass: ScExtendsBlock, targetContainer: Either[ScFile, ScTemplateDefinition]): Array[ScalaVariableData] = {
+  private def collectUsedVariables(anonClass: ScExtendsBlock): Array[ScalaVariableData] = {
     var res: List[ScTypedDefinition] = Nil
     anonClass.accept(new ScalaRecursiveElementVisitor() {
       override def visitReferenceExpression(expression: ScReferenceExpression) = {
         val refElement = expression.resolve
         refElement match {
           case p: ScParameter =>
-            val containingClass = p.getDeclarationScope
-            if (targetContainer.fold(_ => false, scTemplateDefinition => containingClass != scTemplateDefinition && !containingClass.isAncestorOf(scTemplateDefinition)))
               res = p :: res
           case r: ScReferencePattern =>
-            val containingClass = r.containingClass
-            if (targetContainer.fold(_ => false, scTemplateDefinition => containingClass != scTemplateDefinition && !containingClass.isAncestorOf(scTemplateDefinition) && !findTargetContainer(r).isLeft))
               res = r :: res
           case _ =>
         }
         super.visitReferenceExpression(expression)
       }
     })
-    res.reverse.distinct.map(parameter =>
-      new ScalaVariableData(parameter, true, parameter.`type`().getOrNothing)
-    ).toArray
+
+    res
+      .reverse
+      .filter(!anonClass.isAncestorOf(_))
+      .distinct
+      .map(parameter => new ScalaVariableData(parameter, true, parameter.`type`().getOrNothing))
+      .toArray
 
   }
 
@@ -110,14 +123,8 @@ object ScalaAnonymousToInnerHandler {
       None
   }
 
-  private def createClass(name: String, anonClass: ScExtendsBlock, variables: Array[ScalaVariableData], project: Project): (ScClass, Option[PsiComment]) = {
+  private def createClass(name: String, anonClass: ScExtendsBlock, variables: Array[ScalaVariableData], project: Project): ScClass = {
     implicit val projectContext: ProjectContext = new ProjectContext(project)
-
-    val containsVars = variables.exists(_.element.isVar)
-    val warningComment = if (containsVars)
-      Some(createElementFromText[PsiComment]("// refactor your code or use scala.runtime.ObjectRef", anonClass))
-    else
-      None
 
     val parameters = variables.map(v => s"${v.name}: ${v.`type`.getPresentableText}").mkString(", ")
     val text = s"class $name($parameters) extends ${anonClass.getText}"
@@ -139,7 +146,7 @@ object ScalaAnonymousToInnerHandler {
       }
     })
 
-    (newClass, warningComment)
+    newClass
   }
 
   private def findTargetContainer(elem: PsiElement): Either[ScFile, ScTemplateDefinition] = {
--- a/scala/scala-impl/src/org/jetbrains/plugins/scala/lang/refactoring/move/anonymousToInner/ScalaAnonymousToInnerHandler.scala
+++ b/scala/scala-impl/src/org/jetbrains/plugins/scala/lang/refactoring/move/anonymousToInner/ScalaAnonymousToInnerHandler.scala
@@ -5,7 +5,7 @@ import com.intellij.openapi.command.CommandProcessor
 import com.intellij.openapi.diagnostic.Logger
 import com.intellij.openapi.project.Project
 import com.intellij.openapi.util.NlsContexts
-import com.intellij.psi.PsiElement
+import com.intellij.psi.{PsiComment, PsiElement}
 import com.intellij.util.IncorrectOperationException
 import org.jetbrains.annotations.VisibleForTesting
 import org.jetbrains.plugins.scala.ScalaBundle
@@ -48,8 +48,12 @@ object ScalaAnonymousToInnerHandler {
       val action: Runnable = () => {
         try {
           val innerClassNewTemplate = newTemplateForInnerClass(project, className, variables, originalElement)
-          val newClassFromAnonymous = createClass(className, anonClass, variables, project)
-          targetContainer.fold(file => file.add(newClassFromAnonymous), classOrObject => classOrObject.addMember(newClassFromAnonymous, None))
+          val (newClassFromAnonymous, maybeWarningComment) = createClass(className, anonClass, variables, project)
+          val containerWithNewInnerClass = targetContainer
+            .fold(file => file, classOrObject => classOrObject)
+            .add(newClassFromAnonymous)
+
+          maybeWarningComment.foreach(comment => containerWithNewInnerClass.addBefore(comment, newClassFromAnonymous))
           originalElement.replace(innerClassNewTemplate)
         }
         catch {
@@ -102,18 +106,18 @@ object ScalaAnonymousToInnerHandler {
       None
   }
 
-  private def createClass(name: String, anonClass: ScExtendsBlock, variables: Array[ScalaVariableData], project: Project): ScClass = {
+  private def createClass(name: String, anonClass: ScExtendsBlock, variables: Array[ScalaVariableData], project: Project): (ScClass, Option[PsiComment]) = {
     implicit val projectContext: ProjectContext = new ProjectContext(project)
 
-    val parameters = for {
-      v <- variables
-      scTypedDeinition = v.element
-      name = v.name
-      isVar = if (scTypedDeinition.isVar) "var " else ""
-      typeText = v.`type`.getPresentableText
-    } yield s"$isVar$name: $typeText"
+    val containsVars = variables.exists(_.element.isVar)
+    val warningComment = if (containsVars)
+      Some(createElementFromText[PsiComment]("// refactor your code or use scala.runtime.ObjectRef", anonClass))
+    else
+      None
 
-    val text = s"class $name(${parameters.mkString(", ")}) extends ${anonClass.getText}"
+    val parameters = variables.map(v => s"${v.name}: ${v.`type`.getPresentableText}").mkString(", ")
+    val text = s"class $name($parameters) extends ${anonClass.getText}"
+
     val newClass = createElementFromText[ScClass](text, anonClass)
 
     val parameterOldNameToNewName = variables.map(v => v.variable.getName -> v.name).toMap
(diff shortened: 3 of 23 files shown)


scala/scala-impl/resources/messages/ScalaBundle.properties
```diff
@@ -1804,6 +1804,7 @@ move.members=Move members
 move.members.supported.only.stable.objects=Move refactoring is supported only for stable object members
 move.members.not.supported.implicits=Move refactoring is not supported for implicit definitions
 move.members.not.supported.overridden=Move refactoring is not supported for overridden definitions
+move.anonymousToInner.name=Convert Anonymous to Inner...
```
unkarjedy: In the dialog title it should not have "...", like in Java alternative.
https://jetbrains.design/intellij/text/punctuation/#ellipsis

unkarjedy:
The refactoring should be able to handle closures.
For a start, it would be enough to at least handle read operations.
Ideally, write operations could be also handled.

For example, in this case the refactoring breaks the code:
```scala
trait MyTrait {
  def foo0: Int
  def foo1: Int
  def foo2: Int
}

object Example {
    var localVar1 = 42

    def innerFoo(): Unit = {
      var localVar2 = 42
      
      val x = new MyTrait {
        override def foo0: Int = args.length
        override def foo1: Int = localVar1
        override def foo2: Int = localVar2
      }
    }
  }
}
```
image

scala/scala-impl/src/org/jetbrains/plugins/scala/lang/refactoring/move/anonymousToInner/ScalaAnonymousToInnerHandler.scala
```diff
@@ -0,0 +1,162 @@
...
+import org.jetbrains.plugins.scala.lang.psi.api.{ScFile, ScalaPsiElement, ScalaRecursiveElementVisitor}
+import org.jetbrains.plugins.scala.lang.psi.impl.ScalaPsiElementFactory.createElementFromText
+import org.jetbrains.plugins.scala.lang.refactoring.extractMethod.ScalaVariableData
+import org.jetbrains.plugins.scala.lang.refactoring.move.anonymousToInner.ScalaAnonymousToInnerHandler.getRefactoringName
+import org.jetbrains.plugins.scala.project.ProjectContext
+
+import scala.annotation.tailrec
+import scala.reflect.ClassTag
+
+class ScalaAnonymousToInnerHandler {
```
unkarjedy: I see at least two issues with the current implementation:
1. It does not handle local val/var (see `org.jetbrains.plugins.scala.lang.psi.api.statements.ScValueOrVariable`).
I think for a start, we can just handle read operations. Handling write operations to `var` could be handled in theory, but that would require a more complex design and implementation.

2. It captures class parameters of the class it extracts the class (`classParameter`, `classParameterField`).

Example (1 + 2)
```scala
trait MyTrait {
  def traitFoo: Int
}

class Example(classParameter: Int, val classParameterField: Int) {
  val fieldVal: Int = 0
  val fieldVar: Int = 0

  def foo(functionParameter: Int): Unit = {
    val localValue = 0
    var localVariable = 0

    def innerFoo(innerFunctionParameter: Int): Unit = {
      val localValueInner = 0
      var localVariableInner = 0

      val x = new MyTrait {
        override def traitFoo: Int = {
          //writes (might skip in the initial implementation)
          localVariable = 42
          localVariableInner = 23

          //reads
          classParameter + classParameterField +
            fieldVal + fieldVar +
            functionParameter + innerFunctionParameter +
            localValue + localVariable +
            localValueInner + localVariableInner
        }
      }
    }
  }
}
```

image

image


Example in Java (for 2)
```java
interface MyInterface {
    int foo();
}

class Scratch {
    public int fieldPublic = 0;
    public int fieldPrivate = 0;

    public void foo(int parameter) {
        int localVar = 0;
        System.out.println(new MyInterface() {
            @Override
            public int foo() {
                return parameter + localVar + fieldPublic + fieldPrivate;
            }
        });
    }
}
```

image

Note that in Java implementation it shows an extra option to create "static" class if there are no usages of class fields.
In future, in Scala refactoring, we could add an option to create the class in the companion object.
In that case it would capture class fields/parameters as well.
But it's fine if we don't have it in the first version.

scala/scala-impl/src/org/jetbrains/plugins/scala/lang/refactoring/move/anonymousToInner/ScalaAnonymousToInnerDialog.scala
```diff
@@ -0,0 +1,116 @@
...
+import org.jetbrains.plugins.scala.lang.psi.api.toplevel.templates.ScExtendsBlock
+import org.jetbrains.plugins.scala.lang.psi.api.toplevel.typedef.ScTemplateDefinition
+import org.jetbrains.plugins.scala.lang.psi.types.ScParameterizedType
+import org.jetbrains.plugins.scala.lang.psi.types.api.designator.ScDesignatorType
+
+import java.awt.BorderLayout
+import javax.swing.{JComponent, JPanel}
+import scala.jdk.CollectionConverters._
+
+class ScalaAnonymousToInnerDialog(project: Project, anonClass: ScExtendsBlock, variables: Array[VariableData], target: Either[ScFile, ScTemplateDefinition]) extends DialogWrapper(project) {
```
unkarjedy: Am I right that the original implementation was inspired by `com.intellij.refactoring.anonymousToInner.AnonymousToInnerDialog`?

If so, let's add a ScalaDoc comment:
```scala
/**
 * @note original implementation was inspired by [[com.intellij.refactoring.anonymousToInner.AnonymousToInnerDialog]]
 */
```

It might help in future in case we want to adapt some new improvements from the Java version.

unkarjedy: The same is actual for other classes if applicable

scala/scala-impl/src/org/jetbrains/plugins/scala/lang/refactoring/move/anonymousToInner/ScalaAnonymousToInnerHandler.scala
```diff
@@ -0,0 +1,162 @@
...
+          case e: IncorrectOperationException =>
+            LOG.error(e)
+        }
+
+      }
+      ApplicationManager.getApplication.runWriteAction(action)
+
+  }, getRefactoringName, null)
+
+  @tailrec
```
unkarjedy: We already have some utility extension methods for that.
Please see `org.jetbrains.plugins.scala.extensions.PsiElementExt#parentOfType`.

scala/scala-impl/src/org/jetbrains/plugins/scala/lang/refactoring/move/anonymousToInner/ScalaAnonymousToInnerHandler.scala
```diff
@@ -0,0 +1,162 @@
...
+  }
+
+  private def collectUsedVariables(anonClass: ScExtendsBlock): Array[VariableData] = {
+    var res: List[ScParameter] = Nil
+    anonClass.accept(new ScalaRecursiveElementVisitor() {
+      override def visitReferenceExpression(expression: ScReferenceExpression) = {
+        val refElement = expression.resolve
+        refElement match {
+          case p: ScParameter =>
+            val containingClass = PsiTreeUtil.getParentOfType(p, classOf[ScClass])
```
unkarjedy: Did you also consider `ScTrait`, `ScEnum` in addition to `ScClass` and `ScObject`?

scala/scala-impl/src/org/jetbrains/plugins/scala/lang/refactoring/move/anonymousToInner/ScalaAnonymousToInnerHandler.scala
```diff
@@ -0,0 +1,162 @@
...
+      new ScalaVariableData(parameter, true, parameter.`type`().getOrNothing)
+    ).toArray
+
+  }
+
+  private case class DialogResult(className: String, variables: Array[VariableData])
+
+  private def showRefactoringDialog(project: Project, anonClass: ScExtendsBlock, usedVariables: Array[VariableData], target: Either[ScFile, ScTemplateDefinition]) = {
+    val dialog = new ScalaAnonymousToInnerDialog(project, anonClass, usedVariables, target)
+    if (dialog.showAndGet()) {
```
unkarjedy: (minor) might omit wrapping braces `{}` here (as you did on the next line with `None`)

scala/scala-impl/src/org/jetbrains/plugins/scala/lang/refactoring/move/anonymousToInner/ScalaAnonymousToInnerHandler.scala
```diff
@@ -0,0 +1,162 @@
...
+        }
+        super.visitReferenceExpression(expression)
+      }
+    })
+
+    newClass
+  }
+
+  @tailrec
+  private def findAnonymousClass(element: PsiElement): Option[ScExtendsBlock] =
```
unkarjedy: The naming is confusing.
I would expect findAnonymous**Class** to return a subtype of "class" (ScNewTemplateDefinition or PsiClass).
Let's replace this with `element.parentOfType[ScNewTemplateDefinition]`.
(requires `import org.jetbrains.plugins.scala.extensions.PsiElementExt`)

Then you can call `ScTemplateDefinition#extendsBlock` method

scala/scala-impl/src/org/jetbrains/plugins/scala/lang/refactoring/move/anonymousToInner/ScalaAnonymousToInnerHandler.scala
```diff
@@ -0,0 +1,162 @@
...
+    Option(element) match {
+      case Some(n: ScNewTemplateDefinition) =>
+        n.findFirstChildByTypeScala[ScExtendsBlock](ScalaElementType.EXTENDS_BLOCK)
+      case Some(_) => findAnonymousClass(element.getParent)
+      case None => None
+    }
+
+  @tailrec
+  private def findTargetContainer(elem: PsiElement): Either[ScFile, ScTemplateDefinition] = {
```
unkarjedy, line 154: What about `ScTrait` and `ScEnum`?

scala/scala-impl/src/org/jetbrains/plugins/scala/lang/refactoring/move/anonymousToInner/ScalaAnonymousToInnerHandler.scala
```diff
@@ -0,0 +1,162 @@
...
+
+class ScalaAnonymousToInnerHandler {
+  private val LOG = Logger.getInstance(classOf[ScalaAnonymousToInnerHandler])
+
+  def invoke(project: Project, element: ScNewTemplateDefinition): Unit = for {
+    (anonClass, variables, targetContainer) <- parseInitialAnonymousClass(element)
+    DialogResult(className, renamedVariables) <- showRefactoringDialog(project, anonClass, variables, targetContainer)
+  } yield performRefactoring(project, className, renamedVariables, anonClass, element, targetContainer)
+
+  def parseInitialAnonymousClass(element: ScNewTemplateDefinition): Option[(ScExtendsBlock, Array[VariableData], Either[ScFile, ScTemplateDefinition])] = for {
```
unkarjedy: Please mark this method with  `@VisibleForTesting` annotation
(`org.jetbrains.annotations.VisibleForTesting`)

scala/scala-impl/src/org/jetbrains/plugins/scala/lang/refactoring/move/anonymousToInner/ScalaMoveAnonymousToInnerDelegate.scala
```diff
@@ -0,0 +1,35 @@
...
+
+  override def canMove(elements: Array[PsiElement], targetContainer: PsiElement, reference: PsiReference): Boolean =
+    findNewTemplateDefinition(elements(0), reference).exists(_.isAnonimous)
+
+  private def findNewTemplateDefinition(element: PsiElement, reference: PsiReference): Option[ScNewTemplateDefinition] = {
+    val maybeElement = Option(reference).map(_.getElement).orElse(Option(element))
+    maybeElement.flatMap(element => Option(PsiTreeUtil.getParentOfType(element, classOf[ScNewTemplateDefinition])))
+  }
+
+  val handler = new ScalaAnonymousToInnerHandler
```
unkarjedy: I wonder if we need this handler as a class at all.
AFAIS it doesn't contain any state.
Maybe we can make it `ScalaAnonymousToInnerHandler`

scala/scala-impl/src/org/jetbrains/plugins/scala/lang/refactoring/move/anonymousToInner/ScalaAnonymousToInnerHandler.scala
```diff
@@ -0,0 +1,162 @@
...
+  @tailrec
+  private def findAnonymousClass(element: PsiElement): Option[ScExtendsBlock] =
+    Option(element) match {
+      case Some(n: ScNewTemplateDefinition) =>
+        n.findFirstChildByTypeScala[ScExtendsBlock](ScalaElementType.EXTENDS_BLOCK)
+      case Some(_) => findAnonymousClass(element.getParent)
+      case None => None
+    }
+
+  @tailrec
```
unkarjedy: (minor) alternative implementation without recursion (in current method)

```scala
  private def findTargetContainer(elem: PsiElement): Either[ScFile, ScTemplateDefinition] = {
    val container = elem.parents.collectFirst {
      case obj: ScObject => Right(obj)
      case clazz: ScClass => Right(clazz)
      case file: ScFile => Left(file) //Scala3 top level definition
    }
    container.get //assuming that there is at least "ScFile" parent, so it's safe to call "get"
  }
```

scala/scala-impl/test/org/jetbrains/plugins/scala/lang/refactoring/anonymousToInner/ScalaAnonymousToInnerTest.scala
```diff
@@ -0,0 +1,152 @@
...
+    scalaFixture.configureFromFileText(initialText)
+
+    val editor = CommonDataKeys.EDITOR.getData(DataManager.getInstance().getDataContextFromFocusAsync.blockingGet(5, TimeUnit.SECONDS))
+    val caretModel = editor.getCaretModel
+    assert(caretModel.getCaretCount == 1, "Expected exactly one caret.")
+    assert(caretModel.getOffset > 0, s"Not specified caret marker in test case. Use <caret> in scala file for this.")
+
+    val scalaFile = getFile.asInstanceOf[ScalaFile]
+
+    val myElement = scalaFile.findReferenceAt(caretModel.getOffset).getElement
```
unkarjedy: `my` prefix is used for java fields.
Just `element` is enough.

scala/scala-impl/test/org/jetbrains/plugins/scala/lang/refactoring/anonymousToInner/ScalaAnonymousToInnerTest.scala
```diff
@@ -0,0 +1,152 @@
...
+    val newTemplateDefinition = handler.findParentOfType[ScNewTemplateDefinition](myElement).getOrElse(throw new Exception("expected to find the `new` keyword before the anonymous type"))
```
unkarjedy: Too long line, can move the exception to the new line:
```scala
val newTemplateDefinition = handler.findParentOfType[ScNewTemplateDefinition](element).getOrElse {
      throw new Exception("expected to find the `new` keyword before the anonymous type")
    }
    ```

File: scala/scala-impl/test/org/jetbrains/plugins/scala/lang/refactoring/anonymousToInner/ScalaAnonymousToInnerTest.scala (hunk 12 above)
Comment (unkarjedy) [reply scala/scala-impl/test/org/jetbrains/plugins/scala/lang/refactoring/anonymousToInner/ScalaAnonymousToInnerTest.scala]: Also this can be shorter:
`val element = scalaFile.findElementAt(caretModel.getOffset)`

**Comment by unkarjedy** [comment]
@leonbur 
Thanks a lot for the contribution!
It's a nice feature which I would also personally use.

File: scala/scala-impl/resources/messages/ScalaBundle.properties (hunk 1 above)
Comment (leonbur) [reply scala/scala-impl/resources/messages/ScalaBundle.properties]: from your link:
>End an action name with an ellipsis if it opens a dialog where input is required or possible. An ellipsis helps users understand if an action is immediate or additional interaction is to follow.

but the action does open a dialog and input is required. also, this action comes as a replacement to "Move class..." which also has ellipsis.

File: scala/scala-impl/src/org/jetbrains/plugins/scala/lang/refactoring/move/anonymousToInner/ScalaAnonymousToInnerHandler.scala (hunk 4 above)
Comment (leonbur) [reply scala/scala-impl/src/org/jetbrains/plugins/scala/lang/refactoring/move/anonymousToInner/ScalaAnonymousToInnerHandler.scala]: lol, I even used it elsewhere. will amend.

File: scala/scala-impl/resources/messages/ScalaBundle.properties (hunk 1 above)
Comment (unkarjedy) [reply scala/scala-impl/resources/messages/ScalaBundle.properties]: The action does open a dialog, but in the dialog itself also has a title and it shouldn't have the `...`.
See, for example `JavaRefactoringBundle.properties` - it has two keys:
```
anonymousToInner.refactoring.name=Convert Anonymous to Inner
...
convert.anonymous.to.inner.action.name=Convert Anonymous to Inner\u2026
```

One is used in the action and another in the title.
`\u2026` ~ UTF-8 for `...`

File: scala/scala-impl/src/org/jetbrains/plugins/scala/lang/refactoring/move/anonymousToInner/ScalaAnonymousToInnerHandler.scala (hunk 11 above)
Comment (leonbur) [reply scala/scala-impl/src/org/jetbrains/plugins/scala/lang/refactoring/move/anonymousToInner/ScalaAnonymousToInnerHandler.scala]: there is no `parents` collection on `PsiElement`. if there were, I would have used that instead.

Comment (unkarjedy) [reply scala/scala-impl/src/org/jetbrains/plugins/scala/lang/refactoring/move/anonymousToInner/ScalaAnonymousToInnerHandler.scala]: It's available in the same extension methods wrapper as `parentOfType`:
`import org.jetbrains.plugins.scala.extensions.PsiElementExt`
Also, IntelliJ suggests to import it (though you need to first type the name)

image

File: scala/scala-impl/src/org/jetbrains/plugins/scala/lang/refactoring/move/anonymousToInner/ScalaAnonymousToInnerHandler.scala (hunk 7 above)
Comment (leonbur) [reply scala/scala-impl/src/org/jetbrains/plugins/scala/lang/refactoring/move/anonymousToInner/ScalaAnonymousToInnerHandler.scala]: thanks for noticing this, this entire function is unnecessary since I now pass it ScNewTemplateDefinition anyway (it wasn't like that before). so I'll delete it

File: scala/scala-impl/src/org/jetbrains/plugins/scala/lang/refactoring/move/anonymousToInner/ScalaAnonymousToInnerHandler.scala (hunk 11 above)
Comment (leonbur) [reply scala/scala-impl/src/org/jetbrains/plugins/scala/lang/refactoring/move/anonymousToInner/ScalaAnonymousToInnerHandler.scala]: oh, IJ didn't suggest it to me at first. I will use your version.
regarding your comment on `container.get`, why wouldn't there be an `ScFile` at the top?

Comment (unkarjedy) [reply scala/scala-impl/src/org/jetbrains/plugins/scala/lang/refactoring/move/anonymousToInner/ScalaAnonymousToInnerHandler.scala]: > why wouldn't there be an ScFile at the top?

There should be.
But it's generally not nice to use `.get`, so with the explicit comment I reassure that we don't expect NPE because at least `ScFile` is expected as one of the parents.

With your implementation, I would add a similar comment to `elem.getParent`

File: scala/scala-impl/src/org/jetbrains/plugins/scala/lang/refactoring/move/anonymousToInner/ScalaAnonymousToInnerHandler.scala (hunk 8 above)
Comment (leonbur) [reply scala/scala-impl/src/org/jetbrains/plugins/scala/lang/refactoring/move/anonymousToInner/ScalaAnonymousToInnerHandler.scala:154]: I'll add `ScTrait`. but `ScEnum` is redundant because it extends `ScClass` anyway

Comment (unkarjedy) [reply scala/scala-impl/src/org/jetbrains/plugins/scala/lang/refactoring/move/anonymousToInner/ScalaAnonymousToInnerHandler.scala:154]: Could you please at least add a small meaningful test for ScEnum case (and ScTrait)
The fact that `ScEnum` extends `ScClass` is an implementation detail.
We had a lot of refactorings of `ScEnum` in 2023.2/3 and in future there is still a chance the hierarchy will be remade.

File: scala/scala-impl/src/org/jetbrains/plugins/scala/lang/refactoring/move/anonymousToInner/ScalaAnonymousToInnerHandler.scala (hunk 5 above)
Comment (leonbur) [reply scala/scala-impl/src/org/jetbrains/plugins/scala/lang/refactoring/move/anonymousToInner/ScalaAnonymousToInnerHandler.scala]: I'm removing the parent check as I don't even remember why I added it (I started work on this a few months ago). it works even without it.

File: scala/scala-impl/src/org/jetbrains/plugins/scala/lang/refactoring/move/anonymousToInner/ScalaAnonymousToInnerHandler.scala (hunk 2 above)
Comment (leonbur) [reply scala/scala-impl/src/org/jetbrains/plugins/scala/lang/refactoring/move/anonymousToInner/ScalaAnonymousToInnerHandler.scala]: My solution is basically to add the referenced vars and declare them as `var` in the parameter declaration of the class so that they could be mutated inside the newly created class.

File: scala/scala-impl/src/org/jetbrains/plugins/scala/lang/refactoring/move/anonymousToInner/ScalaAnonymousToInnerHandler.scala (hunk 8 above)
Comment (leonbur) [reply scala/scala-impl/src/org/jetbrains/plugins/scala/lang/refactoring/move/anonymousToInner/ScalaAnonymousToInnerHandler.scala:154]: I have added a test for enum but there's some issue in the testkit where it creates the virtual file for the test from the string it doesn't build the AST correctly and doesn't contain the `ScEnum` in it. manually running the function when intelliJ is running works correctly.

Since the test doesn't create the ScEnum node correctly, the test fails because the created class is generated on the top level instead of inside the enum. please try to run `ScalaAnonymousToInnerTest.scala#testInsideEnum`.
The rest of the tests pass correctly.

Comment (unkarjedy) [reply scala/scala-impl/src/org/jetbrains/plugins/scala/lang/refactoring/move/anonymousToInner/ScalaAnonymousToInnerHandler.scala:154]: Consider overriding `supportedIn` method in the test class.
Example:
```scala
override protected def supportedIn(version: ScalaVersion): Boolean = version.isScala3
```

leonbur: It worked! thanks

unkarjedy:
@leonbur Please let us know once you are done and are not planning to push any new changes

leonbur:
> @leonbur Please let us know once you are done and are not planning to push any new changes

@unkarjedy 
I'm done. do I need to squash?

scala/scala-impl/test/org/jetbrains/plugins/scala/lang/refactoring/anonymousToInner/ScalaAnonymousToInnerTest.scala
```diff
@@ -0,0 +1,233 @@
...
+    )
+
+    val extendsBlock = newTemplateDefinition.extendsBlock
+    val (variables, targetContainer) = ScalaAnonymousToInnerHandler.parseInitialExtendsBlock(extendsBlock)
+    ScalaAnonymousToInnerHandler.performRefactoring(getProject, className, variables, extendsBlock, newTemplateDefinition, targetContainer)
+
+    Assert.assertEquals(expectedText, scalaFile.getText)
+  }
+
+  override protected def supportedIn(version: ScalaVersion): Boolean = version.isScala3
```
unkarjedy: Please move it to the beginning of the class, before all the methods in the test class

unkarjedy:
> do I need to squash?

@leonbur 
Yes please.
Also, please add a prefix to the commit message: `[refactoring] `

I will then manually pick the change and run it through our internal CI

leonbur:
@unkarjedy well, I did squash my commits but I think I have made a mess because I pulled from upstream during my development that's why there's a merge in the middle.

unkarjedy:
I will cherry-pick this change manually and run it using our internal CI.
Please watch https://youtrack.jetbrains.com/issue/SCL-227 for status update
I will include it to 2023.3 Nightly channel (it will be automatically available in 2024.1 as well)

Thanks again for the contribution!

scala/scala-impl/src/org/jetbrains/plugins/scala/lang/refactoring/move/anonymousToInner/ScalaAnonymousToInnerHandler.scala
unkarjedy: @leonbur 
Oh, sorry, I missed this update.
> My solution is basically to add the referenced vars and declare them as var in the parameter declaration of the class so that they could be mutated inside the newly created class.

Unfortunately, this is not a correct solution.
Such refactoring breaks the semantics of the code, which is definitely not expected from the "refactoring"

For example, suppose you have this code:
```scala
trait MyTrait {
  def foo(): Unit
}

object Example {
    var localVar = 0
    val t = new MyTrait {
      override def foo(): Unit = {
        localVar += 1
      }
    }
    t.foo()
    t.foo()
    println(localVar)
  }
}
```
it outputs `2`

After you apply the refactoring the code becomes as follows:
```scala
trait MyTrait {
  def foo(): Unit
}

object Example {
    var localVar = 0
    val t = new MyMyTrait(localVar)
    t.foo()
    t.foo()
    println(localVar)
  }

  class MyMyTrait(var localVar: Int) extends MyTrait {
    override def foo(): Unit = {
      localVar += 1
    }
  }
}
```
and the output is `0`, because the original `var` is not changed.

unkarjedy: When I wrote
> Handling write operations to var could be handled in theory, but that would require a more complex design and implementation.

I kept in mind another solution.
You would need to create a `Ref` wrapper for the var.
That's what actually Scala compiler automatically does in the JVM byte code.
This is the decompiled version:
image

However I am not 100% sure that we should do the same.
Users might want to address this issue in some different way.
We might leave `localVar` an ordinary parameter.
The resulting code won't compile in the places where `localVar` is updated, but it's better than silently breaking the code.
We can add some special line comment in every `localVar =`, something like
`// refactor your code or use scala.runtime.ObjectRef`
(we the actual infer `*Ref` name from the variable type)

Ideally, we could add some setting in the dialog in order user can do it automatically.
But the UX needs more thinking.
For this first version it would be OK to do as described above.

unkarjedy: Also, this part is still unresolved:
> It captures class parameters of the class it extracts the class (classParameter, classParameterField).
>...
>Note that in Java implementation it shows an extra option to create "static" class if there are no usages of class fields.
In future, in Scala refactoring, we could add an option to create the class in the companion object.
In that case it would capture class fields/parameters as well.
But it's fine if we don't have it in the first version.

image

unkarjedy:
@leonbur Sorry, I missed one comment, so I reopened the ticket

scala/scala-impl/src/org/jetbrains/plugins/scala/lang/refactoring/move/anonymousToInner/ScalaAnonymousToInnerHandler.scala
leonbur: ok, now understand. I'll add a comment and remove the `var` prefix for writable parameters.

regarding fieldVal,classParameter,... do you want them to NOT become parameters of the created class?

unkarjedy: > regarding fieldVal,classParameter,... do you want them to NOT become parameters of the created class?

Yes, because those will be accessible in the extracted inner class automatically.

leonbur:
@unkarjedy please re-review and let me know if I can squash

scala/scala-impl/src/org/jetbrains/plugins/scala/lang/refactoring/move/anonymousToInner/ScalaAnonymousToInnerHandler.scala
unkarjedy: The new line is added in the wrong place for the comment
image

Other thoughts:
1. Let's prepend "todo"
2. Let's mention the var names in order it's clear what that comment is about 

Something like
```scala
//todo: refactor your code or use scala.runtime.ObjectRef (localVar)
//todo: refactor your code or use scala.runtime.ObjectRef (localVar1, localVar2)
```

scala/scala-impl/src/org/jetbrains/plugins/scala/lang/refactoring/move/anonymousToInner/ScalaAnonymousToInnerHandler.scala
```diff
@@ -0,0 +1,157 @@
...
+      Some(DialogResult(dialog.getClassName, dialog.getVariables))
+    else
+      None
+  }
+
+  private def createClass(name: String, anonClass: ScExtendsBlock, variables: Array[ScalaVariableData], project: Project): (ScClass, Option[PsiComment]) = {
+    implicit val projectContext: ProjectContext = new ProjectContext(project)
+
+    val containsVars = variables.exists(_.element.isVar)
+    val warningComment = if (containsVars)
```
unkarjedy: (I will add it as another comment to track it easier)
I've just realised that simply ignoring "write" operations and adding the comment is not enough.
Here is a modified example:
```scala
trait MyTrait {
  def foo(): Unit
}

object Example {
    var localVar = 0

    val t = new MyTrait {
      override def foo(): Unit = {
        println(localVar)
      }
    }

    t.foo()
    localVar += 1
    t.foo()
    localVar += 1

    println(localVar)
  }
}
```

Notice that `localVar` is not modified inside the anonymous class.
It's only accessed inside it.
But the variable is modified in some other places outside the anonymous class.

The code outputs:
```
0
1
2
```

When you extract the class it looks like this:
```scala
trait MyTrait {
  def foo(): Unit
}

object Example {
    var localVar = 0

    val t = new MyMyTrait(localVar)

    t.foo()
    localVar += 1
    t.foo()
    localVar += 1

    println(localVar)
  }
  // refactor your code or use scala.runtime.ObjectRef

  class MyMyTrait(localVar: Int) extends MyTrait {
    override def foo(): Unit = {
      println(localVar)
    }
  }
}
```

Now `MyMyTrait` accesses its local copy of the var.
And the output of the program is:
```
0
0
2
```

unkarjedy: It's a simple example, but it illustrates that the refactoring still can break the code silently.
I see two solutions:

1. Ignore `var`s, don't include them as a parameter.
The code can potentially become red, and user will have to handle it manually.
Leave `//todo` comments where needed to mention that we can't handle vars yet and user needs to use `scala.runtime.*Ref` objects manually.

2. Rewrite the code to `*Ref`
E.g. the example from above will look like:
```scala
import scala.runtime.IntRef

trait MyTrait {
  def foo(): Unit
}

object Example {
    val localVar = new IntRef(0)

    val t = new MyMyTrait(localVar)

    t.foo()
    localVar.elem += 1
    t.foo()
    localVar.elem += 1

    println(localVar)
  }

  class MyMyTrait(localVar: IntRef) extends MyTrait {
    override def foo(): Unit = {
      println(localVar.elem)
    }
  }
}
```

The first approach is the easiest.
If you choose it, please create a new YT ticket (and link to https://youtrack.jetbrains.com/issue/SCL-227) describing the limitation and mentioning the 2nd approach.

The second approach is the correct one, but would require some extra effort.
It might be practical to handle it as a separate change.
The refactoring without `var` handling would be already useful.
Some points to consider:
- Mind that there is `scala.runtime.ObjectRef` and multiple `*Ref` classes for primitive types.
- Mind that there are `scala.runtime.*Ref` and `scala.runtime.Volatile*Ref`

scala/scala-impl/src/org/jetbrains/plugins/scala/lang/refactoring/move/anonymousToInner/ScalaAnonymousToInnerHandler.scala
```diff
@@ -0,0 +1,157 @@
...
+    if (dialog.showAndGet())
+      Some(DialogResult(dialog.getClassName, dialog.getVariables))
+    else
+      None
+  }
+
+  private def createClass(name: String, anonClass: ScExtendsBlock, variables: Array[ScalaVariableData], project: Project): (ScClass, Option[PsiComment]) = {
+    implicit val projectContext: ProjectContext = new ProjectContext(project)
+
+    val containsVars = variables.exists(_.element.isVar)
```
unkarjedy: Handling of vars should be covered in tests

scala/scala-impl/src/org/jetbrains/plugins/scala/lang/refactoring/move/anonymousToInner/ScalaAnonymousToInnerHandler.scala
```diff
@@ -0,0 +1,157 @@
...
+  }
+
+  private def collectUsedVariables(anonClass: ScExtendsBlock, targetContainer: Either[ScFile, ScTemplateDefinition]): Array[ScalaVariableData] = {
+    var res: List[ScTypedDefinition] = Nil
+    anonClass.accept(new ScalaRecursiveElementVisitor() {
+      override def visitReferenceExpression(expression: ScReferenceExpression) = {
+        val refElement = expression.resolve
+        refElement match {
+          case p: ScParameter =>
+            val containingClass = p.getDeclarationScope
```
unkarjedy: Ignoring of containing class fields/parameters should be covered with tests

scala/scala-impl/src/org/jetbrains/plugins/scala/lang/refactoring/move/anonymousToInner/ScalaAnonymousToInnerHandler.scala
```diff
@@ -0,0 +1,157 @@
...
+            if (targetContainer.fold(_ => false, scTemplateDefinition => containingClass != scTemplateDefinition && !containingClass.isAncestorOf(scTemplateDefinition)))
+              res = p :: res
+          case r: ScReferencePattern =>
+            val containingClass = r.containingClass
+            if (targetContainer.fold(_ => false, scTemplateDefinition => containingClass != scTemplateDefinition && !containingClass.isAncestorOf(scTemplateDefinition) && !findTargetContainer(r).isLeft))
```
unkarjedy: This looks very similar to the code above.
Does this part matter
`!findTargetContainer(r).isLeft)`
?

If so, please, cover it with a separate test case.
It's not that obvious without context, and it's easy to have regressions here.

If it's not imporant difference, then the code with two branches could be refactored to avoid duplication.

unkarjedy: There are multiple issues with the condition:
1. It's quite long and unreadable.
It's better to at least rename `scTemplateDefinition` to just `td`, it's ok for the short lambda parameter

2. It doesn't reflect the semantics.
The condition reads as "add or not add element to collection" without hints about the purpose.
We could fix it by 1. extracting it to var 2. negating the condition
Example
```scala
val isAccessible = targetContainer.fold(_ => true, td => cc == td || cc.isAncestorOf(td) || findTargetContainer(r).isLeft)
if (isAccessible)
  res = r :: res
```
Now the purpose is slightly more clear

scala/scala-impl/src/org/jetbrains/plugins/scala/lang/refactoring/move/anonymousToInner/ScalaAnonymousToInnerHandler.scala
leonbur: In option 1 if we don't include the vars in the parameters they can accidentally reference another field with the same name in the target container and therefore compile without errors. extending your example:

```
trait MyTrait {
  def foo(): Unit
}

object Example {

  var localVar = 10
  
    var localVar = 0

    val t = new MyTrait {
      override def foo(): Unit = {
        println(localVar)
      }
    }

    t.foo()
    localVar += 1
    t.foo()
    localVar += 1

    println(localVar)
  }
}
```
that prints
```
0
1
2
```


would become:
```trait MyTrait {
  def foo(): Unit
}

object Example {

  var localVar = 10
  
    var localVar = 0

    val t = new MyMyTrait()

    t.foo()
    localVar += 1
    t.foo()
    localVar += 1

    println(localVar)
  }

  class MyMyTrait() extends MyTrait {
    override def foo(): Unit = {
      println(localVar) //todo: refactor your code or use scala.runtime.ObjectRef
    }
  }
}
```
will print
```
10
10
2
```

As a possible quick solution, I could just throw an error to the user if the code contains var references instead of opening the dialog

unkarjedy: > As a possible quick solution, I could just throw an error to the user if the code contains var references instead of opening the dialog

Indeed, a good tradeoff.
Let's do it.
We can show a red error tooltip to the user, like this:
image
You might use `com.intellij.refactoring.util.CommonRefactoringUtil#showErrorHint`

unkarjedy:
I will squash it myself and add some extra commits from me.
I will run it through our CI.

PR #656, opened 2024-03-25
[inspection] re-enable comparing unrelated types inspection for scala 3

disordered:
Adds support for '--language:strictEquality' compiler flag and 'CanEqual' when running inspection on scala 3 code

### DIFF
--- a/scala/scala-impl/src/org/jetbrains/plugins/scala/codeInspection/typeChecking/ComparingUnrelatedTypesInspection.scala
+++ b/scala/scala-impl/src/org/jetbrains/plugins/scala/codeInspection/typeChecking/ComparingUnrelatedTypesInspection.scala
@@ -1,6 +1,6 @@
 package org.jetbrains.plugins.scala.codeInspection.typeChecking
 
-import com.intellij.codeInspection.{LocalInspectionTool, ProblemHighlightType, ProblemsHolder}
+import com.intellij.codeInspection.{LocalInspectionTool, ProblemsHolder}
 import com.intellij.psi.PsiMethod
 import com.siyeh.ig.psiutils.MethodUtils
 import org.jetbrains.annotations.Nls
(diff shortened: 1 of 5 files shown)


disordered:
I need help to finish this PR.

`testCanEqualIncompatibleCaseObjects` fails because `ScTypeElement.type()` method results in failure for `case object` types. I'm not sure how else I can convert it to `ScType`.

These tests fail because I can't figure out how I can lookup imported `given` implicits:
  - `testCanEqualImportedFromCompanion`
  - `testCanEqualImported`
  - `testCanEqualDerivedOnTrait` - I assume this will result in auto import as well

SrTobi:
Hi @disordered. Thanks for the PR. We discussed it today and you are absolutely right, that we are missing the multiversal equality at the moment. I have not completely looked into this, but we will probably activate the inspection only if *strictEquality* is **not** used. When *strictEquality* is used, we will do something in the Annotator to issue a real error, as errors are normally not the job of inspections.

Also we are currently pretty close to the next release, so I'll probably start working on this only in about two weeks. I'll keep you up to date.

disordered:
> Hi @disordered. Thanks for the PR. We discussed it today and you are absolutely right, that we are missing the multiversal equality at the moment. I have not completely looked into this, but we will probably activate the inspection only if _strictEquality_ is **not** used. When _strictEquality_ is used, we will do something in the Annotator to issue a real error, as errors are normally not the job of inspections.
> 
> Also we are currently pretty close to the next release, so I'll probably start working on this only in about two weeks. I'll keep you up to date.

EDIT: On second read, I think you meant that `CanEqual` inspection should not be performed at all, including when compiler is not in `strictEquality` mode. If that's the case, then there are still fixes in the PR for tests in Scala 3.

I have already covered that part. The inspection doesn't run when `strictEquality` is enabled and it is in scala 3, except if it is a primitive type or `String` (we could add other types later), since you can make scala compiler happy by declaring `CanEqual[String, Int]` but it is not possible to implement this, unless some reflection tricks are used. Well, technically inspection runs, but it does not inspect `CanEqual` declarations. When declaration is missing, IntelliJ already shows an error stating that types are not comparable.

I mentioned above what I think are the only things missing for this to go out. The failing tests are part of the PR already.

I also just realised, I did not include a test that checks with `strictEquality` enabled, but I also cannot find an example of how to configure a test run with specific compiler flags, so help there is needed as well.

PR #660, opened 2024-04-29
[annotator] support scala 3.4 irrefutable patterns in for-comprehension #SCL-22274 fixed

disordered:
This fixes false missing `withFilters` error for irrefutable patterns in for comprehensions introduced with Scala 3.4 and Scala 3.0 + `-source:future`.

Related #SCL-22468

### DIFF
--- a/scala/scala-impl/src/org/jetbrains/plugins/scala/lang/psi/impl/expr/ScForImpl.scala
+++ b/scala/scala-impl/src/org/jetbrains/plugins/scala/lang/psi/impl/expr/ScForImpl.scala
@@ -184,13 +184,16 @@ class ScForImpl(node: ASTNode) extends ScExpressionImplBase(node) with ScFor wit
     }
 
     def needsPatternMatchFilter(pattern: ScPattern): Boolean = {
-      val hasSc3Case = pattern.isScala3OrSource3Enabled &&
-        pattern.prevSiblingNotWhitespace.exists(_.textMatches(ScalaModifier.CASE))
-      val hasSc3IrrefutablePatternsEnabled = pattern.isScala3OrSource3Enabled &&
-        (pattern.features.hasSourceFutureFlag || pattern.scalaLanguageLevel.exists(_ >= ScalaLanguageLevel.Scala_3_4))
+      lazy val hasSc3Case = pattern.prevSiblingNotWhitespace.exists(_.textMatches(ScalaModifier.CASE))
+      lazy val hasSc3IrrefutablePatternsEnabled =
+        pattern.features.hasSourceFutureFlag || pattern.scalaLanguageLevel.exists(_ >= ScalaLanguageLevel.Scala_3_4)
+      lazy val isRefutablePattern = !pattern.isIrrefutableFor(if (forDisplay) pattern.expectedType else None)
 
-      hasSc3Case ||
-        !(hasSc3IrrefutablePatternsEnabled || pattern.isIrrefutableFor(if (forDisplay) pattern.expectedType else None))
+      if (pattern.isScala3OrSource3Enabled) {
+        hasSc3Case || (!hasSc3IrrefutablePatternsEnabled && isRefutablePattern)
+      } else {
+        isRefutablePattern
+      }
     }
 
     val resultText = new mutable.StringBuilder()
--- a/scala/scala-impl/src/org/jetbrains/plugins/scala/lang/psi/impl/expr/ScForImpl.scala
+++ b/scala/scala-impl/src/org/jetbrains/plugins/scala/lang/psi/impl/expr/ScForImpl.scala
@@ -184,12 +184,13 @@ class ScForImpl(node: ASTNode) extends ScExpressionImplBase(node) with ScFor wit
     }
 
     def needsPatternMatchFilter(pattern: ScPattern): Boolean = {
+      val features = pattern.features
       lazy val hasSc3Case = pattern.prevSiblingNotWhitespace.exists(_.textMatches(ScalaModifier.CASE))
       lazy val hasSc3IrrefutablePatternsEnabled =
-        pattern.features.hasSourceFutureFlag || pattern.scalaLanguageLevel.exists(_ >= ScalaLanguageLevel.Scala_3_4)
+        features.hasSourceFutureFlag || pattern.scalaLanguageLevel.exists(_ >= ScalaLanguageLevel.Scala_3_4)
       lazy val isRefutablePattern = !pattern.isIrrefutableFor(if (forDisplay) pattern.expectedType else None)
 
-      if (pattern.isScala3OrSource3Enabled) {
+      if (features.isScala3) {
         hasSc3Case || (!hasSc3IrrefutablePatternsEnabled && isRefutablePattern)
       } else {
         isRefutablePattern
--- a/scala/scala-impl/src/org/jetbrains/plugins/scala/lang/psi/impl/expr/ScForImpl.scala
+++ b/scala/scala-impl/src/org/jetbrains/plugins/scala/lang/psi/impl/expr/ScForImpl.scala
@@ -8,7 +8,7 @@ import com.intellij.psi.tree.IElementType
 import org.jetbrains.plugins.scala.ScalaBundle
 import org.jetbrains.plugins.scala.caches.{BlockModificationTracker, cached}
 import org.jetbrains.plugins.scala.extensions.{Model, ObjectExt, PsiElementExt, StringsExt}
-import org.jetbrains.plugins.scala.lang.lexer.ScalaTokenTypes
+import org.jetbrains.plugins.scala.lang.lexer.{ScalaModifier, ScalaTokenTypes}
 import org.jetbrains.plugins.scala.lang.psi.ScalaPsiUtil
 import org.jetbrains.plugins.scala.lang.psi.api.base.patterns._
 import org.jetbrains.plugins.scala.lang.psi.api.expr._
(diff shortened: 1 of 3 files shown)


scala/scala-impl/src/org/jetbrains/plugins/scala/lang/psi/impl/expr/ScForImpl.scala
```diff
@@ -183,8 +183,15 @@ class ScForImpl(node: ASTNode) extends ScExpressionImplBase(node) with ScFor wit
       normalizeUnderscores(expr).getText
     }
 
-    def needsPatternMatchFilter(pattern: ScPattern): Boolean =
-      !pattern.isIrrefutableFor(if (forDisplay) pattern.expectedType else None)
+    def needsPatternMatchFilter(pattern: ScPattern): Boolean = {
+      val hasSc3Case = pattern.isScala3OrSource3Enabled &&
```
disordered: It currently works without this check, however, with Scala 3, when pattern starts with `case`, compiler will require the right hand to have `withFilter`.
I believe this check should stay for this reason.

disordered:
It doesn't address the case where type narrowing occurs (couldn't figure out how to detect type narrowing):
```scala
for y :: ys <- List(List(1), List(2)) do println(y)
```

IntelliJ will not produce an error for Scala 3.4, but compiler will produce this:
```text
pattern's type ::[Int] is more specialized than the right hand side expression's type List[Int]

If the narrowing is intentional, this can be communicated by adding the `case` keyword before the full pattern,
which will result in a filtering for expression (using `withFilter`).
This patch can be rewritten automatically under -rewrite -source 3.2-migration.
    for y :: ys <- List(List(1), List(2)) do println(y)
```
and Scala 3.0+ will produce the same but as a warning.

unkarjedy:
Please see this part of our [guidelines](https://github.com/JetBrains/intellij-scala/blob/1c021da910b7621d45193a3a587770a07964b80c/CONTRIBUTING.md#some-guidelines-for-code-commits-and-pull-requests)

> Every commit message should reference the YouTrack issue number in format #SCL-XXXXX
Place primary issue number in the end of the first message line
If you have multiple equally-important related issues place them in the end of the first message line
If you have some non-primary, but related issues you can reference them in the commit message body.
In the last commit of a series which fixes an issue, append #SCL-XXXXX fixed to the end of the first message line issue which is fixed. This will close the YouTrack issues automatically when the change is merged into main branch.

Right now the commit message doesn't mention YT issues.
I see the mention in the PR, we can just copy it in the commit itself

scala/scala-impl/test/org/jetbrains/plugins/scala/annotator/ForComprehensionHighlightingTest.scala
```diff
@@ -506,4 +505,96 @@ class ForComprehensionSemicolonTest extends ForComprehensionHighlightingTestBase
...
     assertEquals(2 + 2 + 3 + 3, errors("for{\n; \n; \nx <- Seq(1)\n;\n;\n;\nif x == 3\n;\n;;;\n y = x\n;;\n;} ()"))
-}
\ No newline at end of file
+}
+
+class ForComprehensionRefutableTest_3 extends ForComprehensionHighlightingTestBase {
+
+  import Message.Error
+
+  override protected def supportedIn(version: ScalaVersion): Boolean = version >= LatestScalaVersions.Scala_3_0
```
unkarjedy: Let's use the exact version `Scala_3_3`

scala/scala-impl/test/org/jetbrains/plugins/scala/annotator/ForComprehensionHighlightingTest.scala
```diff
@@ -506,4 +505,96 @@ class ForComprehensionSemicolonTest extends ForComprehensionHighlightingTestBase
...
+        |for
+        |  (a, b) <- Right("" -> 1)
+        |yield ()
+      """.stripMargin
+
+    assertNoErrors(code)
+  }
+}
+
+class ForComprehensionIrrefutableTest_3_future extends ForComprehensionHighlightingTestBase {
```
unkarjedy: Here as well, let's use the exact version `Scala_3_3`

unkarjedy: Also, to ensure all the cases are covered, maybe let's inherit from `ForComprehensionRefutableTest_3` and override needed tests

(same for Scala 3.4 tests)

scala/scala-impl/src/org/jetbrains/plugins/scala/lang/psi/impl/expr/ScForImpl.scala
```diff
@@ -183,8 +183,15 @@ class ScForImpl(node: ASTNode) extends ScExpressionImplBase(node) with ScFor wit
...
+        pattern.prevSiblingNotWhitespace.exists(_.getText == ScalaModifier.CASE)
```
unkarjedy: As IntelliJ suggests, please replace `_.getText == ScalaModifier.CASE` with `_.textMatches(ScalaModifier.CASE)`

disordered:
> Please see this part of our [guidelines](https://github.com/JetBrains/intellij-scala/blob/1c021da910b7621d45193a3a587770a07964b80c/CONTRIBUTING.md#some-guidelines-for-code-commits-and-pull-requests)
> 
> > Every commit message should reference the YouTrack issue number in format #SCL-XXXXX
> > Place primary issue number in the end of the first message line
> > If you have multiple equally-important related issues place them in the end of the first message line
> > If you have some non-primary, but related issues you can reference them in the commit message body.
> > In the last commit of a series which fixes an issue, append #SCL-XXXXX fixed to the end of the first message line issue which is fixed. This will close the YouTrack issues automatically when the change is merged into main branch.
> 
> Right now the commit message doesn't mention YT issues. I see the mention in the PR, we can just copy it in the commit itself

Want me to force update first commit or just touch up the PR title and description?
Also, should I mention the 22468, if it will auto close it? I think that narrowing should be addressed as well, to make intellij highlight the error, but maybe in a separate PR.

scala/scala-impl/test/org/jetbrains/plugins/scala/annotator/ForComprehensionHighlightingTest.scala
disordered: Do you mean only to bump to 3.3 or change it to `version == LatestScalaVersions.Scala_3_3`? 
If the latter, why? These tests should pass for all Scala 3 versions (as of right now at least).

EDIT: I just realised there are some incorrect assumptions in the test cases. I'll restructure the tests and then let's take another round on the changes.

scala/scala-impl/src/org/jetbrains/plugins/scala/lang/psi/impl/expr/ScForImpl.scala
disordered: Changed, though, I don't get that suggestion. Possibly, because I don't have access to plugin `com.jetbrains.intellij.api.watcher`.

scala/scala-impl/src/org/jetbrains/plugins/scala/lang/psi/impl/expr/ScForImpl.scala
unkarjedy: It seems like the check **is** required, but currently not covered with tests.
It will be more clear once [comment](https://github.com/JetBrains/intellij-scala/pull/660/files#r1584668155) is addressed

unkarjedy: Is `OrSource3Enabled` part needed in this check?
AFAIU the logic should be only about Scala 3 and not about Scala 2.
`pattern.isInScala3File` should do.

Or even extract `val features = pattern.features` and use `features.isScala3`, `features.hasSourceFutureFlag`

unkarjedy:
> Want me to force update first commit or just touch up the PR title and description?

Force push to your branch should be ok

unkarjedy:
> Also, should I mention the 22468, if it will auto close it?

You can omit the "fixed" part and just mention the ticket.
This is still used by our automation

>I think that narrowing should be addressed as well, to make intellij highlight the error, but maybe in a separate PR.

Sounds like a good plan.
This issue is less critical compared to irrefutability

scala/scala-impl/src/org/jetbrains/plugins/scala/lang/psi/impl/expr/ScForImpl.scala
unkarjedy: Ah, right the inspection should work in internal mode.
https://github.com/JetBrains/intellij-scala/blob/7a050b30da86b846ca88fa92f51f875b7b805c1d/README.md#prerequisites

> (optional but recommended)
Enable [internal mode](https://plugins.jetbrains.com/docs/intellij/enabling-internal.html) in IDEA to get access to helpful internal actions and debug information

disordered:
Alright, all comments should be addressed now (including first commit messaging).
I restructured test inheritance a bit. Also, changed the check logic for `withFilter` requirement as I realized that bulk of tests are run under Scala `2.10` and was not behaving correctly under `3.3` and `3.4`.

Should I do something about it?
One option could be to add `@RunWithScalaVersions(Array(...))` on `ForComprehensionHighlightingTest`. If so, I guess with versions `2.13`, `3.3` and `3.4`?

unkarjedy:
Thanks for the contribution!
I will manually cherry-pick the changes and run through our internal CI.
We might even include it in 2024.1 EAP/Release

> Should I do something about it?
One option could be to add @RunWithScalaVersions(Array(...)) on ForComprehensionHighlightingTest. If so, I guess with versions 2.13, 3.3 and 3.4?

No, thanks. I will deal with it.

PR #665, opened 2024-08-15
Remove CMSClassUnloadingEnabled from default SBT flags

andrzejressel:
This setting was removed from Java 17 and it now throws "Unrecognized VM option 'CMSClassUnloadingEnabled'"

Should I also create YouTrack ticket for this?


unkarjedy:
Could you also patch `com.intellij.scala.play.runConfigurations.SbtLauncherCommandLineBuilder#DefaultVmOptions`?

andrzejressel:
I cannot find this file, where is it?

unkarjedy:
Sorry for the long reply. 
I will manually do that patch in our small close source part of the plugin and run it though our CI.
Thanks for the contirbution.

andrzejressel:
I think it was merged: https://github.com/JetBrains/intellij-scala/commit/77d7e7fc3a8d2b4fb69a3e104c36c18e0b324150

PR #666, opened 2024-09-04
Add `.mill` extension to Scala language support

lihaoyi:
The Mill Scala build tool recently moved its file format extension from `.sc` to `.mill` (https://github.com/com-lihaoyi/mill/pull/3426) to better differentiate it from other Scala scripts (Ammonite, Scala-CLI, IntelliJ/VSCode worksheets). This updates intellij to understand that `.mill` files should be highlighted as Scala

PR #667, opened 2024-09-13
Support mill extension as scala file

azdrojowa123:
The whole context is in https://github.com/JetBrains/intellij-scala/pull/666


lihaoyi (commented):
Looks good to me, and seemed to work when I tried it locally!

azdrojowa123:
@lihaoyi 
It should be available in the nightly plugin 2024.2 and will be available in the 2024.3 release :)

PR #670, opened 2024-10-22
Update scalafmt default version to 3.8.3

987Nabil:
If one activates scalafmt formatting without a config file, currently 1.5.1 is downloaded. That is a very old version and does not support Scala 3. I don't see a reason to not update this.

### DIFF
--- a/scala/scala-impl/src/org/jetbrains/plugins/scala/lang/formatting/scalafmt/ScalafmtDynamicService.scala
+++ b/scala/scala-impl/src/org/jetbrains/plugins/scala/lang/formatting/scalafmt/ScalafmtDynamicService.scala
@@ -44,7 +44,7 @@ trait ScalafmtDynamicService {
 
 object ScalafmtDynamicService {
 
-  val DefaultVersion = ScalafmtVersion(1, 5, 1)
+  val DefaultVersion = ScalafmtVersion(3, 8, 3)
 
   def instance: ScalafmtDynamicService = ApplicationManager.getApplication.getService(classOf[ScalafmtDynamicService])
 


vasilmkd:
Hi, thanks for your contribution.
We tried doing this a while ago. Unfortunately, new scalafmt versions (after 3.1.0) require manually specifying a Scala dialect.
https://scalameta.org/scalafmt/docs/configuration.html#scala-dialects

You would essentially have to build some sort of default given the Scala version in the project.
I think it would be fine, as a default (which may not be 100% correct, and power users can still adjust the generated file manually) to do the following mapping:

Scala 2.11 -> `scala211`
Scala 2.12 -> `scala212`
Scala 2.13 -> `scala213`
Scala 3 -> `scala3`

Would you like to continue with this work? Thanks again.

987Nabil:
I'll take a look

hmemcpy:
I'll throw another possible wrench into this: there's also `-Xsource:3` compatibility.

vasilmkd:
Which is exactly why we haven't allocated the time to do it yet. 🥶

There are edge cases to work through.

unkarjedy:
Some useful pointers:
Extension methods to find the Scala version of the current module:
- `org.jetbrains.plugins.scala.project.ModuleExt#scalaMinorVersion`
- `org.jetbrains.plugins.scala.project.ModuleExt#languageLevel`

Extension methods to detect `-xsource`
- `org.jetbrains.plugins.scala.project.ModuleExt#isSource3Enabled`
- `org.jetbrains.plugins.scala.project.ModuleExt#source3Options` 
(not sure how scalafmt handles these, though)

There is currently limited support for multiple scalafmt configuration in the project.
This can be the case when there are multiple Scala versions in the project.
See https://youtrack.jetbrains.com/issue/SCL-16675/Add-an-option-to-use-.scalafmt.conf-from-the-module-in-multi-module-projects

So, the fix would probably work only for the case when there is one Scala version used in the project.
You could take the module representing the root directory, assuming that it has the Scala SDK with the "primary" version.
Though it might be not always true. E.g. the root module might have an empty sources directory, and the main code can be located in subprojects. In this case we might find the "Most frequently used Scala SDK version" among the modules.

987Nabil:
just fyi, I'll still take a look, but was quite busy. I hope I can find some time before the year ends.

mkurz:
Sonst usw 3.8.3, it's broken. Wait for next release.

PR #672, opened 2024-11-12
Fix SCL-23198: Direct references to package objects should be allowed in `.mill` files

lihaoyi:
Fixes https://youtrack.jetbrains.com/issue/SCL-23198/Direct-references-to-package-objects-should-be-allowed-in-.mill-files

Tested manually, seems to make the problematic highlight go away

Screenshot 2024-11-12 at 10 45 08 AM

### DIFF
--- a/scala/scala-impl/src/org/jetbrains/plugins/scala/lang/psi/impl/expr/ScReferenceExpressionImpl.scala
+++ b/scala/scala-impl/src/org/jetbrains/plugins/scala/lang/psi/impl/expr/ScReferenceExpressionImpl.scala
@@ -195,7 +195,10 @@ class ScReferenceExpressionImpl(node: ASTNode) extends ScReferenceImpl(node) wit
       case inf: ScInfixExpr if this == inf.operation || this == inf.getBaseExpr =>
         StdKinds.refExprQualRef
       case _ =>
-        StdKinds.refExprLastRef
+        // Mill files allow direct references to package
+        // objects, even though normal .scala files do not
+        if (this.containingScalaFile.exists(_.isMillFile)) StdKinds.refExprQualRef
+        else StdKinds.refExprLastRef
     }
   }
 
--- a/scala/scala-impl/test/org/jetbrains/plugins/scala/annotator/MillHighlightingTests.scala
+++ b/scala/scala-impl/test/org/jetbrains/plugins/scala/annotator/MillHighlightingTests.scala
@@ -1,0 +1,24 @@
+package org.jetbrains.plugins.scala.annotator
+
+/**
+ * Check customized behavior for Mill build files
+ */
+class MillHighlightingTests extends ScalaHighlightingTestBase {
+  //SCL-23198
+  def testAllowDirectPackageReferences(): Unit = {
+    val code = """package foo{
+                 |  object `package`
+                 |}
+                 |
+                 |package bar{
+                 |  object qux{
+                 |    val reference = foo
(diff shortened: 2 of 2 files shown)


unkarjedy:
Please add at least a simple test covering the change (checking that it fails without the changes).
Our test code is a little bit of a mess and it can be hard to navigate through it.

As a base test class you can use one of these:
- `ScalaHighlightingTestBase`
(see example `org.jetbrains.plugins.scala.annotator.Scala3HighlightingTestsMix`)
- `ScalaLightCodeInsightFixtureTestCase` + mixin `ScalaHighlightingTestLike`
(see example `org.jetbrains.plugins.scala.annotator.ScTemplateDefinitionAnnotatorTest`)

lihaoyi:
@unkarjedy I tried to add a simple test in https://github.com/JetBrains/intellij-scala/pull/672/commits/37cf7b9d10147fc2217a6764d2d62a30553aef9e. However, I am having trouble running tests, where `JAVA_HOME=/Library/Java/JavaVirtualMachines/amazon-corretto-17.jdk/Contents/Home/ sbt "scala-impl/testOnly org.jetbrains.plugins.scala.annotator.MillHighlightingTests"` is hanging for over 10 minutes and doesn't seem to be making any progress even before getting to my own test code. Is there a different command I should be using to run individual test classes in the test suite?

unkarjedy:
Hmm, I don't know why it hangs.
When I try to run
```
JAVA_HOME=/Users/dmitrii.naumenko/Library/Java/JavaVirtualMachines/corretto-17.0.13/Contents/Home sbt "scala-impl/testOnly org.jetbrains.plugins.scala.annotator.ScTemplateDefinitionAnnotatorTest"
```
I see the error in 10 seconds:
```
[error] Expected ID character
[error] Not a valid command: scala-impl
[error] Expected project ID
[error] Expected configuration
[error] Expected ':'
[error] Expected key
[error] Not a valid key: scala-impl (similar: scalaHome)
[error] scala-impl/testOnly org.jetbrains.plugins.scala.annotator.ScTemplateDefinitionAnnotatorTest
[error]      
```

Our project has a non-standard sbt configuration as the packaging mechanism for IntelliJ plugins is pretty unique.
You don't need to add a project prefix to the command.
You can use the test command directly in the root project.
```
JAVA_HOME=/Users/dmitrii.naumenko/Library/Java/JavaVirtualMachines/corretto-17.0.13/Contents/Home sbt "testOnly org.jetbrains.plugins.scala.annotator.ScTemplateDefinitionAnnotatorTest"
 ```
 
 If the process is still, stuck, could you share the logs?
 Which line does it hang on?
 
 We in the team mostly run the test directly from IDEA though:
 
image

lihaoyi:
Running from the IDE works. Not sure why our SBT executions are different, but if IDE is the blessed workflow I can do that.

I have updated the PR with tests and verify they fail on the base branch

unkarjedy:
But does the updated sbt command hang as well?
Can you share the sbt output when it hangs?

vasilmkd:
Please run `testOnly` instead of `scala-impl/testOnly`. Tests cannot be run within modules, only in the root. Thank you.

azdrojowa123 (approved):

azdrojowa123:
It was merged with our internal CI. 
You can expect this in the next nightly version and in the next minor 2024.3 release (in about a week and a half). 

Thank you for the contribution :)

PR #674, opened 2024-11-17
Mill support: enable importing projects which contain build.mill(.scala) but not wrapper script

sideeffffect:
Not all Mill projects necessarily have a wrapper script in the root. The wrapper script may be on the PATH.

But _all_ Mill projects contain `build.mill` or `build.mill.scala`, so that's _the_ robust way to recognize them.

/cc @lihaoyi

### DIFF
--- a/bsp/src/org/jetbrains/bsp/project/importing/MillProjectInstaller.scala
+++ b/bsp/src/org/jetbrains/bsp/project/importing/MillProjectInstaller.scala
@@ -16,7 +16,11 @@ class MillProjectInstaller extends BspProjectInstallProvider {
 
   override def canImport(workspace: File): Boolean =
     Option(workspace) match {
-      case Some(directory) if directory.isDirectory => isBspCompatible(directory) || isLegacyBspCompatible(directory) || BspUtil.findFileByName(directory, "build.mill").isDefined || BspUtil.findFileByName(directory, "build.mill.scala").isDefined || BspUtil.findFileByName(directory, "build.sc").isDefined
+      case Some(directory) if directory.isDirectory =>
+        isBspCompatible(directory) ||
+          isLegacyBspCompatible(directory) ||
+          BspUtil.findFileByName(directory, "build.mill").isDefined ||
+          BspUtil.findFileByName(directory, "build.mill.scala").isDefined
       case _ => false
     }
 
@@ -34,7 +38,7 @@ class MillProjectInstaller extends BspProjectInstallProvider {
         Success(Seq(file.getAbsolutePath, "-i", "mill.bsp.BSP/install"))
       case Some(file) if isLegacyMill =>
         Success(Seq(file.getAbsolutePath, "-i", "mill.contrib.BSP/install"))
-      case None =>
+      case _ =>
         Success(Seq("mill", "-i", "mill.bsp.BSP/install"))
     }
   }
--- a/bsp/src/org/jetbrains/bsp/project/importing/MillProjectInstaller.scala
+++ b/bsp/src/org/jetbrains/bsp/project/importing/MillProjectInstaller.scala
@@ -35,10 +35,13 @@ class MillProjectInstaller extends BspProjectInstallProvider {
     val millFileOpt = getMillFile(workspace)
     millFileOpt match {
       case Some(file) if isLegacyMill && !isMillFileBspCompatible(file, workspace) =>
+        // run this only if we're confident this is legacy Mill
         Seq(file.getAbsolutePath, "-i", "mill.contrib.BSP/install")
-      case Some(file) => // executes if isMillFileBspCompatible
+      case Some(file) =>
+        // otherwise run the normal BSP install command
         Seq(file.getAbsolutePath, "-i", "mill.bsp.BSP/install")
       case _ =>
+        // as a fallback, use Mill from PATH in case we couldn't find launcher in the project root
         Seq("mill", "-i", "mill.bsp.BSP/install")
     }
   }
--- a/bsp/src/org/jetbrains/bsp/project/importing/MillProjectInstaller.scala
+++ b/bsp/src/org/jetbrains/bsp/project/importing/MillProjectInstaller.scala
@@ -16,7 +16,7 @@ class MillProjectInstaller extends BspProjectInstallProvider {
 
   override def canImport(workspace: File): Boolean =
     Option(workspace) match {
-      case Some(directory) if directory.isDirectory => isBspCompatible(directory) || isLegacyBspCompatible(directory) || BspUtil.findFileByName(directory, "build.mill").isDefined || BspUtil.findFileByName(directory, "build.mill.scala").isDefined
+      case Some(directory) if directory.isDirectory => isBspCompatible(directory) || isLegacyBspCompatible(directory) || BspUtil.findFileByName(directory, "build.mill").isDefined || BspUtil.findFileByName(directory, "build.mill.scala").isDefined || BspUtil.findFileByName(directory, "build.sc").isDefined
       case _ => false
     }
 
--- a/bsp/src/org/jetbrains/bsp/project/importing/MillProjectInstaller.scala
+++ b/bsp/src/org/jetbrains/bsp/project/importing/MillProjectInstaller.scala
@@ -16,7 +16,7 @@ class MillProjectInstaller extends BspProjectInstallProvider {
 
   override def canImport(workspace: File): Boolean =
     Option(workspace) match {
-      case Some(directory) if directory.isDirectory => isBspCompatible(directory) || isLegacyBspCompatible(directory)
+      case Some(directory) if directory.isDirectory => isBspCompatible(directory) || isLegacyBspCompatible(directory) || BspUtil.findFileByName(directory, "build.mill").isDefined || BspUtil.findFileByName(directory, "build.mill.scala").isDefined
       case _ => false
     }
 
@@ -34,7 +34,8 @@ class MillProjectInstaller extends BspProjectInstallProvider {
         Success(Seq(file.getAbsolutePath, "-i", "mill.bsp.BSP/install"))
       case Some(file) if isLegacyMill =>
         Success(Seq(file.getAbsolutePath, "-i", "mill.contrib.BSP/install"))
-      case _ => Failure(new IllegalStateException("Unable to install BSP as this is not a Mill project"))
+      case None =>
+        Success(Seq("mill", "-i", "mill.bsp.BSP/install"))
     }
   }
 
--- a/bsp/src/org/jetbrains/bsp/project/importing/MillProjectInstaller.scala
+++ b/bsp/src/org/jetbrains/bsp/project/importing/MillProjectInstaller.scala
@@ -17,10 +17,10 @@ class MillProjectInstaller extends BspProjectInstallProvider {
   override def canImport(workspace: File): Boolean =
     Option(workspace) match {
       case Some(directory) if directory.isDirectory =>
-        isBspCompatible(directory) ||
-          isLegacyBspCompatible(directory) ||
-          BspUtil.findFileByName(directory, "build.mill").isDefined ||
-          BspUtil.findFileByName(directory, "build.mill.scala").isDefined
+        BspUtil.findFileByName(directory, "build.mill").isDefined ||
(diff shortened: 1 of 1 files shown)


nafg:
What about pre-0.12 projects, they would have `build.sc`

sideeffffect:
They would need to update to >= 0.12? My attitude towards this is "_who cares about old Mill_"?

But if there's an actual demand, I'm not opposed. And the change is trivial, so I actually did it :)

nafg:
"old mill"

Right now I'm blocked on https://github.com/com-lihaoyi/mill/issues/3955, which was fixed but needs to be released as 0.12.3

sideeffffect:
@nafg fair enough.

Btw I thought I did it, but I didn't, so I've made the change now: `build.sc` is now also used to detect if it's a Mill project :+1:

sideeffffect:
Hello @azdrojowa123 @vasilmkd @SrTobi @timaliberdov @unkarjedy , could you please have a look at this PR? :pray:

lefou (commented):
Looks good to me.

bsp/src/org/jetbrains/bsp/project/importing/MillProjectInstaller.scala
```diff
@@ -34,7 +34,8 @@ class MillProjectInstaller extends BspProjectInstallProvider {
         Success(Seq(file.getAbsolutePath, "-i", "mill.bsp.BSP/install"))
       case Some(file) if isLegacyMill =>
         Success(Seq(file.getAbsolutePath, "-i", "mill.contrib.BSP/install"))
-      case _ => Failure(new IllegalStateException("Unable to install BSP as this is not a Mill project"))
+      case None =>
```
unkarjedy: I would leave it `case _` not to think about exhaustivity and get unexpected errors at runtime

bsp/src/org/jetbrains/bsp/project/importing/MillProjectInstaller.scala
```diff
@@ -16,7 +16,7 @@ class MillProjectInstaller extends BspProjectInstallProvider {
 
   override def canImport(workspace: File): Boolean =
     Option(workspace) match {
-      case Some(directory) if directory.isDirectory => isBspCompatible(directory) || isLegacyBspCompatible(directory)
+      case Some(directory) if directory.isDirectory => isBspCompatible(directory) || isLegacyBspCompatible(directory) || BspUtil.findFileByName(directory, "build.mill").isDefined || BspUtil.findFileByName(directory, "build.mill.scala").isDefined || BspUtil.findFileByName(directory, "build.sc").isDefined
```
unkarjedy: 1. The line with the condition is too long and could be better formatted (every `||` condition on a new line)

2. Given that just on a single line the same patter is repeated 3 times: `BspUtil.findFileByName(directory, fileName).isDefined` it's worth extracting into a method.
I looked into the implementation of `findFileByName` and it calls `listFiles` every time.
So it's worth having another method, something like `def directoryContainsFile(directory: Path, fileNames: Seq[String]): Boolean` that would do one iteration over the directory files.

unkarjedy: AFAIU `BspUtil.findFileByName(directory, "build.sc").isDefined` checks for the "old mill"?
But we already have `isLegacyBspCompatible` method, so this line isn't needed, is it?
AFAIU `isLegacyBspCompatible` also avoids some false positives when a directory has `build.sc` file which doesn't represent a mill build file by checking for a magic line `import $ivy.com.lihaoyi::mill-contrib-bsp:$MILL_VERSION`

bsp/src/org/jetbrains/bsp/project/importing/MillProjectInstaller.scala
```diff
@@ -34,7 +34,8 @@ class MillProjectInstaller extends BspProjectInstallProvider {
         Success(Seq(file.getAbsolutePath, "-i", "mill.bsp.BSP/install"))
       case Some(file) if isLegacyMill =>
         Success(Seq(file.getAbsolutePath, "-i", "mill.contrib.BSP/install"))
-      case _ => Failure(new IllegalStateException("Unable to install BSP as this is not a Mill project"))
+      case None =>
+        Success(Seq("mill", "-i", "mill.bsp.BSP/install"))
```
unkarjedy: Did you check how it would look on UI when there is also no mill available on PATH?
The process will fail with some non zero exit code, but will there be a meaningful error message in the UI?

bsp/src/org/jetbrains/bsp/project/importing/MillProjectInstaller.scala
lefou: > y checking for a magic line `import $ivy.`com.lihaoyi::mill-contrib-bsp:$MILL_VERSION``

That line isn't required for BSP to work as intended for many Mill versions.

lefou: Instead, just looking for Mill imports or uses might be a better heuristic.

Finding a `.mill-version` (or `.config/mill-version`) file next to `build.sc` is also strong indicator that you found a Mill project.

unkarjedy: Ok, got it.

My main point is that the new code: `BspUtil.findFileByName(directory, "build.sc").isDefined` 
makes the `isLegacyBspCompatible` method call redundant.
Inside that method there is `BspUtil.findFileByName(workspace, "build.sc").exists(<has some magic line>)`.
So `isLegacyBspCompatible || BspUtil.findFileByName(directory, "build.sc").isDefined` 
is in practice the same as `BspUtil.findFileByName(directory, "build.sc").isDefined`.

I would remove the new bit and enhance the `isLegacyBspCompatible` method according to your vision in the comment above.
If the name is not the most appropriate, in your opinion, then feel free to propose a better one.

unkarjedy: (BTW, thanks for contributing!)

bsp/src/org/jetbrains/bsp/project/importing/MillProjectInstaller.scala
sideeffffect: done

bsp/src/org/jetbrains/bsp/project/importing/MillProjectInstaller.scala
sideeffffect: I didn't. I assume it will fail the same way as when `Seq(file.getAbsolutePath, "-i", "mill.bsp.BSP/install")` fails -- which is it will print the errors in the terminal (build view).

bsp/src/org/jetbrains/bsp/project/importing/MillProjectInstaller.scala
sideeffffect: There's no point in making it even more sophisticated. Once there is a file `build.mill` (or `build.mill.scala`), we can be fairly certain that we're dealing with a Mill project. No need to search for `.mill-version` or anything like that.

lefou: Sure, the `.mill-version` lookup was only suggested in case you need to disambiguate the purpose of a detected `build.sc`, since it is common for various tools: Mill, but also Scala-CLI, Ammonite, workbooks.

bsp/src/org/jetbrains/bsp/project/importing/MillProjectInstaller.scala
```diff
@@ -16,25 +16,30 @@ class MillProjectInstaller extends BspProjectInstallProvider {
...
     val millFileOpt = getMillFile(workspace)
     millFileOpt match {
-      case Some(file) if isMillFileBspCompatible(file, workspace) =>
-        Success(Seq(file.getAbsolutePath, "-i", "mill.bsp.BSP/install"))
-      case Some(file) if isLegacyMill =>
-        Success(Seq(file.getAbsolutePath, "-i", "mill.contrib.BSP/install"))
-      case _ => Failure(new IllegalStateException("Unable to install BSP as this is not a Mill project"))
+      case Some(file) if isLegacyMill && !isMillFileBspCompatible(file, workspace) =>
+        Seq(file.getAbsolutePath, "-i", "mill.contrib.BSP/install")
+      case Some(file) => // executes if isMillFileBspCompatible
```
unkarjedy: Now the comment:
```scala
// executes if isMillFileBspCompatible
```
is read as if the branch is only executed when `isMillFileBspCompatible == true`.
But AFAIU it's not the case: if `isLegacyMill` is `false` and `isMillFileBspCompatible` is `false` then the branch is still executed.

>is read as if the branch is only executed when `isMillFileBspCompatible == true`.

If it's indeed what is implied, it's better to return the original explicit check as it's less obfuscated.
If it's not what is implied, then the comment is probably not needed?

sideeffffect: I've improved the comment. It now gives a better idea what it's aiming for.

The code itself is now more robust. It tries as much as possible to import the Mill build, with the best the user has.

bsp/src/org/jetbrains/bsp/project/importing/MillProjectInstaller.scala
lefou: `build.sc` is still supported by Mill 0.12.

sideeffffect: Oh, sure, I'm not removing the support `build.sc`. I'm just adding detection for `build.mill`.

sideeffffect:
Hello everybody, what can I do to help this move forward?

I see that there is an error in CI, but it doesn't seem related to this PR:
 * https://github.com/JetBrains/intellij-scala/actions/runs/12679537151/job/35360743666?pr=674#step:9:25219

Can you please re-run?

unkarjedy:
The tests failure is indeed unrelated and is actually muted on our CI.

unkarjedy:
@azdrojowa123 Could you or Antoni please take a look at it and merge it to 243 Nightly via personal branch?

bsp/src/org/jetbrains/bsp/project/importing/MillProjectInstaller.scala
azdrojowa123: I decided to check this and have some thoughts - in the "old implementation" when `Success(Seq(file.getAbsolutePath, "-i", "mill.contrib.BSP/install"))` was executed indeed the errors were displayed in the terminal, but because we were sure that the mill script existed we can assume that these errors were useful (at least to some certain level). 
Now If the Mill is not installed globally, a very unclear error message is displayed in the terminal. 
Screenshot 2025-01-22 at 13 51 41

To prevent this, we can implement the same solution as in the `ScalaCliProjectInstaller`. In the `#installCommand`, there should be a check to verify if Mill is installed globally. If it's not, a more meaningful error message should be displayed. Perhaps the best approach would be to add this check to the `#canImport` method, but for now, let's stay consistent with the implementation in `ScalaCliProjectInstaller`. If needed, it can always be refactored and moved in the future.

sideeffffect: > In the `#installCommand`, there should be a check to verify if Mill is installed globally.

I'm not against. But I'm not familiar with the code base. Can you please make this improvement _after_ this PR is merged?

azdrojowa123: @sideeffffect 
I will add this before merging (anyway I have to pass it on my own through our internal CI) but I'll take care of it tomorrow :)

azdrojowa123:
Hello, 
It’s merged and will be available in the next IntelliJ IDEA 2024.3 minor release (likely in a week or two). Alternatively, you can try it out in IntelliJ IDEA 2025.1 EAP (Early Access Program), which you can [download](https://www.jetbrains.com/idea/nextversion/) directly or via the [Toolbox App](https://www.jetbrains.com/toolbox-app/), with the Scala plugin version 2025.1.446 or newer.

sideeffffect:
Thank you everybody :bow:

PR #675, opened 2025-03-13
fix #IDEA-365234

Alex1005a:
I don't know why [this](https://youtrack.jetbrains.com/issue/IDEA-365234/Scala-Implicit-argument-pop-up-does-not-show-for-method-with-implicit-argument) is in the IDEA issue tracker. It's a Scala plugin issue.


unkarjedy:
Thanks for the PR!
I will manually merge it via our internal CI

unkarjedy (approved):

unkarjedy:
The fix will be available in the next 2025.1 Nightly/EAP/Release versions in the order they are published
Thanks again!

PR #676, opened 2025-03-19
Support `SIP-68: Reference-able Package Objects`

lihaoyi:
Follow up to https://github.com/JetBrains/intellij-scala/pull/672/files. SIP-68 has been approved for experimental use, and will probably become preview/final at some point later, so opening this PR to get ready for that.

Not sure if there's a better way to do this, e.g. if we want to continue supporting the older behavior for older Scala versions. This is just a first draft to start the discussion


azdrojowa123:
Thank you for the PR. 
I've created a [YouTrack ticket](https://youtrack.jetbrains.com/issue/SCL-23743/Support-SIP-68-Reference-able-Package-Objects-3.7.0) for this issue.

SrTobi:
@lihaoyi I fixed this a little bit differently, but it should work starting from Scala 3.7. Thanks for the PR

PR #677, opened 2025-03-23
fix #SCL-23707

Alex1005a:
To fix the problem I had to change the parsing of interpolated string. I'm not sure if it's correct, but I couldn't find another way.

https://youtrack.jetbrains.com/issue/SCL-23707

### DIFF
--- a/scala/scala-impl/src/org/jetbrains/plugins/scala/lang/parameterInfo/ScalaFunctionParameterInfoHandler.scala
+++ b/scala/scala-impl/src/org/jetbrains/plugins/scala/lang/parameterInfo/ScalaFunctionParameterInfoHandler.scala
@@ -11,13 +11,14 @@ import org.jetbrains.plugins.scala.lang.lexer.ScalaTokenTypes
 import org.jetbrains.plugins.scala.lang.parameterInfo.ScalaFunctionParameterInfoHandler.{AnnotationParameters, UniversalApplyCall, UniversalApplyCallContext}
 import org.jetbrains.plugins.scala.lang.psi.api.ScalaPsiElement
 import org.jetbrains.plugins.scala.lang.psi.api.base.types.{ScParameterizedTypeElement, ScTypeArgs, ScTypeElementExt}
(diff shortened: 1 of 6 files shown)


unkarjedy:
Thank you for the contribution and sorry for the late reply!
Please add tests to all the effected subsystems (in this case parser, parameter info, annotator/type inference (original issue is the red code))

unkarjedy:
I don't think that changing the parser is needed here.
It also goes against Scala definition of what Scala interpolated string is 
https://docs.scala-lang.org/scala3/reference/syntax.html

Alex1005a:
Yes, from the specification point of view it is not correct. I did it this way because interpolated string is parsed first into an intermediate structure, and not immediately [desugar](https://github.com/scala/scala/blob/2.13.x/src/compiler/scala/tools/nsc/ast/parser/Parsers.scala#L1557) as in the scala compiler. Desugaring occurs in the `desugaredExpression` method, and if no arguments are passed to the function from interpolator during it, the type check will fail.
There is probably a more correct way to do this, but I don't know about it, since I am not very familiar with the code base and the project architecture.

SrTobi:
@Alex1005a hi. Thanks for your PR. I used the tests, but ended up not using your implementation. The syntax tree that was generated was already correct. The problem was more that the annotator couldn't handle the de-sugaring correctly. That's why you pulled the arguments into the desugaring. This didn't help with some other cases, though, like

```scala
implicit class CustomInterpolator(private val stringContext: StringContext) {
  def test(values: Any*)(i: Int): Unit = {}
}

// this was still annotated as missing a parameter list, although the compiler allows this
val x = test"" // type of x is then Int => Unit

// you would expect the compiler to complain and demand something like
val x = test"" _
// but that is not the case ¯\_(ツ)_/¯
```

In the end I just tweaked the annotator a little bit to be aware of the de-sugaring and disable missing-parameter clauses for interpolated string literals altogether :grin:

I merged it manually, so this PR can be closed. Thanks again :+1:

PR #678, opened 2025-04-05
Start of using bloop rifle

hamnis:
Should fix https://youtrack.jetbrains.com/issue/SCL-21024/Use-a-locally-installed-bloop-with-provided-connection-file#focus=Comments-27-11406726.0-0

* Remove old bloop launcher
* Fix some warnings which became errors in the build, by adding @scala.annotation.nowarn("cat=deprecation") where needed.

TODO:
* Make the bloop version configurable.

### DIFF
--- a/.idea/kotlinc.xml
+++ b/.idea/kotlinc.xml
@@ -1,13 +1,9 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
-    <component name="Kotlin2JvmCompilerArguments">
-        <option name="jvmTarget" value="17"/>
-    </component>
-    <component name="KotlinCommonCompilerArguments">
-        <option name="apiVersion" value="1.9"/>
-        <option name="languageVersion" value="1.9"/>
-    </component>
-    <component name="KotlinJpsPluginSettings">
-        <option name="version" value="2.1.10"/>
-    </component>
+  <component name="Kotlin2JvmCompilerArguments">
+    <option name="jvmTarget" value="17" />
+  </component>
+  <component name="KotlinJpsPluginSettings">
+    <option name="version" value="2.1.10" />
+  </component>
 </project>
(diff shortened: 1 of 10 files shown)


.idea/kotlinc.xml
hamnis, line 1: i dont understand why this file was edited at all.

azdrojowa123:
Hello, 
I apologize for the late response.
I was able to reproduce the issue on an m1 machine. From what I've checked, it can be fixed simply by upgrading from `bloop-launcher_2.12 1.5.6` to `bloop-launcher-core_2.12 1.6.0`. Such a fix would be much simpler. Does this solution also work for you?

I noticed that you used the new bloop rifle. What are the practical benefits of using it?

hamnis:
The bloop-launcher is no longer maintained, and is replaced by bloop-rifle.

azdrojowa123:
Hello, 
The fix from this PR has been merged via JetBrains internal tools, but I slightly minimized it by removing the transitive dependencies to the newer version of `bsp4j`. It's the [commit](https://github.com/JetBrains/intellij-scala/commit/4558c83b8b96c95e23c8350140ee52a9b984b1aa). 

The fix will be available in the next 2025.2 nightly version. If you're not using nightly versions, you can enable them in the `Settings | Languages & Frameworks | Scala | Updates` settings. 

Thank you again for the contribution.

azdrojowa123:
Because this PR is merged, [SCL-21024](https://youtrack.jetbrains.com/issue/SCL-21024) is also marked as "fixed". 

At the bottom of SCL-21024, there is a note incorrectly indicating that the PR was declined. This happened because the cherry-pick wasn't done with your changes, although they are explicitly mentioned in the merged commit.

hamnis:
Great news. Thanks a lot!

PR #694, opened 2025-05-02
[SCL-23269] Find usages for context bound name

Alex1005a:
I was also wondering if it would be possible to make a unused context bound highlighting, even if it is not named, as in the case of given. Is there a separate SCL for this?


unkarjedy:
@Alex1005a Hello. Did you mention [SCL-23707](https://youtrack.jetbrains.com/issue/SCL-23707) by mistake?

unkarjedy:
@Alex1005a 
Could you please create a YT ticket with the description of the original problem?
It will be easier than reverse-engineering it from the solution.
If there are any similar or duplicate tickets we will figure it out

Alex1005a:
Sorry, I accidentally entered the wrong ticket. The correct one is: [SCL-23269](https://youtrack.jetbrains.com/issue/SCL-23269/Find-usages-doesnt-work-for-context-bound-name).

SrTobi:
Hi @Alex1005a, thanks for the PR. I reviewed it and merged it basically unchanged (I just renamed the commit messages a little). I merged it manually, so I'm closing the pr here. Thanks again :beer:

SrTobi:
Ah damn. And I didn't change the ticket id :grin:

PR #698, opened 2025-05-12
SCL-23855: show type parameters in XRay mode

Alex1005a:
There are some cases when the type parameter is not inferred, and hint simply displays a type variable, for example `A_`. I'll try to figure out these cases, but what to do if the type variable is not inferred: show as is or don't display the hint at all?


unkarjedy:
Hello!
Thanks for the contribution! We will take a look.

A general comment, please get familiar with these guidelines:
https://github.com/JetBrains/intellij-scala/blob/idea251.x/CONTRIBUTING.md#guidelines-for-code-commits-and-pull-requests, including the examples of the commit messages.

unkarjedy:
@pavelfatin Kind ping

unkarjedy:
Sorry, it takes us longer to start looking into it due to other priorities, but we will get back to it soon.

SrTobi:
Hi @Alex1005a, I took your pr, reworked it a little bit and merged it already (See c29aa14489..b30145860a). I fixed some stuff regarding infix calls and apply calls. Regarding the not-inferred types, I just used `ScType.removeAbstracts`, which has the scala compiler semantic in choosing either the upper or lower bound. I also added tests.

Do you have more examples where it didn't quite work for you? That would be helpful to fix further bugs

PR #701, opened 2025-08-02
SCL-24180 Allow 'final' modifier for top-level package declarations

fan-tom:
Closes SCL-24180

See task for details.

TLDR: `final` modifier is currently prohibited in top-level package declarations, like
```scala
package foo
final val x = ???
```
This PR tries to fix this

### DIFF
--- a/scala/scala-impl/src/org/jetbrains/plugins/scala/annotator/modifiers/ModifierChecker.scala
+++ b/scala/scala-impl/src/org/jetbrains/plugins/scala/annotator/modifiers/ModifierChecker.scala
@@ -138,7 +138,7 @@ private[annotator] object ModifierChecker {
                     } else {
                       checkIllegalCombinations(modifierPsi, Final)
                     }
-                  case e: ScMember if e.getParent.is[ScalaFile] =>
+                  case e: ScMember if e.isTopLevel =>
                     checkIllegalCombinations(modifierPsi, Final)
                   case e: ScClassParameter =>
                     if (PsiTreeUtil.getParentOfType(e, classOf[ScTypeDefinition]).hasFinalModifier) {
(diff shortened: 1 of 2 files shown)


scala/scala-impl/src/org/jetbrains/plugins/scala/annotator/modifiers/ModifierChecker.scala
```diff
@@ -138,7 +138,7 @@ private[annotator] object ModifierChecker {
                     } else {
                       checkIllegalCombinations(modifierPsi, Final)
                     }
-                  case e: ScMember if e.getParent.is[ScalaFile] =>
+                  case e: ScMember if e.getParent.is[ScalaFile, ScPackaging] =>
```
fan-tom: I'm not sure it is the best way though, I just looked what's the parent's type here when this warning is emitted and saw this trait

unkarjedy: Please see `ScMember#isTopLevel`

fan-tom: Cool, replaced both type checks with this method. IDK though why more specific type `ScalaFile` was used instead of `ScFile` used in `isTopLevel`, hope it won't break anything

SrTobi: `ScFile` could also be a worksheet file, where normal definitions are not really at the top level. 

Note additionally that `isTopLevel` works for top level definitions that are inside a "package block" ala
```scala
package innerPackage {
  final val test: Int = 3
}
```

Would be good if we could have a test case for that as well :pray:

fan-tom: > ScFile could also be a worksheet file, where normal definitions are not really at the top level.

I see that all subtypes of `ScFile` are also subtypes of `ScalaFile`, so, effectively, I didn't make this branch accept any additional types except `ScPackaging`:
изображение

> Note additionally that isTopLevel works for top level definitions that are inside a "package block"

Yes, and that's ok for our use-case, as such code is valid:
```scala
package outer {
  package inner {
    final val x = ???
  }
}
```
I've extended a test to covert that too

SrTobi:
@fan-tom I manually merged your changes into the bugfix branch. So they should be available in the next bugfix release. And, of course, in the next nightly. Thanks for the pr!

PR #706, opened 2025-10-09
SCL-22406 Single character variable name expanded with "{}" in interpolated string

rochala:
I believe this was caused by a bug in platform which does not increment current start offset if the document.insert().offset is the same as the current start offset.

### DIFF
--- a/scala/scala-impl/src/org/jetbrains/plugins/scala/lang/completion/handlers/ScalaInsertHandler.scala
+++ b/scala/scala-impl/src/org/jetbrains/plugins/scala/lang/completion/handlers/ScalaInsertHandler.scala
@@ -90,6 +90,7 @@ object ScalaInsertHandler {
         .getStartOffset
 
       document.insertString(tailOffset, "}")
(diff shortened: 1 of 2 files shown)


SrTobi:
Issue: https://youtrack.jetbrains.com/issue/SCL-22406

SrTobi:
Hi @rochala, thanks for the pr. I merged it via the internal pipeline. Thanks :pray:

SrTobi:
Ahh regarding the bug you mentioned: I don't think it's a bug, just that it's ambiguous what should happen and it chooses the other option of what you think :grinning: Normally when you have a cursor `|` in let's say `aaa|ccc` and we insert `bbb` at the cursor we get `aaabbb|ccc`, but when you think about it more as index instead of a cursor, we could also get `aaa|bbbccc`. So I think in this case it's just the later version. Actually the platform is often aware of it. For example: there is a Marker with which you can mark a  range in a Document and which tracks the ranges through changes to the document. That marker can then be configured to be right-greedy and/or left-greedy, meaning that when text is inserted at the border of the range, whether to include that text or not.

PR #707, opened 2025-10-11
SCL-21400 Highlight missing enum constructor arguments without "extends"

meszka:
Fixes [SCL-21400](https://youtrack.jetbrains.com/issue/SCL-21400/Highlight-missing-enum-constructor-arguments-without-extends)


sugakandrey:
@meszka Thanks for the help, I've cherry-picked your changes to run them through our CI pipeline. Soon they should be available in the nightly plugin.

PR #708, opened 2025-10-14
An action to add a parameter to a method #SCL-24374

makingthematrix:
If the user adds an argument to the method call and opens the action menu when the new argument is highlighted, they should see now a new intention: "Add parameter(s) to a method". After choosing, a new parameter of the inferred type will be added the method's declaration, and the caret will be moved to the parameters clause, in the style of "Create Method".

Limitations:
1. The action works only for arguments added to the first parameters clause. If the method has more clauses, the intention does not appear in the popup.
2. The caret is moved to the first parameter in the clause even if new parameters start at subsequent places.

### DIFF
--- a/scala/codeInsight/resources/META-INF/codeInsight.xml
+++ b/scala/codeInsight/resources/META-INF/codeInsight.xml
@@ -106,6 +106,13 @@
             <language>Scala</language>
             <categoryKey>intention.category.scala.argument.conversion</categoryKey>
             <bundleName>messages.ScalaCodeInsightBundle</bundleName>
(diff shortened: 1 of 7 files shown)


makingthematrix:
Screenshot for the limitation #1:
Screenshot 2025-10-14 at 15 28 11

Screenshot for the limitation #2:
Screenshot 2025-10-14 at 15 27 31

SrTobi:
Issue: https://youtrack.jetbrains.com/issue/SCL-24374

SrTobi:
Issue: https://youtrack.jetbrains.com/issue/SCL-24374
I reworked the code a bit and merged it :+1:

PR #709, opened 2025-10-15
SCL-24481: Support running separate ScalaTest tests via Bazel

kusaeva:
Previously, support for running individual ScalaTest tests through Bazel was implemented in the Bazel plugin (ijwb). However, with the introduction of the new official Bazel plugin from JetBrains, it makes more sense to move this functionality to the Scala plugin, where test framework-specific logic naturally belongs.

With this change, Bazel users can run individual test specs using gutter icons:
Screenshot 2025-10-14 at 14 51 25

Note: Running individual tests from a spec still requires some changes on the Bazel plugin side (which I hope will be accepted as well).


scala/integration/intellij-bazel/src/org/jetbrains/plugins/scala/bazel/BazelTestRunLineMarkerContributor.scala
```diff
@@ -0,0 +1,77 @@
...
+class BazelTestRunLineMarkerContributor extends BazelRunLineMarkerContributor {
+  private val TEST_ARG = "-t"
+
+  override def isDumbAware: Boolean = true
+
+  override def shouldAddMarker(psiElement: PsiElement): Boolean =
+    psiElement match {
+      case leaf: LeafPsiElement if leaf.getElementType == ScalaTokenTypes.tIDENTIFIER =>
+        leaf.getParent match {
+          case _: ScTypeDefinition | _: ScReferenceExpression => true
```
unkarjedy: Wouldn't hurt adding a comment why only these 2 types are handled here

scala/integration/intellij-bazel/src/org/jetbrains/plugins/scala/bazel/BazelTestRunLineMarkerContributor.scala
```diff
@@ -0,0 +1,77 @@
...
+import org.jetbrains.plugins.scala.lang.psi.api.expr.{ScInfixExpr, ScMethodCall, ScReferenceExpression}
+import org.jetbrains.plugins.scala.lang.psi.api.statements.ScFunctionDefinition
+import org.jetbrains.plugins.scala.lang.psi.api.toplevel.typedef.{ScClass, ScTypeDefinition}
+
+import org.jetbrains.plugins.scala.extensions.ObjectExt
+import org.jetbrains.plugins.scala.testingSupport.test.scalatest.ScalaTestConfigurationProducer
+
+import java.util
+
+class BazelTestRunLineMarkerContributor extends BazelRunLineMarkerContributor {
```
unkarjedy: (minor)
I would add a scaladoc link to `org.jetbrains.plugins.scala.testingSupport.test.ui.ScalaTestRunLineMarkerProvider` here as a reminder of where is the rest scala test frameworks handling is made

scala/integration/intellij-bazel/src/org/jetbrains/plugins/scala/bazel/BazelTestRunLineMarkerContributor.scala
```diff
@@ -0,0 +1,77 @@
...
+      case leaf: LeafPsiElement if leaf.getElementType == ScalaTokenTypes.tIDENTIFIER =>
+        leaf.getParent match {
+          case _: ScTypeDefinition | _: ScReferenceExpression => true
+          case _ => false
+        }
+      case _ =>
+        false
+    }
+
+  override def getSingleTestFilter(element: PsiElement): String =
```
unkarjedy: Hm, so the `getSingleTestFilter` method comes from `BazelJavaRunLineMarkerContributor` (and from `BazelRunLineMarkerContributor`) and has some default implementation.

It looks then, that it's implied that one must not implement 2 separate providers for tests and non-tests (It's hard to tell what was the original idea without the documentation).

However, currently 2 providers are implemented:
`BazelScalaRunLineMarkerContributor` and `BazelTestRunLineMarkerContributor` which means that 2 test logics will be executed, which seems "off".

cc @jastice

scala/integration/intellij-bazel/src/org/jetbrains/plugins/scala/bazel/BazelTestRunLineMarkerContributor.scala
```diff
@@ -0,0 +1,77 @@
...
+      for {
+        testClass <- Option(PsiTreeUtil.getParentOfType(psiElement, classOf[ScClass]))
+        location <- Option(PsiLocation.fromPsiElement(testClass.getProject, psiElement))
+        testClassWithTestName <- ScalaTestConfigurationProducer.apply().getTestClassWithTestName(location)
+        testName <- testClassWithTestName.testName
+      } yield escape(testName)
+    name.getOrElse(util.List.of[String]())
+  }
+
+  private def escape(testName: String): util.List[String] =
```
unkarjedy: Shouldn't the same be relevant to the regular ScalaTest runners (non-Bazel)?
Have you checked if there is any handling for it there?

unkarjedy:
Note, there are no any tests covering the functionality, which will certainly eventually lead to regressions.
I understand that it can be hard to set them up (the hardest part might be setting up the Bazel project).
I don't expect you to invent the tests for Scala-Bazel support, it can require quite some time to do it.
I am just stating the current status out loud.

scala/integration/intellij-bazel/src/org/jetbrains/plugins/scala/bazel/BazelTestRunLineMarkerContributor.scala
kusaeva: To be honest, I borrowed this part from [ijwb's scala-bazel integration](https://github.com/bazelbuild/intellij/blob/v2025.07.24-stable/scala/src/com/google/idea/blaze/scala/run/producers/ScalaTestContextProvider.java#L125-L129)

I tried to look for an analog somewhere in the ScalaTestRunLineMarkerProvider, but unsuccessfully

scala/integration/intellij-bazel/src/org/jetbrains/plugins/scala/bazel/BazelTestRunLineMarkerContributor.scala
kusaeva: You're right, I see the concern now. When I implemented this, I followed the pattern from the existing ScalaRunLineMarkerContributor and ScalaTestRunLineMarkerProvider, which are split into separate classes. I tried to keep my changes less invasive by maintaining this same structure.

However, I agree that the presence of the getSingleTestFilter method in BazelRunLineMarkerContributor does suggest that both test and non-test execution should be implemented in a single class. 
Should I try to merge BazelScalaRunLineMarkerContributor and BazelTestRunLineMarkerContributor into a single contributor class?

unkarjedy: I asked folks from the Bazel team for their thoughts as well.
Let's wait

kusaeva:
Meanwhile I've added support for running zio-test when the zio-intellij plugin is installed. 

Given that ZIO is quite popular in the Scala community, I thought this would be a useful addition. However, if you feel this is outside the scope of this project, please let me know and I'll revert the changes.

scala/integration/intellij-bazel/resources/scalaCommunity.intellij-bazel.xml
```diff
@@ -16,5 +17,6 @@
     </extensions>
     <extensions defaultExtensionNs="com.intellij">
         <runLineMarkerContributor implementationClass="org.jetbrains.plugins.scala.bazel.BazelScalaRunLineMarkerContributor" language="Scala" order="before ScalaRunLineMarkerContributor"/>
+        <runLineMarkerContributor implementationClass="org.jetbrains.plugins.scala.bazel.BazelTestRunLineMarkerContributor" language="Scala" order="first"/>
```
kusaeva: `order="first"` here to override [zio-intellij's](https://github.com/zio/zio-intellij) runLineMarkerContributor

unkarjedy: Again, the best would be to have it as a line-comment directly in the `.xml` file, not to loose this information

scala/integration/intellij-bazel/src/org/jetbrains/plugins/scala/bazel/BazelTestRunLineMarkerContributor.scala
jastice: @mrkocot maybe you have some in put here?

mrkocot: Having two separate contributors will not break anything, but `BazelRunLineMarkerContributor` is used for both test and non-test situations. Merging these into one contributor should simplify the flow and make it easier for future maintenance

kusaeva: @unkarjedy so what do you think now: should I merge contributors or keep as they are?

unkarjedy: Yes, please merge them given we are talking about maintaining code in Scala Plugin repository now.
We still could extract test logic calculation to separate some separate class to not to mix test and non-test logic, especially if  there is a lot of code in the merged version.

The main point is to not register 2 separate providers

kusaeva: Done, review please

unkarjedy: @kusaeva We are currently in the end-of-the-year pre-release state with a lot of things on the table.
So please, expect some delays, we will try to handle/review it within a week.

kusaeva: Got it, thanks for the clarity

kusaeva: @unkarjedy Gentle ping on this PR 🙂

unkarjedy: Actually the main point was to not register a single instance of `BazelJavaRunLineMarkerContributor` (given that this is what "Bazel" plugin dictates.
But it sill makes sense to extract test-related logic to a different file/class/etc...
This will simplify the maintenance.

I will do it myself while merging the PR

unkarjedy:
I pushed the squashed changes via my personal branch to run through our internal CI.
It also includes some fixes from me.

PR #710, opened 2025-12-12
Update README.md to mention Java 21

lihaoyi:
Seems to be necessary to build intellij-scala since b114d38200 which set `kotlincJvmTarget := "21"`


unkarjedy:
@vasilmkd Can you take a look? Also ARAIR in 261 it's now JDK 25, right? If so, could you update it in 261 in advance then?

vasilmkd:
Yes, building the Scala Plugin for IntelliJ IDEA 2025.3 now requires JDK 21.
For 2026.1, the requirement may become JDK 25. This is not currently decided by the platform.
In 2026.1, the Java runtime of IDEA is indeed JDK 25, but it is still not necessary for building the code.

@lihaoyi Thanks for the contribution. Please change the target branch to `idea253.release` and I will get this merged. Thank you.

vasilmkd (changes requested):
Please change the target branch to `idea253.release`.

lihaoyi:
@vasilmkd I tried doing it, but it looks like github went crazy with 1000s of files changed. Might be easier to just close this PR and make the change yourself

vasilmkd:
Ok, thank you!

PR #711, opened 2025-12-19
make scalaCommunity.bsp and scalaCommunity.testing-support content modules public #SCL-24490

chashnikov:
This is a quick fix for the main problem described in SCL-24490, it'll unblock enabling visibility checks in the platform.


SrTobi:
was merged manually into idea261.x. Thanks

PR #716, opened 2026-03-15
SCL-25169 Scala plugin creates incorrect run configuration for ScalaTest using Bazel

pjanczyk:
**Before:** IntelliJ executes `bazel test //src/test/scala:tests --test_arg='-t "example"'`, which fails with `Argument unrecognized by ScalaTest's Runner` error.

**After:** IntelliJ executes `bazel test //src/test/scala:tests --test_arg='-t' --test_arg='example'`, which works.

---

Nested test cases with spaces and special characters also work as expected:


vasilmkd (approved):
Sorry for the late reply! Thank you for the contribution. I'll merge this commit in the internal CI to run all tests. The PR will automatically close. Please do not make further changes.

PR #717, opened 2026-04-17
[change signature] support types defined in function signature #SCL-24118

AdamJKing:
Fixes [#SCL-24118](https://youtrack.jetbrains.com/issue/SCL-24118/Change-generic-parameter-to-Any-after-change-signature-refactor). This issue was happening for two reasons;
1. The search space for type paramters uses the fragment's context, which the change-signature dialog sets to the surrounding class and not the target function. This means it misses the function's type parameters.
2. The code maps type resolution errors as `Any`, and forgets the previous type.

Using the function as the context instead and not setting a child (which isn't searched anyway) fixes the issue.


unkarjedy:
Hello.
Sorry for the latest delay, we were quite overloaded and didn't have time to look into it.
It's still on our radar.

One of the issues is that AFAIR the issue is more complicated.
Last time I looked into it 6 months ago I hit some IntelliJ Platform limitations that didn't allow me to fix it without workarounds **_in a general case_**.
Unfortunately I lost the context and this topic requires diving into it again.

PR #718, opened 2026-05-05
[testing-support][intellij-bazel] specs2 gutter triangles + Bazel --test_filter (#SCL-25391)

gosha212:
## Summary

Closes #SCL-25391.

Two related fixes for IntelliJ Scala's Specs2 support:

1. **`[testing-support]` Per-test gutter triangles for specs2.** `ScalaTestRunLineMarkerProvider#infoForTestMethodRef` previously had cases for ScalaTest and MUnit but not Specs2, so individual tests (`"x" in {…}`, `>>`, `!`) and scope blocks (`should`, `can`) had no gutter icons even though the class itself did. Adds a `Specs2TestLocationsFinder` parallel to `ScalaTestTestLocationsFinder`, using the existing `TestNodeProvider.isSpecs2TestExpr/isSpecs2ScopeExpr` predicates already shared with `Specs2ConfigurationProducer`.

2. **`[intellij-bazel]` Bazel `--test_filter` for specs2.** Following SCL-24481 (PR #709) which wired ScalaTest+ZIO single-test filters via `BazelScalaTestRunLineMarkerLogic`, this restores specs2 parity. Without it, clicking the gutter on a specs2 spec produces `--test_filter=<classFQN>`, which fails to whole-string-match specs2-junit's scope-prefixed `Description` names — `JUnit4Runner` reports `No tests found matching RegEx`. Adds `Specs2BazelTestFilter`, a Scala 3 port of [Google's deleted IJwB `Specs2Utils.java`](https://github.com/bazelbuild/intellij/blob/f62b2670648d/scala/src/com/google/idea/blaze/scala/run/Specs2Utils.java) (commit `f62b2670648d`, removed in `36a4550602` *"Remove IJwB build (#7868)"*). Filter format identical to Google's:
   - test ref: `<classFQN>#\Q<scope> should::<test>\E$`
   - scope ref: `<classFQN>#\Q<scope> should\E::`

## Architecture note

`Specs2BazelTestFilter` lives in `testing-support` (Scala 2.13) rather than `intellij-bazel` (Scala 3) for two reasons:
1. It is a pure PSI-to-string utility with no Bazel-runtime dependency — only the *call site* in `BazelScalaTestRunLineMarkerLogic` is bazel-specific.
2. `intellij-bazel` has no test source root today; `testing-support` already has `ScalaFixtureTestCase` infrastructure used by the unit test (`Specs2BazelTestFilterTest`).

## Tests

- `Specs2BazelTestFilterTest` (new) — 7 cases covering top-level `in`, scoped `in`/`>>` (under `should`), `in` under `can`, `should` scope, `can` scope, and a non-test-element negative case. Uses the new `WithSpecs2_4` fixture trait.
- `Specs2RunLineMarkerProviderTest` for the gutter side will land as a follow-up commit on this branch — the IntelliJ test fixture boot is slow on our hardware and we wanted to surface the changes for review without further delay.

## Verification

- `sbt testing-support/Test/compile` and `sbt intellij-bazel/Compile/compile` both pass on JDK 21.
- Manual smoke test: opened a `scala_specs2_junit_test` Bazel target in the sandboxed IDE, clicked gutter triangles on test/scope/class, confirmed each generates the expected filter and runs.

## Related

- SCL-24481 / #709 — parent change that introduced Bazel run-line-markers for ScalaTest and ZIO. This extends that work to specs2.
- `bazelbuild/intellij` commit `f62b2670648d` — Google IJwB Scala plugin's specs2 producer (since deleted), source of the proposed filter format.

### DIFF
--- a/scala/integration/intellij-bazel/src/org/jetbrains/plugins/scala/bazel/BazelScalaTestRunLineMarkerLogic.scala
+++ b/scala/integration/intellij-bazel/src/org/jetbrains/plugins/scala/bazel/BazelScalaTestRunLineMarkerLogic.scala
@@ -11,6 +11,7 @@ import org.jetbrains.plugins.scala.lang.psi.api.expr.{ScInfixExpr, ScMethodCall,
 import org.jetbrains.plugins.scala.lang.psi.api.statements.ScFunctionDefinition
 import org.jetbrains.plugins.scala.lang.psi.api.toplevel.typedef.{ScClass, ScDerivesClauseOwner, ScObject, ScTypeDefinition}
 import org.jetbrains.plugins.scala.testingSupport.test.scalatest.ScalaTestConfigurationProducer
+import org.jetbrains.plugins.scala.testingSupport.test.specs2.{Specs2BazelTestFilter, Specs2TestFramework}
 
 /**
  * Utilities inside contain logic for running entire test classes and individual tests from ScalaTest and ZIO-test via Bazel.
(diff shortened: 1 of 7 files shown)


vasilmkd:
Thanks for the contribution. I will run it in our internal CI to see if the tests pass and I will do a review afterwards.

tsimonavicius:
@vasilmkd , how was the CI? could you review the PR already?

vasilmkd:
After adding the changes, the Specs2 tests became flaky in our CI, so I was working on fixing them. Review incoming.

PR #720, opened 2026-06-22
[bsp] fallback to javaVersion before moduleJdkVersion when setting language level during import #SCL-25592 fixed

Konicai:
When importing a Java module using BSP and the `JvmBuildTarget` response is received, the `javaVersion` field may be lower than the version of the JDK located at `javaHome`.

Before these changes, if `BspMetadata.languageLevel` was null, the language level would fallback to that of the JDK at `javaHome`.
These changes fallback the language level of the module to `javaVersion`, and if that is missing for some reason, then it further falls back to the version of the JDK at `javaHome`.

Related: https://youtrack.jetbrains.com/issue/SCL-25592

### DIFF
--- a/bsp-builtin/bsp/src/org/jetbrains/bsp/data/BspMetadataService.scala
+++ b/bsp-builtin/bsp/src/org/jetbrains/bsp/data/BspMetadataService.scala
@@ -44,7 +44,8 @@ class BspMetadataService extends ScalaAbstractProjectDataService[BspMetadata, Mo
       val moduleJdkVersion = moduleJdk.map(_.getVersionString)
 
       Option(data.languageLevel)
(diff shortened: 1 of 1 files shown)


unkarjedy:
Please rebase the changes to idea262.x
Also please make sure there is a regression test covering the change.
We already have plenty of tests for SBT and BSP

azdrojowa123:
Hello @Konicai, 
Thank you for creating PR. 
At the moment, we don't have any dedicated tests for Mill, but the same problem can be reproduced with sbt/BSP.
I'll take care of this, implement the tests, and let you know when the changes are published 🙂

Konicai:
Hi, thank you so much!

